PR #4, opened 2019-05-06
Move files around to comply with Maven standard directory layout

pviotti:
Changes included in this PR:

- move all source code files into `src/main/java` and all tests into `src/test/java` as in the [Maven standard directory layout ](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) - so all tests have been separated from the main application code
- remove a few lines not needed from `pom.xml` 
  * `<groupId>org.apache.maven.plugins</groupId>` can be omitted
  * `<include>**/*.xml</include>` : there are no xml file in `src`
  * `<exclude>**/*.java</exclude>`:  there are no java files in rsc

(To fully comply with the Maven directory layout we should also move the resource files in `src/test/resources` and `src/main/resources`, but that's a more delicate change and can be done later...)


pviotti:
Hi there! Any feedback on this PR? 🙈 🙉 🙊

mwithi:
Hi @pviotti , no time to process it yet... Anyway since it would change seriously the code layout we will give it low priority (we already postponed this kind of refactoring) in order to facilitate the comparison with the old code still ongoing on SourceForge for last adjustments (like sync with 1.8.4) and small new features.

Does it create any problem for the rest of the work on the [OP-68](https://openhospital.atlassian.net/projects/OP/issues/OP-68)?

pviotti:
So is there still some active development on SourceForge? Wouldn't it make more sense to close the project there and treat this one as the only source of truth for current and future development?
Anyway, this PR is not changing any line of code, and it'd be just a first step to modernize and clean up the code base (also to make it more approachable for new contributors!).  
And it's sort of related to OP-68 as well, as it makes jar artifacts smaller (because they don't include test files anymore). Besides separating test, source and resource files ease the use of Maven plugins and, for instance, enables [better testing practices](https://www.baeldung.com/spring-jpa-test-in-memory-database).

pviotti:
Also, as I reported at the bottom of [this wiki page](https://openhospital.atlassian.net/wiki/spaces/OHC/pages/568918122/Software+Distribution), I'm having some trouble with OP-68, since none of the currently released  All-In-One distributions for Linux and Windows works on my machines - so I cannot replicate and automate something that doesn't work to begin with :grimacing:  
Should I open an issue about this? Is anyone experiencing the same or it works for you guys on Windows 10 and on Linux?

mwithi:
Hi @pviotti  for the WikiPage you'll be answered over there, the two "portable" versions do work, and until we don't release the version 1.9 in the roadmap, the SourceForge is the official repository, and it will be working (anyway) as long as many projects do rely on its branches.

As for Maven you are absolutely right, it will be done! Only not now, since around the same code are working other people and such change would mess up other pull-requests coming from previous forks. I think this kind of refactorying could be done quickly later on, just before the next planned release version 1.9.

With all the above said how do you think we can accomplish/mediate your requests/desires with the OH dev Community needs?
Can we stay with the code as is now for testing and then optimize it in the last rush after all other contributions will be merged?

Meanwhile we are totally available for any understanding of the project itself.

pviotti:
> Hi @pviotti for the WikiPage you'll be answered over there, the two "portable" versions do work,
 
Mh.. actually they do not work on any of my Linux and Windows 10 machines, I just tried again. I'm sure it's just some minor scripting issue, but since it's the first thing that potential users try, I think it's worth getting it right. Anyway, I'll reply to this on that wiki page.

> Only not now, since around the same code are working other people and such change would mess up other pull-requests coming from previous forks. 

There are 7 public forks, of which 2 have new commits, of which only 1 ([this one](https://github.com/informatici/openhospital-core/compare/master...averni:master)) has changes in a path that would conflict with this PR. So, all in all, if we count only the public forks, I reckon it would be feasible (with the conflicting repo having to do a `git rebase upstream/master`). But anyway, it's your call.

> Can we stay with the code as is now for testing and then optimize it in the last rush after all other contributions will be merged?

Sure. Probably it was just a misunderstanding on my part - and in that case sorry about that.  
I thought the code on GitHub was only for the new major version (2.0), and so we would be less constrained by the legacy  version on SourceForge.

mwithi:
We are ready to release the official 1.9.0. @pviotti do you think we can proceed now with this refactoring?

Does it need to be to be done on "gui" component as well?

PR #25, opened 2019-09-25
U2g op-80 audit fields

MartinezX21:


mwithi:
The job is really good and simple. 

Nevertheless the problem comes when JOIN query are done with plain sql code, like on JUnit Tests **_testIoGetAdmittedPatients_** (means _findAllBySearchAndDateRanges_):
`Encountered a duplicated sql alias [CREATED_BY] during auto-discovery of a native-sql query;`

Either we convert all SQL-like codes into JPQL or we find the way to prefix the new fields for each table.

mwithi:
> The job is really good and simple.
> 
> Nevertheless the problem comes when JOIN query are done with plain sql code, like in **_findAllBySearchAndDateRanges_** (just follow JUnit Tests **_testIoGetAdmittedPatients_**):
> `Encountered a duplicated sql alias [CREATED_BY] during auto-discovery of a native-sql query;`
> 
> Either we convert all SQL-like codes into JPQL or we find the way to prefix the new fields for each table.

A third option could be to add ALIASES to all tables in all native-queries, not very fast but easier maybe

uni2grow:
@mwithi Changes has been done as requested.

mwithi:
Hi @uni2grow, I think this solution will work fine, but before to test it let me ask you if you tried to introduce a "prefix" in the code for each table so that all fields (also existing) got it automatically?

mwithi:
Dear @uni2grow, I think that for the GROUPMENU we can reutilize the GM_ACTIVE field, by converting 'Y' -> 1 and 'N' -> and then ALTER the column to TINYINT(1).
All *_ACTIVE fields should be TINYINT(1) - 1 byte only
Also, I can see that some JUnit Test are failing for persistence problems, can you work on it?

PR #26, opened 2019-10-10
Update applicationContext.xml

mwithi:
Fixing startup when connectivity is not available

### DIFF
--- a/rsc/applicationContext.xml
+++ b/rsc/applicationContext.xml
@@ -3,22 +3,24 @@
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:p="http://www.springframework.org/schema/p"
 	xmlns:context="http://www.springframework.org/schema/context"
- 	xmlns:jpa="http://www.springframework.org/schema/data/jpa" 
- 	xmlns:tx="http://www.springframework.org/schema/tx"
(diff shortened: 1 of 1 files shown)


rsc/applicationContext.xml
```diff
@@ -3,22 +3,24 @@
...
+		http://www.springframework.org/schema/context/spring-context-4.3.xsd
+		http://www.springframework.org/schema/aop 
+		http://www.springframework.org/schema/aop/spring-aop-4.3.xsd
+		http://www.springframework.org/schema/data/jpa 
+		http://www.springframework.org/schema/data/jpa/spring-jpa-1.11.xsd
+		http://www.springframework.org/schema/tx 
+		http://www.springframework.org/schema/tx/spring-tx-4.3.xsd
+		http://www.springframework.org/schema/data/repository 
+		http://www.springframework.org/schema/data/repository/spring-repository-1.11.xsd
+		">
```
pviotti: I'm not sure I understand: what did change here exactly? I mean: isn't it going to call these addresses to get the schemas anyway upon startup..? :thinking:

mwithi: It's because of the SAX Parser: if you specify the exact version you have in the jars, it WILL NOT fetch them from URLs, it will otherwise (that's what was causing the issue)

PR #33, opened 2020-01-04
OP-143: "patient profession" and "marital status" fields

aalsabag:
These are the necessary changes for openhospital-core for OP-143 . They are the addition of the maritalStatus and profession fields. The necessary updates have been made to the sql scripts to add these two columns. The Pr for the gui changes has also been submitted


mwithi:
Hi @aalsabag , sorry for the late review, the changes to the DB should be addressed in a separated script "step_XX_patient_profession_and_marital_status.sql" (keep the XX).

This because if we want to upgrade a previous version of OpenHospital with your changes, we just need to apply the different scripts between the two versions.

Please can you make this change first and test that everything is fine? Thanks!

aalsabag:
> Hi @aalsabag , sorry for the late review, the changes to the DB should be addressed in a separated script "step_XX_patient_profession_and_marital_status.sql" (keep the XX).
> 
> This because if we want to upgrade a previous version of OpenHospital with your changes, we just need to apply the different scripts between the two versions.
> 
> Please can you make this change first and test that everything is fine? Thanks!

@mwithi all done. A new sql script has been created and tested.

aalsabag:
Thanks for this catch. Sorry this must have slipped my mind. Updated! @mwithi

PR #38, opened 2020-02-04
Op 102 master refactoring for api

npasquetto:
This pull request bring:
* added mvnw to the project, to simplify the build process for new developers; this has required also [a fix](https://github.com/intesys/openhospital-core/blob/30fc43d7e841b148e556f1cef92ebb05fcd75837/.travis.yml#L6) in travis pipeline definition.
* made some classes public to use them in [API project](https://github.com/informatici/openhospital-api); this has required also [files renaming](https://github.com/intesys/openhospital-core/commit/3e375514eeda938745601eba2783e16bf06df0fc).
* mapping of the PAT_DELETED column of the Patient @Entity, to better manage it in [API project](https://github.com/informatici/openhospital-api).


src/org/isf/patient/service/PatientIoOperationRepository.java
```diff
@@ -21,6 +21,8 @@
 
     @Query(value = "SELECT * FROM PATIENT WHERE (PAT_DELETED='N' OR PAT_DELETED IS NULL) ORDER BY PAT_NAME, ?#{#pageable}", nativeQuery= true)
     public List<Patient> findAllWhereDeleted(Pageable pageable);
+
+    public List<Patient> findAllByDeletedIsNullOrDeletedEqualsOrderByName(String patDeleted, Pageable pageable);
```
mwithi: Why the @Query annotation is missing here?

pom.xml
```diff
@@ -12,7 +12,7 @@
 	<groupId>org.isf</groupId>
 	<artifactId>OH-core</artifactId>
 	<packaging>jar</packaging>
-	<version>1.9</version>
+	<version>2.0-SNAPSHOT</version>
```
mwithi: Version should be kept 1.9 to be compliant with other modules

src/org/isf/patient/service/PatientIoOperationRepository.java
npasquetto: Because this is not a custom Query but the ORM deriving the query from the method name directly.

mwithi: I see, thanks!

PR #46, opened 2020-02-28
OP-225 Update to Java 8

larsgrefer:


larsgrefer:
@mwithi Is there any documentation about what these native libraries are doing and what they are used for?

It also seems that not all native libs are available for each OS/Arch combination. Is this correct?

mwithi:
> @mwithi Is there any documentation about what these native libraries are doing and what they are used for?
@larsgrefer documentation is missing, we as much we can on the linked Jira issue (OP-108) and its linked ones. In short libraries are there for Webcam and for GSM module communication. Also some issues are there for the internal DICOM viewer.

> It also seems that not all native libs are available for each OS/Arch combination. Is this correct?
Is correct, in the same OP-108 has been uploaded an (old) compatibility matrix.

If we move all related features (let's say Webcam, GSM and DICOM) to a different component, we should be also able to bump the Java version for "core" and "gui" painlessly...

mwithi:
> > @mwithi Is there any documentation about what these native libraries are doing and what they are used for?

you can find something on [Admin Manual](https://github.com/informatici/openhospital-doc/blob/master/doc_admin/AdminManual.adoc) about SMSENABLED, VIDEOMODULEENABLED, DICOMMODULEENABLED.
It could also affect RECEIPTPRINTER but I'm not sure.

mwithi:
Hi @larsgrefer, we are ager to bump Java version (we released `1.10`). We can rebase this onto the new `develop` branch, but maybe it is bettere to start as fresh from the latter. 

Please let us know if you can still work on it. Thanks!

larsgrefer:
I think it will be better to recreate this from scratch as there are many merge conflicts which would need to be resolved.

Would you like to update to Java 8 or directly to 11?

mwithi:
> Would you like to update to Java 8 or directly to 11?

I think that Java 8 is the most pain-less transition we can think as for now, it's still LTS so we can do it quite serenely, what do you think?

larsgrefer:
We could start bumping to Java 8 and then see if and when we want to jump to 11.

The latest version ist currently 14. Java 11 is already in LTS mode, too.

mwithi:
Yeah, let's do it step by step.

mwithi:
Changed PR target to the new default branch `develop`

pviotti:
> Nevertheless I think that this merge cannot be independent from changing also https://github.com/informatici/openhospital-gui and other components.

Actually this PR only impacts this repo. Anyway, I've submitted the PR for the gui project as well: https://github.com/informatici/openhospital-gui/pull/183

mwithi:
> Wdyt?

What if we just create another PR for that like https://github.com/informatici/openhospital-gui/pull/183?

Then this PR would just remain unchanged or needing rebase after #102 will be merged (this weekend)

mwithi:
> Then this PR would just remain unchanged or needing rebase after #102 will be merged (this weekend)

#102 superseded by #132, that has been merged now

mwithi:
superseded by https://github.com/informatici/openhospital-core/pull/156

PR #50, opened 2020-03-01
Fix code style issues

larsgrefer:


src/org/isf/admission/test/Tests.java
```diff
@@ -354,8 +354,8 @@ public void testIoNewAdmission()
 			
 			Admission admission = testAdmission.setup(ward, patient, admissionType, diseaseIn, diseaseOut1, 
 					diseaseOut2, diseaseOut3, operation, dischargeType, pregTreatmentType, 
-					deliveryType, deliveryResult, true);;
-					result = admissionIoOperation.newAdmission(admission);
+					deliveryType, deliveryResult, true);
+            result = admissionIoOperation.newAdmission(admission);
```
mwithi: is the conseguent indentation correct? tabs have been converted into spaces

larsgrefer:
@mwithi These are just some of the default inspections of IntelliJ IDEA

src/org/isf/admission/test/Tests.java
mwithi: I see, so you should also have noticed a number of warnings we could get rid of.

Anyway, is the above conseguent indentation correct? Tabulations have been converted into spaces maybe

larsgrefer: Seems like some tabs were replaced with spaces. I see if I can fix this.

larsgrefer:
@mwithi I've re-created the changes with the correct indentation and added a .editorconfig file, so all editorconfig-aware editors will use tabs from now on.

src/org/isf/admission/test/Tests.java
mwithi: ok

PR #58, opened 2020-03-13
Op 80 audits fields

mwithi:


mwithi:
Hi @MartinezX21 , should we have had introduced default values for already existing data?

MartinezX21:
I don't think that should be done, since the lack of values doesn't cause a problem for the existing data. If default values are put in, they will be irrelevant

mwithi:
ok, what do you think about to make an audit report/feature in order to analize those data?

PR #61, opened 2020-03-15
OP-131: Replace native SQL with JPQL (modules with M)

marcoebbinghaus:


src/org/isf/medicals/service/MedicalsIoOperationRepository.java
```diff
@@ -11,48 +11,48 @@
 
 @Repository
 public interface MedicalsIoOperationRepository extends JpaRepository<Medical, Integer> {
-	@Query(value = "SELECT * FROM MEDICALDSR where MDSR_DESC like :description order BY MDSR_DESC", nativeQuery= true)
+	@Query(value = "SELECT m FROM Medical m where m.description like :description order BY m.description")
     List<Medical> findAllWhereDescriptionOrderByDescription(@Param("description") String description);
-	@Query(value = "SELECT * FROM MEDICALDSR JOIN MEDICALDSRTYPE ON MDSR_MDSRT_ID_A = MDSRT_ID_A order BY MDSR_DESC", nativeQuery= true)
+	@Query(value = "SELECT m FROM Medical m JOIN m.type mt order BY m.description")
```
tomek39856: join is not needed jpa should do it for you

src/org/isf/medicals/service/MedicalsIoOperations.java
```diff
@@ -87,39 +86,34 @@ public Medical getMedical(
...
 				medicals = getMedicals(null);
 			} else {
-				medicals = (ArrayList<Medical>)repository.findAllOrderBySmartCodeAndDescription();
+				medicals = repository.findAllOrderBySmartCodeAndDescription();
 			}
 
 		}
 
-		return medicals;
+		return medicals == null ? new ArrayList<Medical>() : new ArrayList<Medical>(medicals);
```
tomek39856: is it possible that medicals could be null? jpa repository will return empty list if nothing is found

src/org/isf/medicals/service/MedicalsIoOperations.java
```diff
@@ -87,39 +86,34 @@ public Medical getMedical(
...
 	public ArrayList<Medical> getMedicals(String type, boolean nameSorted) throws OHServiceException {
-		ArrayList<Medical> medicals = null;
-
-		List<Object> parameters = new ArrayList<Object>();
-
-		StringBuilder query = new StringBuilder();
-		query.append("SELECT * FROM MEDICALDSR JOIN MEDICALDSRTYPE ON MDSR_MDSRT_ID_A = MDSRT_ID_A ");
+		List<Medical> medicals;
 		
 		if (type != null) {
```
tomek39856: It's only suggestion but maybe it's worth to extract 2 private methods?

src/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -575,37 +572,37 @@ protected boolean updateMedicalWardQuantity(
...
-		}
-		
-		// remve empy lots
-		ArrayList<Lot> emptyLots = new ArrayList<Lot>();
-		for (Lot aLot : lots) {
-			if (aLot.getQuantity() == 0)
-				emptyLots.add(aLot);
-		}
-		lots.removeAll(emptyLots);
+		List<Object[]> lotList = lotRepository.findAllWhereMedical(medical.getCode());
```
tomek39856: Isn't it possible to just return List<Lot> from repository? Than maybe you could avoid doing conversion?

src/org/isf/medicalstock/service/MovementIoOperationRepositoryImpl.java
```diff
@@ -214,23 +226,24 @@ private String _getMovementForPrint(
...
+				"left join mov.lot lot " +
+				"left join mov.supplier sup " +
 				"where ";
 
 		if ((medicalDescription != null) || (medicalTypeCode != null)) 
 		{
 			if (medicalDescription == null) 
 			{
-				query += "(MDSR_MDSRT_ID_A = \"" + medicalTypeCode + "\") ";
+				query += "(medtype.code=:medicalType) ";
```
tomek39856: JPA has something called criteria builder which fits perfect form building queries with optional where clauses like this. It should make code much cleaner and safer. I see that there are named parameters but building queries manually can easily lead to mistake and sql injection vulnerability

https://vladmihalcea.com/query-entity-type-jpa-criteria-api/

src/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
```diff
@@ -13,29 +13,26 @@
 @Repository
```
tomek39856: it's not needed on jpa repository interfaces

src/org/isf/malnutrition/service/MalnutritionIoOperationRepository.java
```diff
@@ -11,10 +11,10 @@
 @Repository
```
tomek39856: repository annotation is not needed

src/org/isf/medicals/service/MedicalsIoOperationRepository.java
```diff
@@ -11,48 +11,48 @@
...
     List<Medical> findAllByOrderByDescription();
-    @Query(value = "SELECT * FROM MEDICALDSR JOIN MEDICALDSRTYPE ON MDSR_MDSRT_ID_A = MDSRT_ID_A where MDSRT_DESC like :type order BY MDSR_DESC", nativeQuery= true)
+    @Query(value = "SELECT m FROM Medical m JOIN m.type mt where mt.description like :type order BY m.description")
```
tomek39856: This will also work without join sth like: 
``` SELECT m FROM Medical m where m.type.description like :type order BY m.description ```
I don't know what is the approach in this project but omitting joins will give shorter queries. And in my opinion more readable but it's only my opinion

src/org/isf/medicals/service/MedicalsIoOperations.java
mwithi: This is not very formally defined in the previous code... so we can follow/introduce the JPA standard keeping in mind that whatever change we do in 'core' must reflect and work as well in 'gui' component and for org.isf.test.Tests

src/org/isf/medicals/service/MedicalsIoOperations.java
mwithi: @tomek what do you mean here?

src/org/isf/medicalstock/service/MedicalStockIoOperations.java
mwithi: @marcoebbi that part should not be removed because it perform an important task on the lot (remove from the list the empty ones). So, until we introduce a different mechanism to do have the same result, we have to keep it. 

@all please pay attention to methods involving the Pharmacy (modules org.isf.medical* and related ones) and the Billing (org.isf.accounting.*)

src/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
mwithi: Hi @tomek, we had a specific tasks for that, are you sure that Spring will not complain in model mapping?
https://openhospital.atlassian.net/browse/OP-102

src/org/isf/medicalstock/service/MedicalStockIoOperations.java
marcoebbinghaus: @mwithi In my opinion I did not change any behaviour here. Have a look at the repository method (findAllWhereMedical(medicalCode)). I just moved the empty lot filtering from code to database query. Should I still revert it?

mwithi: @marcoebbinghaus Ok, I've seen it, but two things: in some cases we could need to know all lots in the history, also the ones with zero quantity, so still we need to retrieve all of them from the DB; don't know if there is already another method for that, maybe we can parameter in case. Second, the query you wrote is slight different, because the original one calculates the LOT quantity as a SUM over a GROUP of movements concerning the same LOT.
Does it make sense also for you?

marcoebbinghaus: @mwithi absolutely. Now that you mentioned it..I remember I struggled with that original query including the grouping and a corresponding replacement with JPQL. But I will have a second look. Thanks for the clarification.

mwithi: thanks @marcoebbinghaus, sure that mechanism should be improved in the future, other option is to leave it as is meanwhile.

src/org/isf/medicals/service/MedicalsIoOperations.java
marcoebbinghaus: I think @tomek39856 you are right. @mwithi I think I don't fully understand what you mean here. The tests are all green (I executed them before each push) and there was nothing changed on the method signature or something. There shouldn't be reason that the tests or clients like the gui might be affected by this change imo. 

I would change this to: return new ArrayList<Medical>(medicals); (as medicals will never be null)

mwithi: @marcoebbinghaus
> There shouldn't be reason that the tests or clients like the gui might be affected by this change imo

It could, because if a check is performed as `resultset == null` instead of `resultset.isEmpty()`, the change would cause the first one to be always `true` and so can lead to unexpected results in the gui if not verified.

marcoebbinghaus: I still think it could not. I don't understand the isEmpty- and == null- comparison either. There was no check (ifempty or == null) in the former code (before my change) and so it always returned simply the assigned value (which could never be null, because the value always came from a repository method call, which never returns null but an empty list in case of a query that didn't return at least one entity).

 There are four places where medicals is assigned a value: 

- `medicals = repository.findAllWhereTypeOrderByDescription(type);`
- `medicals = repository.findAllWhereTypeOrderBySmartCodeAndDescription(type);`
- `medicals = getMedicals(null);`
- `medicals = repository.findAllOrderBySmartCodeAndDescription();`

the third one will result in:

`medicals = (ArrayList<Medical>)repository.findAllWhereDescriptionOrderByDescription(description);` or `medicals = (ArrayList<Medical>)repository.findAllByOrderByDescription();`

So where ever the value of medicals list might be set, the assigned value will come from a repository which will never return null (empty list in worst case).

Therefore the null check is superfluous - as @tomek39856 said - and I would change it to what I proposed above. Also I would extract both outer cases (the block of `if (type != null)` and the correspondig `else` block. In that method into two private methods as suggested by @tomek39856 and I would clarify in the JavaDoc that those methods will never return null.

Please have a look at my newest push, which should clarify what I mean. And if you don't agree please tell me what I should change to code to.

src/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
marcoebbinghaus: So I am not sure if I should remove those annotations or not. But I think if it is possible to remove them, then they should be removed from ALL repositories and not just from the Repositories used in classes of this specific ticket. So I tend to leave it as is in the course of this PR. @mwithi maybe this could be reviewed in the course of the mentioned ticket OP-102?

src/org/isf/medicalstock/service/MovementIoOperationRepositoryImpl.java
marcoebbinghaus: this is awesome! I already changed one method to use the criteria api and it is amazing how much cleaner the code is now! I will push after replacing the other methods using this Query-String-Build-approach as well.

src/org/isf/medicals/service/MedicalsIoOperations.java
mwithi: Sorry @marcoebbinghaus, I thought the comparation `medicals == null` was already there for some reason, hence your decision to maintain it. I agree with @tomek39865 that is superfluous and so with your last push.

I was talking as general way to proceed when changing return values between 'core' layers (like between 'service' and 'manager') and/or 'gui'.

src/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
mwithi: Yes I agree @marcoebbinghaus, I think we can leave it as is at the moment, and discuss the @tomek suggestion on the OP-102 page and, in case, start a new ticket for the overall code base change.

src/org/isf/medicals/service/MedicalsIoOperations.java
tomek39856: Regarding to clean code we should avoid multiple nested ifs so you can extract 2 private methods getWithType and get WithoutType. For instance sth like type != null ? getWithType(type) : getWithoutType()

mwithi: @tomek39856 I get you now and I agree, if it is not too overload for @marcoebbinghaus, generally to change many things in one PR, also with good improvments but out of the issue's scope itself, leads to merge errors. If the change can be done both easily and ensuring tests and gui, it's ok for me.

src/org/isf/medicalstock/service/MedicalStockIoOperations.java
marcoebbinghaus: @mwithi as there is so much logic inside that native query I'm afraid I'm overwhelmed migrating it to use JPQL instead - so I would leave it as is (rollback my changes here). Maybe this can be separated within a following ticket.

marcoebbinghaus:
I applied the feedback given from tomasz and alessandro. I did a rollback for my own dynamic query enhancements (QueryParameterContainer) and use criteria api instead now (which reduces code amount enormously). I also did a rollback for my try to replace the native query in lot repository as there is too much logic inside the query I don't understand completely. I additionally removed superfluous joins in JPQL queries. Please have another look at this PR, thank you.

PR #75, opened 2020-03-31
OP-133: Replace native SQL with JPQL (modules from V to X)

tomek39856:


tomek39856:
Hi I'll fix it during this weekend if it's ok

PR #81, opened 2020-04-13
OP-96 button print logic inside labManager and labIoOperations

giuseppesorge:


giuseppesorge:
After the review, we can avoid to change the core. I'm closing this pull request and I'm asking a new Pull request for the GUI.

PR #88, opened 2020-04-29
Op 206: Replace native SQL with JPQL (modules from O to P)

tomek39856:
1. ApplicationEventPublisher was introduced into PationIoOperations. It publishes event that patient was merged. 
2. Than other modules that contain reference to patient in their repositories has event listeners for PatientMerged event.
3. When they get notified they update references to Patient (changes obsolete patient into merged one)

Everything happens within the same transaction so if some part of merging fails whole operation is rolled back.

Test cases were introduced both in Patient and Patient connected modules. In Patient we check only if Patient entity was updated and if correct event was emmited. Than in each dependent module there is a test which checks if when event arrives patient reference will be updated.

https://www.baeldung.com/spring-events

### DIFF
--- a/src/org/isf/accounting/test/Tests.java
+++ b/src/org/isf/accounting/test/Tests.java
@@ -11,6 +11,7 @@ import org.isf.accounting.model.BillItems;
 import org.isf.accounting.model.BillPayments;
 import org.isf.accounting.service.AccountingIoOperations;
 import org.isf.patient.model.Patient;
+import org.isf.patient.model.PatientMergedEvent;
 import org.isf.patient.test.TestPatient;
 import org.isf.patient.test.TestPatientContext;
 import org.isf.priceslist.model.PriceList;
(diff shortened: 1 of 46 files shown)


src/org/isf/accounting/test/Tests.java
```diff
@@ -845,5 +869,19 @@ private void  _checkBillPaymentsIntoDb(
 		
 		return;
 	}
+
+	private Patient _setupTestPatient(
+		boolean usingSet) throws OHException
+	{
+		Patient patient;
+
+
```
mwithi, line 878: clean out too many empty lines

src/org/isf/operation/test/Tests.java
```diff
@@ -143,6 +144,33 @@ public void testIoGetOperations() throws OHException
 		
 		return;
 	}
+
+	@Test
+	public void testIoGetOperationsShouldFindOperationsWithoutProvidingDescription() throws OHException
+	{
+		String code = "";
+
+
```
mwithi, line 153: avoid many empty lines

PR #84, opened 2020-04-19
Remove reference to Boundless maven repository

pviotti:
The Boundless repository appears to be offline and is causing CI builds failures - see for instance https://travis-ci.org/github/informatici/openhospital-core/jobs/676842582#L1085

This repository builds fine without it anyway.

PR #97, opened 2020-05-24
Replace Travis with GitHub Actions for CI

pviotti:
Since periodically we have problems with Travis CI (see for instance the discussion here: https://github.com/informatici/openhospital-core/pull/84 ), I thought it'd be easier to just switch to GitHub Actions.

Will wait for https://github.com/informatici/openhospital-core/pull/83 to be merged before merging this one.

UPDATE: this can be merged even before https://github.com/informatici/openhospital-core/pull/83 since I've added `docker` support for the tests.


mwithi:
should this be done on each repository? what about actual release process on https://github.com/informatici/openhospital/blob/master/.travis.yml ?

pviotti:
> should this be done on each repository?

No, each repo can have its own CI system. Clearly it'd be better to have the same everywhere but it's not essential.

> what about actual release process on https://github.com/informatici/openhospital/blob/master/.travis.yml ?

That can stay as it is for now. Eventually we can migrate it to Github Actions too - it's not difficult: the only "complex" bit there is the hack to get the changelog.

I think the priority, and the goal for this PR, is to have a way to validate PRs in this repo as soon as possible, since apparently at the moment Travis doesn't do this (not sure why, and I can't help on that as I don't have admin access).

pviotti:
@mwithi this can be merged, so that we have a CI pipeline working.
Here is the first working build: https://github.com/pviotti/openhospital-core/runs/725337916?check_suite_focus=true

mwithi:
> @mwithi this can be merged, so that we have a CI pipeline working.

I was just checking the pricing, GitHub Actions are not *free*, although assuming Travis' insights (not working) we should be fine:
- https://github.com/organizations/informatici/settings/billing
- https://travis-ci.org/github/informatici?tab=insights

pviotti:
> https://github.com/organizations/informatici/settings/billing

 (I don't have permission to see this)

> I was just checking the pricing, GitHub Actions are not free, although assuming Travis' insights (not fully working at the moment) we should be fine (<3000min/months):

yes there are limits to the free tier (same goes for Travis). Looking at [this page](https://github.com/pricing) it seems it's actually 2000 min/month per repository (?). The build I triggered took 2min so I think we're more than fine (we shouldn't have more than 2k push to master or PRs per month... :smile: )

mwithi:
On the organization's page it shows 3000 but as "reduced"
image
but if follow the "Read the announcement" it is 404... I think they had some confusion after the open-up for covid 😅

Anyway we can also put a spend limit, so we should be ok.

mwithi:
Merged... let's see how it works...

pviotti:
it run fine https://github.com/informatici/openhospital-core/actions/runs/120827732
I think you should port it to the target branch of the PR you need to validate. So to `1.10` as well. either using `git cherry-pick <commit code>` and then push or with another PR

mwithi:
yep, done.

mwithi:
and again it is building with tests without problems (like on Travis).

I wonder where the problem is...

PR #114, opened 2020-06-13
Fix obvious spelling errors.

dbmalkovsky:
Spelling errors in method names, message bundle parameters, public objects, etc. were left unchanged.  Only changed comments and few local variable names.

BTW, why are there changelogs in some of the files?  Isn't git history sufficient?

### DIFF
--- a/src/test/java/org/isf/pricesothers/test/TestPricesOthersContext.java
+++ b/src/test/java/org/isf/pricesothers/test/TestPricesOthersContext.java
@@ -35,8 +35,8 @@ public class TestPricesOthersContext
     {
 		jpa.beginTransaction();			
 		jpa.createQuery("SELECT * FROM PRICESOTHERS", PricesOthers.class, false);
-		List<PricesOthers> PricesOtherss = (List<PricesOthers>)jpa.getList();
-		for (PricesOthers pricesOthers: PricesOtherss) 
+		List<PricesOthers> PricesOthers = (List<PricesOthers>)jpa.getList();
+		for (PricesOthers pricesOthers: PricesOthers)
 		{    		
 			int index = savedPricesOthers.indexOf(pricesOthers);
 			
--- a/src/test/java/org/isf/pricesothers/test/Tests.java
+++ b/src/test/java/org/isf/pricesothers/test/Tests.java
@@ -118,9 +118,9 @@ public class Tests
 		{		
 			id = _setupTestPricesOthers(false);
 			PricesOthers foundPricesOthers = (PricesOthers)jpa.find(PricesOthers.class, id); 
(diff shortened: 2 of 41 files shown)


mwithi:
Very nice @dbmalkovsky ! 😉

> BTW, why are there changelogs in some of the files? Isn't git history sufficient?

Do you refer to changelogs in comment? If so, just an old fashioned way coming from far 2005 . Before the code was hosted on CVS...

If you refer to doc/changelogs, it's yet another old fashioned way to distribute offline commit history. They could be removed and linked in the [CHANGELOG.md](https://github.com/informatici/openhospital/blob/master/CHANGELOG.md) file on each release.

src/test/java/org/isf/pricesothers/test/TestPricesOthersContext.java
```diff
@@ -35,8 +35,8 @@ public void deleteNews(
     {
 		jpa.beginTransaction();			
 		jpa.createQuery("SELECT * FROM PRICESOTHERS", PricesOthers.class, false);
-		List<PricesOthers> PricesOtherss = (List<PricesOthers>)jpa.getList();
-		for (PricesOthers pricesOthers: PricesOtherss) 
+		List<PricesOthers> PricesOthers = (List<PricesOthers>)jpa.getList();
+		for (PricesOthers pricesOthers: PricesOthers)
```
mwithi, line 39: according to standard Java naming this should also have the 'P' lowercase?

src/test/java/org/isf/pricesothers/test/Tests.java
```diff
@@ -118,9 +118,9 @@ public void testIoGetPricesOthers()
 		{		
 			id = _setupTestPricesOthers(false);
 			PricesOthers foundPricesOthers = (PricesOthers)jpa.find(PricesOthers.class, id); 
-			ArrayList<PricesOthers> pricesOtherss = otherIoOperation.getOthers();
+			ArrayList<PricesOthers> pricesOthers = otherIoOperation.getOthers();
```
mwithi, line 121: Like this one (see comment above)

src/test/java/org/isf/pricesothers/test/TestPricesOthersContext.java
dbmalkovsky: Good catch; this is a result of an over zealous change and replace action.   Fixed in another [PR](https://github.com/informatici/openhospital-core/pull/115).

dbmalkovsky:
I was referring to the change log type information in some of the java files.  They seem way out of date and sometimes have very little to do with the current state of the file.  For example, see `org.isf.dictype.model.DischargeType` where some of the comments seem to refer to a `disease` and not a `discharge` (see lines 18-22).

PR #117, opened 2020-06-17
Remove unused or duplicate imports

dbmalkovsky:
I left the package imports as they were as I don't know what the policy for the project is with regards to package level imports.


pviotti:
LGTM. Thanks!

PR #118, opened 2020-06-17
Minor JavaDoc updates

dbmalkovsky:
- Add missing `@Deprecated` annotation
- Correct method name and public variable links/code
- Remove 3 `@param` arguments that don't exist in the method signature


pviotti:
Looks good, thank you.

PR #101, opened 2020-05-31
OP-181 Standard Maven Layout

mwithi:
Moved files to comply Standard Maven Layout

**To Check**:
- [x] 32/347 tests fails, mostly related to equal assertions between dates
- [x] suite /openhospital-core/src/test/java/org/isf/test/Tests.java be removed, correct?
- [x] included OP-216

Work item: https://openhospital.atlassian.net/browse/OP-181

PR #120, opened 2020-07-05
[OP-282] [OP-101] Modernize time typing, upgrade Spring and Hibernate

pviotti:
This PR started with the aim of bumping the version of `mysql-java-connector`, but ended up being a revamp and modernization of Java typing for time-related data, and an upgrade of Hibernate and Spring dependencies - thanks also to the work by @dbmalkovsky in PR #203. 

Supersedes: #119 #203


pviotti:
> Indeed, for me it fails 33 tests, on dates comparisons, suggesting to specify the 'serverTimezone' parameter on the JDBC connector... I tried different settings on both server and jdbc URL but it just changes the number of tests failed 😕

I have some failing tests on dates too - I thought it was something about my configuration since the PR validation passed. So let's put this on hold for now.

pviotti:
@mwithi after switching to Java 8 I don't have test failures anymore (which might also be due to the new, faster machine I'm using, if those failures were due to race conditions..). If it's ok there too, feel free to merge it. This Java connector version should be compatible with both MySQL 5.x and 8.x (https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-versions.html)

mwithi:
Thanks @pviotti, I would add more eyes on this, so I add @emecas as reviewer that may give some hints.

pom.xml
```diff
@@ -252,7 +252,7 @@
 		<dependency>
 			<groupId>mysql</groupId>
 			<artifactId>mysql-connector-java</artifactId>
-			<version>5.1.14</version>
+			<version>8.0.21</version>
```
emecas, line 255: Change to this mysql-connector-java version is fair, but the checks on this branch are running against H2

```
/openhospital/core ((dc02440...))
$ cat src/test/resources/database.properties
....
....
jdbc.class=org.h2.Driver
jdbc.url=jdbc:h2:mem:myDb;MODE=MySQL;IGNORECASE=TRUE;DB_CLOSE_DELAY=-1

hibernate.dialect=org.hibernate.dialect.H2Dialect
hibernate.hbm2ddl.auto=update
jdbc.username=root
jdbc.password=root
```

It would be important perform testing against MySQL.

I have tried it locally by pointing to my MySQL installation a couple times and the build got stuck after multiple errors/exceptions: `java.sql.SQLException`, `javax.persistence.PersistenceException`, `javax.persistence.RollbackException`:, this is the last stack trace:

```
[01/Sep/2020 20:29:23] [:] ERROR - >> EXCEPTION: angal.sql.problemsoccurredwiththesqlistruction
[01/Sep/2020 20:29:23] [:] ERROR - >> {}
javax.persistence.RollbackException: Transaction marked as rollbackOnly
        at org.hibernate.ejb.TransactionImpl.commit(TransactionImpl.java:72)
        at org.isf.utils.db.DbJpaUtil.commitTransaction(DbJpaUtil.java:422)
        at org.isf.accounting.test.TestBillPaymentsContext.deleteNews(TestBillPaymentsContext.java:72)
        at org.isf.accounting.test.Tests._restoreContext(Tests.java:775)
        at org.isf.accounting.test.Tests.tearDown(Tests.java:114)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.RunAfters.invokeMethod(RunAfters.java:46)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:33)
        at org.springframework.test.context.junit4.statements.RunAfterTestMethodCallbacks.evaluate(RunAfterTestMethodCallbacks.java:86)
        at org.springframework.test.context.junit4.statements.SpringRepeat.evaluate(SpringRepeat.java:84)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:252)
        at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:94)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.springframework.test.context.junit4.statements.RunBeforeTestClassCallbacks.evaluate(RunBeforeTestClassCallbacks.java:61)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.springframework.test.context.junit4.statements.RunAfterTestClassCallbacks.evaluate(RunAfterTestClassCallbacks.java:70)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
        at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:191)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:141)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:112)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray(ReflectionUtils.java:189)
        at org.apache.maven.surefire.booter.ProviderFactory$ProviderProxy.invoke(ProviderFactory.java:165)
        at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:85)
        at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:115)
        at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:75)
```

My DB could not be up to date, but I just want to make the point, as I've mentioned before, due to the nature of this change, it must be tested against a MySQL database.

mwithi:
Could this PR been already superseded by https://github.com/informatici/openhospital-core/pull/196 ?

In that case checks seem failing for changes needed in the code as well

pom.xml
mwithi: Thanks @emecas, when did you get those exceptions?

pviotti:
> Could this PR been already superseded by #196 ?

no that PR updates to v8.0.16, here I'm updating to 8.0.21 - and the CI passes. So I'll close that PR too (and any other PR Dependabot will open on this dependency - it's not a simple version switch unfortunately)
Later I'll try to see if using MySQL works too.

pom.xml
emecas: Las night, after repointing DB test configuration to MySQL and running mvn clean test

I had not rebuilt MySQL DB in a while, today I got a fresh installation and tried again.

pviotti: ah, some good old time-related error. This one looks like it's not for race condition, but rather due to format and or timezone issues. Seems similar to what we dealt with [here](https://github.com/informatici/openhospital-core/pull/101#issuecomment-637794829).

mwithi:
Again checking on this.

With in-memory DB:
```
Tests run: 430, Failures: 0, Errors: 0, Skipped: 0
```

With MySQL local DB (create_all_en.sql):
```
Tests run: 430, Failures: 27, Errors: 7, Skipped: 0
```

It starts with error:
```
[14/sept./2020 20:55:51] [:] ERROR - HHH000319: Could not get database metadata
java.sql.SQLSyntaxErrorException: Unknown table 'sequences' in information_schema
```
and
```
[14/sept./2020 20:55:51] [:] ERROR - HHH000299: Could not complete schema update
java.sql.SQLSyntaxErrorException: Unknown table 'sequences' in information_schema
```
then (I've set my timezone to +01:00 on my MySQL Server):
```
testBillItemsSets(org.isf.accounting.test.Tests)  Time elapsed: 0.04 sec  <<< FAILURE!
java.lang.AssertionError:
Expecting
  <1984-08-13T01:00:00.000>
to have the same time as:
  <1984-08-14T00:00:00.000>
but actual time is
  <461199600000L>
and expected was:
  <461282400000L>
```
and many errors like this one (it seems it doesn't like how Hibernate generate queries):
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'varchar(255))) like '%testtaxcode%' or lower(patient0_.PAT_SNAME) like '%testtax' at line 1
```
finally a problem if the DB is not empty (how was that test before in-memory db?):
```
[14/sept./2020 20:56:58] [:] ERROR - Duplicate entry 'M' for key 'PRIMARY'
```
Any hint on any above error?

pviotti:
I put the 'help wanted' flag because I don't think I have time (or expertise in ORM frameworks) to solve this anytime soon (and I'm quite puzzled by all these errors - isn't using ORM the whole point of avoiding / abstracting away this kind of stuff?)  
Feel free to fork this branch or start from scratch.

mwithi:
I think as @emecas was stating before, there might be some issues between versions of Hibernate, MySQL connector and MySQL Server combined together... I agree with the label, the mysql connector really should upgrade ASAP

mwithi:
Anyway the application works fine... so it might be only a problem in testing (like context saving and dates formats) but this should be an incentive to solve this kind of problems.

mwithi:
Although with some fixes (https://github.com/informatici/openhospital-core/commit/3eac82855329b2a9aef36a02a31c3e068ee0b701 and https://github.com/informatici/openhospital-core/commit/7cb8851c5c867fbfa337b5aa5de6548ed0451f96) test are running fine on a physical DB on `develop`, so it seems issues on this PR are given exaclty by the new mysql-connector version.

pviotti:
Changing all dates in `Patient` and `Opd` tests and classes from `java.util.Date` to `java.sql.Date` fixed all time-related errors with MySQL (tested with MySQL 5.**7**).  
It may seems a bit drastic, but actually `java.sql.Date`, regardless of all the errors, is the most appropriate data type [for just a date](https://stackoverflow.com/a/2306051) (no timezones, no time, etc) that is persisted to a DATE field in MySQL. It seems like with MySQL 5.7 and the latest java connector they became more strict with dates, thus we had these errors and other warnings (e.g. `SQL Warning Code: 1292, SQLState: 22007`).  

Now there are only 4 errors (not test failures, actual errors) left to fix. They're all due to some incompatibility between some `sql_mode` server side flag and some existing SELECT / GROUP BY clauses - for instance:
```
java.sql.SQLSyntaxErrorException: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'oh.filedicom0_.DM_FILE_ID' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953)
```
```
java.sql.SQLSyntaxErrorException: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'oh.billitems0_.BLI_ID' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeQuery(ClientPreparedStatement.java:1003)
	at com.mchange.v2.c3p0.impl.NewProxyPreparedStatement.executeQuery(NewProxyPreparedStatement.java:116)
	at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.extract(ResultSetReturnImpl.java:79)
```

pviotti:
As I was mentioning above, the remaining 4 errors are due to [ambiguous `select` statements with `group by` clauses](https://stackoverflow.com/a/43699455). There are workarounds to disable those errors in MySQL 5.7, but it would be better to modify those queries, which are:
 - `select b from BillItems b group by b.itemDescription` [here](https://github.com/informatici/openhospital-core/blob/b5c404d794bf617b711604534a76599bc3789a0f/src/main/java/org/isf/accounting/service/AccountingBillItemsIoOperationRepository.java#L41) 
 - `select f from FileDicom f WHERE f.patId = :id group by f.dicomInstanceUID` [here](https://github.com/informatici/openhospital-core/blob/b5c404d794bf617b711604534a76599bc3789a0f/src/main/java/org/isf/dicom/service/DicomIoOperationRepository.java#L41)

If you execute (the SQL equivalent of) those queries in MySQL 5.7, they fail.
I'd appreciate some guidance on how to modify them.

mwithi:
> Changing all dates in `Patient` and `Opd` tests and classes from `java.util.Date` to `java.sql.Date` fixed all time-related errors with MySQL (tested with MySQL 5.**7**).
It may seem a bit drastic, but actually java.sql.Date, regardless of all the errors, is the most appropriate data type for just a date (no timezones, no time, etc) that is persisted to a DATE field in MySQ

It seems also `GregorianCalendar` have been changed, and those are not intented to be persisted as just a date. Moreover, `java.util.Date` in Java does contains time (no timezone, but time yes) so it should be mapped with `java.sql.Timestamp` (that also has no timezone, but time yes). Althought these changes are toward a standardization that I would also suggest as described [here](https://github.com/informatici/openhospital-core/pull/203#issuecomment-692981685), I think this change is not consistent at the moment, because `java.util.Date` has been used many times to handle date AND time, and mapped in the DB (correctly) as DATETIME (apart tests passing, we lose time information where we need it in some places, see comments).

IMHO we should first map the situation as is, with an hard link with the DB as is mapped now (using `@Temporal` annotations for instance), and then perform a JavaType refactoring taking care about each change. Also, all changes in the managers should also be accompained by changes in `gui` and `api`.

> I'd appreciate some guidance on how to modify them.

For the 4 errors about `GROUP BY`, the correct way would be to make explict all the fields required by the Entity (`select b` means `select * from TABLE as b`) but in the case of `FileDicom` for instance, it could be very verbose (although not impossible).
Maybe it could be enough to make explicit only the field in the group by statement (`select b.itemDescription, b.* ...` and `select f.dicomInstanceUID, f.* ...`) but I don't know if JPA mapping to Entity would still work.

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -63,7 +63,7 @@ protected void validateOpd(Opd opd, boolean insert) throws OHDataValidationExcep
 		Disease disease=opd.getDisease();
 		Disease disease2=opd.getDisease2();
 		Disease disease3=opd.getDisease3();
-		if (opd.getDate() == null) opd.setDate(new Date());
+		if (opd.getDate() == null) opd.setDate(new Date((new java.util.Date()).getTime()));
```
mwithi, line 66: this is old record registration date and time, it is mapped with DATETIME in the DB

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -154,7 +154,7 @@ protected void validateOpd(Opd opd, boolean insert) throws OHDataValidationExcep
 	 * @return the list of Opds. It could be <code>null</code>.
 	 * @throws OHServiceException 
 	 */
-	public ArrayList<Opd> getOpd(String diseaseTypeCode,String diseaseCode, GregorianCalendar dateFrom,GregorianCalendar dateTo,int ageFrom, int ageTo,char sex,char newPatient) throws OHServiceException {
+	public ArrayList<Opd> getOpd(String diseaseTypeCode,String diseaseCode, Date dateFrom, Date dateTo,int ageFrom, int ageTo,char sex,char newPatient) throws OHServiceException {
```
mwithi, line 157: This is unclear, because dateFrom and dateTo in queries should always consider time, but in this case are actually compared with visitDate that is actually a Date and persisted as DATE in the DB.

The point is: many time we save correctly the DATEandTIME of an event, but then we must be able to query not caring about time. There are different solution out there, I'm also asking this [here](https://github.com/informatici/openhospital-core/pull/203#issuecomment-692889320).

src/main/java/org/isf/opd/model/Opd.java
```diff
@@ -82,10 +82,10 @@
 
 	@NotNull
 	@Column(name="OPD_DATE_VIS")
-	private GregorianCalendar visitDate;
+	private Date visitDate;
```
mwithi, line 85: this is correct, `visitDate` is mapped as DATE in the DB, so `java.sql.Date` is the correct mapping and I think the field has been introduced to isolate the date of the visit and then query it without be bothered by the time part.

src/main/java/org/isf/opd/model/Opd.java
```diff
@@ -82,10 +82,10 @@
...
         
 	@Column(name="OPD_DATE_NEXT_VIS")
-    private GregorianCalendar nextVisitDate;
+    private Date nextVisitDate;
```
mwithi, line 88: This is not correct, in the DB is a DATETIME and should be saved as that: it's an appointment, with date and time

src/main/java/org/isf/opd/service/OpdIoOperationRepository.java
```diff
@@ -41,13 +42,13 @@
     Integer findMaxProgYear();
 
 	@Query(value = "select max(o.prog_year) from Opd o where o.visitDate >= :dateFrom and o.visitDate < :dateTo")
-    Integer findMaxProgYearWhereDateBetween(@Param("dateFrom") GregorianCalendar dateFrom, @Param("dateTo") GregorianCalendar dateTo);
+    Integer findMaxProgYearWhereDateBetween(@Param("dateFrom") Date dateFrom, @Param("dateTo") Date dateTo);
```
mwithi, line 45: correct, dateFrom is compared with `o.visitDate` so `java.sql.Date == DATE`

src/main/java/org/isf/opd/service/OpdIoOperationRepository.java
```diff
@@ -41,13 +42,13 @@
...
 
     List<Opd> findTop1ByPatient_CodeOrderByDateDesc(Integer code);
 	
 	@Query("select o from Opd o where o.prog_year = :prog_year")
     List<Opd> findByProgYear(@Param("prog_year") Integer prog_year);
 	
 	@Query(value = "select op from Opd op where op.prog_year = :prog_year and op.visitDate >= :dateVisitFrom and op.visitDate < :dateVisitTo")
-    List<Opd> findByProgYearAndVisitDateBetween(@Param("prog_year") Integer prog_year, @Param("dateVisitFrom") GregorianCalendar dateVisitFrom, @Param("dateVisitTo") GregorianCalendar dateVisitTo);
+    List<Opd> findByProgYearAndVisitDateBetween(@Param("prog_year") Integer prog_year, @Param("dateVisitFrom") Date dateVisitFrom, @Param("dateVisitTo") Date dateVisitTo);
```
mwithi, line 53: correct, dateFrom is compared with o.visitDate so java.sql.Date == DATE

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryCustom.java
```diff
@@ -25,10 +25,11 @@
...
 import java.util.GregorianCalendar;
 import java.util.List;
+import java.sql.Date;
 
 public interface OpdIoOperationRepositoryCustom {
 
-	List<Opd> findAllOpdWhereParams(String diseaseTypeCode, String diseaseCode, GregorianCalendar dateFrom,
-									GregorianCalendar dateTo, int ageFrom, int ageTo, char sex, char newPatient);
+	List<Opd> findAllOpdWhereParams(String diseaseTypeCode, String diseaseCode, Date dateFrom,
+									Date dateTo, int ageFrom, int ageTo, char sex, char newPatient);
```
mwithi, line 33: correct, `dateFrom` is (in the other class) compared with `o.visitDate` so `java.sql.Date == DATE`

src/main/java/org/isf/patient/manager/PatientBrowserManager.java
```diff
@@ -391,7 +391,7 @@ protected void validate(Patient patient) throws OHDataValidationException{
     }
 
     private boolean checkAge(Patient patient) {
-	    Date now = new Date();
+	    Date now = new Date((new java.util.Date()).getTime());
```
mwithi, line 394: this is not changing, maybe you mean `Date now = new Date((new java.sql.Date()).getTime());` ?

src/main/java/org/isf/patient/model/Patient.java
```diff
@@ -259,7 +259,7 @@ public Patient(String firstName, String secondName, Date birthDate, int age, Str
 		this.firstName = firstName;
 		this.secondName = secondName;
 		this.name = this.firstName + " " + this.secondName;
-		this.birthDate = birthDate;
+		this.birthDate = new Date(birthDate.getTime());
```
mwithi, line 262: correct

dbmalkovsky:
Please see [comment ](https://github.com/informatici/openhospital-core/pull/203#issuecomment-695028951)in Spring 5 PR where I tried using MySQL and the code changes that were made there.   Hint: got it to work.

pviotti:
*[An update/self-note about the time ORM issues]*
It seems that the best way to [map time data in Java 8](https://www.oracle.com/technical-resources/articles/java/jf14-date-time.html) is using `java.time` classes (as @mwithi was mentioning [here](https://github.com/informatici/openhospital-core/pull/203#issuecomment-692981685)). In particular, with these mappings (MySQL type --> Java type):
- `Date` --> `java.time.LocalDate`
- `Datetime` --> `java.time.LocalDateTime`

Support for the `java.time` API in JPA came with v2.2. Currently we use JPA 2.0 through the Hibernate implementation (`hibernate-jpa-2.0-api` transitive dependency of `hybernate-core`). So to do the [best / most future-proof thing](https://www.baeldung.com/jpa-java-time), we should **[A]** upgrade Hibernate to [v5.3](http://hibernate.org/orm/releases/5.3/) (the first version to support JPA 2.2) (or use `java.persistence-api` v2.2 and manually exclude Hibernate JPA implementation in pom.xml (?)).
Alternatively, **[B]**, according to [this article](https://thorben-janssen.com/persist-localdate-localdatetime-jpa/), we can have some converters for `java.sql.*` classes.

So far I've tried **[A]** but it wouldn't work by simply bumping Hibernate (some weird class loading errors...).

dbmalkovsky:
FWIW, in the Spring 5 testing (other PR), I bumped Hibernate up to `5.4.21.Final` so it seems to work if other method name changes are made.

mwithi:
Before my comment on the ther PR as observed @pviotti, I was carrefully reading and also whatching the Thorben Janssen's video, that's why I was suggesting using `@Temporal`, as the way it should have been done in an application where `java.util.Date` and `java.util.(Gregorian)Calendar` have been used instead of `java.sql.*`. 

In that way, by using `@Temporal(TemporalType.DATE)` where fields have been mapped as DATE in the DB (only three cases in my **Analysis**), and `@Temporal(TemporalType.DATETIME)` for all other cases, we should be able to adjust everything and also better annotating Entities, without changing one line of code, I hope... 🤞

The long-term solution, instead, is to adopt java.time.*, but this requires deep testing in the depending code (gui and api), but at least the above annotation should not bother if present or not, I think... 🙄

Question to @pviotti: why you adopted **java.time.*** and left some **java.sql.*** in few cases?

pviotti:
> we should be able to adjust everything and also better annotating Entities, without changing one line of code, I hope... 

Sometime changing lines of code it's just the best thing to do :) Getting rid of `*Calendar`, `Joda-Time` and `java.util.Date` in favor of `java.time` API seems the most reasonable thing to do in the long term. 

> why you adopted java.time.* and left some java.sql.* in few cases?

I just started going through the modules, one by one. The first was `opd`, the rest will follow. By the end of it also `java.sql.*` should be gone.

mwithi:
> Sometime changing lines of code it's just the best thing to do :) Getting rid of `*Calendar`, `Joda-Time` and `java.util.Date` in favor of `java.time` API seems the most reasonable thing to do in the long term.

I totally agree. Just as reminder: there are cases where we need DateTime in the DB and in the Entity, but some queries need to compare only by Date (without Time). This at the moment is achieved with explicit JPQL queries containing `DATE(field)` operator.

> I just started going through the modules, one by one. The first was `opd`, the rest will follow. By the end of it also `java.sql.*` should be gone.

Ack

So, to better understand: are developments on this #120 overlapping the ones in #203?

pviotti:
> This at the moment is achieved with explicit JPQL queries containing DATE(field) operator. I hope it is supported by most ORM and DMBS.

Ok. This is calling for more tests.

> So, to better understand: are developments on this #120 overlapping the ones in #203?

Strictly speaking not much, but the root problems we found are similar, so there are some overlapping changes - I see there are ~3 date typing changes, the creation of `DateCalendarConverter` and the version bump of Hibernate. I guess once that PR is merged I'll have to rebase and undo some of those changes.

Btw I've created OP-282 to track this.

dbmalkovsky:
The other way to do it is to make all the `java.time` changes and then I can apply the Spring 5 updates here and see what is left.


> > This at the moment is achieved with explicit JPQL queries containing DATE(field) operator. I hope it is supported by most ORM and DMBS.
> 
> Ok. This is calling for more tests.
> 
I believe there are many such tests given my work on #203 and adding additional Junit tests to the `accounting` package for example.

pviotti:
> The other way to do it is to make all the java.time changes and then I can apply the Spring 5 updates here and see what is left.

yeah if we somehow serialize these two changes we can perhaps avoid some complex and error-prone rebase. Btw this will take some time, I'm trying to do one module a day. :construction_worker:

dbmalkovsky:
FWIW, when I run this branch against H2 locally I get:
      **Tests failed: 9, passed: 401 ignored:23 of 432 tests**

All the ignored tests are in the **dicom.test** folder as they failed during test setup.

I got the ignored tests to run by changing in org.isf.dicom.test.TestDicom the line:

`DateTimeFormatter formatter = DateTimeFormatter.ofPattern("EEE MMM dd hh:mm:ss z yyyy", new Locale("en"));`

to read:

`DateTimeFormatter formatter = DateTimeFormatter.ofPattern("EEE MMM dd HH:mm:ss z yyyy", new Locale("en"));`

Note the subtle change from **hh** to **HH** in the pattern.

Now rerunning the tests:

**Tests failed: 12, passed: 420 of 432 tests**

Hope this helps someone.

pviotti:
Thanks David. Indeed there are still some kinks to iron out. 
I had removed `hibernate.hbm2ddl.auto` property because it would fail saying `Unrecognized legacy hibernate.hbm2ddl.auto value : false`. Now I pushed a better fix, replacing `false` with `none` as default.

With the latest commits, 8 tests are failing with H2, 12 with MySQL 5.7.
For MySQL most of failures are the non-deterministic `group by` mentioned above (still not quite sure how to fix them..).
For H2 there are some assertion failures and some exception like `org.hibernate.PersistentObjectException: detached entity passed to persist: org.isf.priceslist.model.PriceList`.
It will take me some more days to fix all of these errors, so if anyone want to have a shot at them, please do - I'll integrate your contributions :)

pviotti:
Now all tests pass for H2, and for MySQL only those 4 tests with `group by` remain to be fixed - suggestions are welcome.
(btw there's some race condition in dicom tests: sometimes, some of them fail - see the PR validation build, it failed for java 8 on a dicom test which passed for java 11)

mwithi:
> only those 4 tests with `group by` remain to be fixed - suggestions are welcome

👆 https://github.com/informatici/openhospital-core/pull/120#issuecomment-69456415

> build a test fails only for java 8

Strange, it seems a NPE ([L1709](https://github.com/informatici/openhospital-core/pull/120/checks?check_run_id=1179125442#step:4:1709))... 🤔 I will check it tomorrow!

Great work, thanks!

pviotti:
About the remaining two `group by` queries failing with MySQL 5.7, as mentioned [in the comment above](https://github.com/informatici/openhospital-core/pull/120#issuecomment-694516833), we could add `ANY_VALUE(column_name)` for every column of the tables (and it's a lot of columns..) as @mwithi was suggesting.  
But I was thinking a) if those queries make any sense at all and b) if they are actually used.  
I don't think they make much sense - for example, for what reason would we want to group by description and take whatever of the resulting bill items? I can't think of a use case for this.  
Where are those two queries used? 
 - the Dicom one is only used in `SqlDicomManager`, which is not used normally (we use  `FileSystemDicomManager` by default, which btw doesn't do any grouping in `loadPatientFiles`..)
 - the BillItem query is only used [here](https://github.com/informatici/openhospital-gui/blob/develop/src/main/java/org/isf/accounting/gui/BillBrowser.java#L832-L833) in the GUI project (and nowhere in the API project), the result of the invocation is never used, and there's a comment like `// FIXME: why need to call getDistinctItems() here ?????`

In short, I suspect those queries should be changed or removed, but I don't have enough context to do so. I'll just tidy up the existing commits (squash together the commits on the same modules, remove useless ones), and switch this PR ready for review.

mwithi:
> * the Dicom one is only used in `SqlDicomManager`, which is not used normally (we use  `FileSystemDicomManager` by default, which btw doesn't do any grouping in `loadPatientFiles`..)

This should be revised, there were three hands on that method, included mine...

> * the BillItem query is only used here in the GUI project (and nowhere in the API project), the result of the invocation is never used, and there's a comment like // FIXME: why need to call getDistinctItems() here ?????

I'm quite sure it's about a feature that get lost with subsequent refactoring... see https://github.com/informatici/openhospital-core/commit/d207993b0c11b62e20efe3b3a8985a6ebdac62ad 

In short: for both we need a little of patience to understand

dbmalkovsky:
I took the latest version of this code branch and applied all the Spring 5 (5.2.9.RELEASE) and Hibernate (5.4.21.Final) changes from the other [PR 203](https://github.com/informatici/openhospital-core/pull/203).

The database is MySQL 8.0.21 and ran the tests with both JRE8 and JRE11 and they were both completely clean.

These are the changes to the `database.properties` file:

<pre>
     jdbc.class=com.mysql.cj.jdbc.Driver
     hibernate.dialect=org.hibernate.dialect.MySQL8Dialect
     jdbc.url=jdbc:mysql://localhost:3306/oh?serverTimezone=US/Eastern
</pre>

With this latest code I no longer need:

<pre>
   SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));
</pre>

The patch file is here (the extension was changed from `patch` to `log` so it could be attached): [Spring_5_Hibernate.log](https://github.com/informatici/openhospital-core/files/5307276/Spring_5_Hibernate.log),

Hope someone finds this useful.

pviotti:
thanks @dbmalkovsky! I've applied your patch in [this commit](https://github.com/informatici/openhospital-core/pull/120/commits/55ead3746232a32b85f767dd0655829c88e8f382) (and squashed some of my commits together).

dbmalkovsky:
@pviotti The commit looks good and the tests all passed.  Sounds like we're on the right path :smile:

Thanks for all you work on this.

If (when) this PR is committed we can close the draft for the Spring 5 update.

dbmalkovsky:
@pviotti @mwithi What version of MySQL is the next release going to support?  Version 8?  If so I think the `hibernate.dialect` default parameter in the `applicationContext.xml` file should be changed to: `org.hibernate.dialect.MySQL8Dialect`

mwithi:
This work is **_epic_**! Thanks a lot @pviotti!

The problem now is that this PR cannot be merged until a similar work will be done on gui, that is also quite big and we don't have unit tests on that side, so it will take time.

For next version will be released compliant with MySQL 5.7.30, so I think we can update hibernate settings just now, as well as the docker mysql image https://github.com/informatici/openhospital-core/blob/develop/Dockerfile-ohdb#L1

pviotti:
> What version of MySQL is the next release going to support?

With MySQL v5.6 all tests pass. Version 5.7 has those `group by` queries issues mentioned above (but works without the `mysql-java-connector` upgraded in this PR). I haven't tested version 8.0 (it requires fiddling with the Docker file in the root folder because some data loading parameters have changed).

> The problem now is that this PR cannot be merged until a similar work will be done on gui

IMO they don't have to be merged *at the same time*: we use `develop` branch right because there things don't have to be stable all the time, API-wise - if developers of GUI project need a "stable" (more compatible) version of Core, they'll just have to avoid doing `mvn install` of Core's develop branch, or, if they do by mistake, do `git checkout v1.10.0 && mvn install` to restore a compatible version with existing Gui code.  
I think that if we want this to be part of the next release, the sooner we merge it, the more we can test it.  
However, I don't know how much work it's required to adapt Gui to the changes in time typing - it might be just a find/replace all, or not.

mwithi:
> I haven't tested version 8.0 (it requires fiddling with the Docker file in the root folder because some data loading parameters have changed).
If I remember well, only the `load data infile` has been disabled by default and it is needed for the CSV loading from scripts. I don't know if to test with MySQL8 rather than start that topic about file-based DB... meanwhile @mizzioisf informed me that the actual scripts and code in `develop` (pre-release) are seemlessly working with mariadb 10.3 and java14.

> IMO they don't have to be merged at the same time: we use develop branch right because there things don't have to be stable all the time

Ok, but if we already know it is much breaking, we usually do siblings PRs on other repos, as I'm suggesting [here](https://github.com/informatici/openhospital-gui/pull/220#issuecomment-702418416). 

> I think that if we want this to be part of the next release, the sooner we merge it, the more we can test it.
However, I don't know how much work it's required to adapt Gui to the changes in time typing - it might be just a find/replace all, or not.

Exactly, honestly I would not merge something that brakes the stack for an undefined span of time... since we are going to release soon from `develop`, and we must include Gui as well, I would avoid that unstability right now.

pviotti:
@mwithi once again, I kind of understand but not fully agree with this repotting of branches. This way:
 - the original context of the PR with its comments and contributions got one more layer of indirection 
 - 32 [carefully tidied up] commits got squashed into one very large commit, so: a) it will be impossible to revert single problematic changes and b) the history of commits got lost.

What are the pros of this? Not sure, maybe now admins of this repo have "full ownership" of the changes in this branch.

Also, in general I would _strongly_ discourage this behavior, as some contributors (not me:) might find confusing and frustrating that their PR is hijacked to some secondary branch without notice.

mwithi:
@pviotti sorry, the intention is not obviously to hijack anything but just to buffer the developments in order to sync them with other similar developments on other repos, until all of them are ready to be merged into 'develop' all at once.

IMHO the external contributor cannot decide where his/her development must go in first sight, I think it's something allowed by GitHub for a reason and, as you said, it _has not_ been without notice, I've early communicated this operation [here](https://github.com/informatici/openhospital-core/pull/120#issuecomment-702404227), [here](https://github.com/informatici/openhospital-core/pull/120#issuecomment-702435781) and [here](https://github.com/informatici/openhospital-gui/pull/220#issuecomment-702418416) because the same operation make sense if all contributors are told about it.

Indeed, we used the same way for other big issues, I think with quite accettable understanding, the aim is not to have potting branches but just to sync big issues involving different repos in order to be treated as one. 

I don't know if it's the best way though, other proposals are welcomed.

src/main/java/org/isf/accounting/model/Bill.java
```diff
@@ -74,18 +74,18 @@ public class Bill extends Auditable<String> implements Comparable<Bill>
...
-	private GregorianCalendar update;
+	@Column(name="BLL_UPDATE")		// SQL type: datetime
+	private LocalDateTime update;
 
 	@NotNull
 	@Column(name="BLL_IS_LST")
 	private boolean isList;
 	
-	@ManyToOne
+	@ManyToOne(cascade = {CascadeType.MERGE})
```
mwithi, line 88: @pviotti we are going to remove this cascade ([OP-990](https://openhospital.atlassian.net/browse/OP-990)). Did you remember why it was introduced?

pviotti: :thinking: most likely it was introduced to avoid compilation errors with the mysql-related version bumps of this PR - I can't remember any other possible reason.
It changes the way Patient and Bill [are persisted when they're handled together](https://www.baeldung.com/jpa-cascade-types#3-cascadetypemerge), and it kinda make sense - but if necessary you can try to remove it.

mwithi: Yes we are removing them, because the child entity is not created together with the parent, they have independent life (see OP-990). Thanks!

PR #132, opened 2020-07-15
Op 128 sql to jpa

tomek39856:
2 tests in ward still failing, one todo in bill repository, regarding to ward I'll appreciate help


mwithi:
Superseded by https://github.com/informatici/openhospital-ui/pull/32

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -215,7 +217,13 @@ protected boolean updateStockWardQuantity(
             if(medicalWardTo != null) {
             	repository.updateInQuantity(Math.abs(qty), wardTo, medical, lot);
             } else {
-                repository.insertMedicalWard(wardTo, medical, Math.abs(qty), lot);
+				MedicalWard medicalWard = new MedicalWard();
+				medicalWard.setWard(movement.getWard());
```
mwithi: Here is the problem, change to:
medicalWard.setWard(movement.getWardTo());

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -215,7 +217,13 @@ protected boolean updateStockWardQuantity(
...
+				medicalWard.setMedical(movement.getMedical());
+				medicalWard.setInQuantity((float)Math.abs(qty));
+				medicalWard.setOutQuantity(0.0f);
+				medicalWard.setLot(movement.getLot());
+                repository.save(medicalWard);
             }
             repository.updateOutQuantity(Math.abs(qty), ward, medical, lot);
```
mwithi: Remove this line, no need to update the quantity after the previous .save(medicalWard)

tomek39856:
Now all tests are passing, there are still some new queries which came from develop which needs rewriting but it can be done fater merge. @emecas  thanks for help!

src/test/java/org/isf/ward/test/Tests.java
```diff
@@ -8,354 +8,219 @@
...
+		wardIoOperationRepository.deleteAll();
     }
     
     @AfterClass
-    public static void tearDownClass() throws OHException 
-    {    	
+    public static void tearDownClass() throws OHException {
     	return;
     }
-
```
mwithi, line 69: This part has been changed compared to others, for consistency with other tests we need all these methods:

```
    @BeforeClass
    public static void setUpClass() {
		jpa = new DbJpaUtil();
    	testWard = new TestWard();
    	testWardContext = new TestWardContext();
    }

    @Before
    public void setUp() throws OHException {
    	jpa.open();
        
        _saveContext();
		
		return;
    }
    
    @After
    public void tearDown() throws Exception 
    {
        _restoreContext();   
        
        jpa.flush();
        jpa.close();
                
        return;
    }
    
    @AfterClass
    public static void tearDownClass() throws OHException {
    	
    	testWard = null;
    	testWardContext = null;
    	
    	return;
    }

...

private void _saveContext() throws OHException 
    {	
		testWardContext.saveAll(jpa);
		
        return;
    }
	
	private void _restoreContext() throws OHException 
    {
		testWardContext.deleteNews(jpa);
        
        return;
    }
```

src/test/java/org/isf/medicalstockward/test/Tests.java
```diff
@@ -333,6 +338,7 @@ public void testIoNewMovementWard()
 			Double quantity = (double) medicalIoOperation.getCurrentQuantityInWard(wardTo, medical);
 			
 			_checkMovementWardIntoDb(movementWard.getCode());
+			jpa.commitTransaction();
```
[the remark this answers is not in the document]
mwithi: This is true, but if I run all tests, and not only this suite, I still get the same constraint violation, because of dirty context produced by other tests (although the in-memory DB).

See https://github.com/informatici/openhospital-core/pull/132#discussion_r456689646

mwithi:
> Since the values returned by `getCurrentQuantityInWard` for each ward-code parameter available, are not what someone would expect after performing the operation:
> 
> `result = medicalIoOperation.newMovementWard(movementWard);`

Actually it is, only there was a problem on the updateStockWardQuantity method, see comments above:

- https://github.com/informatici/openhospital-core/pull/132#pullrequestreview-450966244
- https://github.com/informatici/openhospital-core/pull/132#discussion_r456681875

src/main/java/org/isf/accounting/service/AccountingIoOperations.java
```diff
@@ -370,7 +370,7 @@ public int compare(Bill o1, Bill o2) {
 	 * @throws OHServiceException
 	 */
 	public List<BillItems> getDistictsBillItems() throws OHServiceException {
-		return billItemsRepository.findAll(); // TODO: restore groupping
+		return billItemsRepository.findAllGroupByDescription(); // TODO: restore groupping
```
mwithi, line 373: remove the TODO comment

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -217,13 +217,7 @@ protected boolean updateStockWardQuantity(
...
             	repository.updateInQuantity(Math.abs(qty), wardTo, medical, lot);
             } else {
-				MedicalWard medicalWard = new MedicalWard();
-				medicalWard.setWard(movement.getWard());
-				medicalWard.setMedical(movement.getMedical());
-				medicalWard.setInQuantity((float)Math.abs(qty));
-				medicalWard.setOutQuantity(0.0f);
-				medicalWard.setLot(movement.getLot());
-                repository.save(medicalWard);
+                repository.insertMedicalWard(wardTo, medical, Math.abs(qty), lot);
```
mwithi: Is it not the equivalent of save(medicalWard) with correct parameters set?

mwithi:
> Now all tests are passing, there are still some new queries which came from develop which needs rewriting but it can be done fater merge. @emecas thanks for help!

I can see that the two tests are still failing https://github.com/informatici/openhospital-core/runs/883616070#step:5:2499
The strang thing is that due to https://github.com/informatici/openhospital-core/pull/132#discussion_r456689646 all subsequent tests fail for me, but not in CI

mwithi:
> Actually, the CI tests are passing now: https://github.com/informatici/openhospital-core/pull/132/checks?check_run_id=884438366

I see... I was looking at the wrong page. Ok, then to merge as is, but I suggest to put some reminders in the code for the things spotted here, wdyt @tomek39856 ?

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
tomek39856: Should be but reverting to insertMedicalWard made tests working again. I'll add a comment to investigate this and restore save

tomek39856: And I see that you switched the ui so two questions:
1. Should I care about old swing based ui? I have quick fix for menu issue after jpa changes in separate pull request
2. How to test changes now? 

And we definitely need to test everything quite deeply. There are many contributors and whole OP-128 is quite big change

mwithi: 1. yes, until the "ui" is not yet ready
2. if you mean test from the "gui", I think smoke-tests and end-users is the way

I agree, I suggest all contributors to fork also "gui" and to have the application up and running for both regression tests and help on the new "ui" design

src/test/java/org/isf/ward/test/Tests.java
tomek39856: Are you sure we need this? Now we start fresh in memory db each time we run tests so context stuff is not needed anymore. Repository is cleaned up in line 35 
public void setUp() {
		wardIoOperationRepository.deleteAll();
so tests are independent, and now we have much less code to maintain (just one line to cleanup db before test)

tomek39856:
Fixed ward stuff, I also added todos and we just need to rewrite queries in ward later, I think you can try merge it to develop and let's check if it blow up...

src/test/java/org/isf/ward/test/Tests.java
mwithi: I know... but so why that context stuff is maintained in all other classes?

When I run all the tests (not only `Ward `or `MedicaStockWard ` packages) I get the error once reached `testIoNewMovementWard`:

```
ERROR - Violación de indice de Unicidad ó Clave primaria: "PRIMARY_KEY_28 ON PUBLIC.WARD(WRD_ID_A) VALUES (CAST('X' AS VARCHAR_IGNORECASE), 59)"
Unique index or primary key violation: "PRIMARY_KEY_28 ON PUBLIC.WARD(WRD_ID_A) VALUES (CAST('X' AS VARCHAR_IGNORECASE), 59)"; SQL statement:
insert into WARD (WRD_ACTIVE, WRD_CREATED_BY, WRD_CREATED_DATE, WRD_LAST_MODIFIED_BY, WRD_LAST_MODIFIED_DATE, WRD_NBEDS, WRD_NAME, WRD_NDOC, WRD_EMAIL, WRD_FAX, WRD_IS_FEMALE, WRD_IS_MALE, WRD_IS_PHARMACY, WRD_LOCK, WRD_NQUA_NURS, WRD_TELE, WRD_ID_A) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) [23505-190]
[18/jul/2020 23:35:51] [:] ERROR - Violación de indice de Unicidad ó Clave primaria: "PRIMARY_KEY_28 ON PUBLIC.WARD(WRD_ID_A) VALUES (CAST('X' AS VARCHAR_IGNORECASE), 59)"
Unique index or primary key violation: "PRIMARY_KEY_28 ON PUBLIC.WARD(WRD_ID_A) VALUES (CAST('X' AS VARCHAR_IGNORECASE), 59)"; SQL statement:
insert into WARD (WRD_ACTIVE, WRD_CREATED_BY, WRD_CREATED_DATE, WRD_LAST_MODIFIED_BY, WRD_LAST_MODIFIED_DATE, WRD_NBEDS, WRD_NAME, WRD_NDOC, WRD_EMAIL, WRD_FAX, WRD_IS_FEMALE, WRD_IS_MALE, WRD_IS_PHARMACY, WRD_LOCK, WRD_NQUA_NURS, WRD_TELE, WRD_ID_A) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) [23505-190]
javax.persistence.PersistenceException: org.hibernate.exception.ConstraintViolationException: could not execute statement
```

The strange thing is that in CI it doesn't happen...

mwithi: ERRATA CORRIGE: it fails only in my Eclipse (Idkw) but in Maven build all goes smooth.

tomek39856: Ok than I can restore it and adjust tests setup in separate pull request what do you think? Of course if you agree with this kind of change but in my opinion possibility to remove this context stuff from tests is one of benefits having in memory db for tests

PR #136, opened 2020-07-20
Update import statements

dbmalkovsky:
Changes:
- Remove unused imports
- Remove duplicated imports
- Expand package imports

Is there a standard ordering of imports for this project?  It is not discernable looking at the files.

If not I'd suggest an order like:

- static imports 
- java imports
- javax imports
- org imports
- com imports
- anything else

and no package level imports


dbmalkovsky:
I don't see a way to define it in `.editorconfig` along with lots of other things.  That is why many projects provide code style files (along with code temples and boiler plate code) that can be imported into either Eclipse or Intellij.  Another project that I'm heavily involved with does exactly that; see this [directory](https://github.com/flowable/flowable-engine/tree/master/ide-settings).

pviotti:
> I don't see a way to define it in .editorconfig along with lots of other things. That is why many projects provide code style files (along with code temples and boiler plate code) that can be imported into either Eclipse or Intellij. Another project that I'm heavily involved with does exactly that; see this directory.

Good point, feel free to submit another PR to add them.
This one looks good to me, thanks.

PR #138, opened 2020-07-25
OP-239 Verify and Complete JUnit Tests coverage for 'dicom' package

fjevictoriano:
Hi @mwithi This a draft PR. I added more tests for the dicom module. The package has 90% test coverage now. I will continue to add more tests for the others module but before I would like know if I'm going in the right direction with the tests, package structure for resources ect..


fjevictoriano:
> This branch is ignoring multiples changes/fixes ([OP-229](https://openhospital.atlassian.net/browse/OP-229) [OP-136](https://openhospital.atlassian.net/browse/OP-136) [OP-128](https://openhospital.atlassian.net/browse/OP-128) [OP-73](https://openhospital.atlassian.net/browse/OP-73) [OP-171](https://openhospital.atlassian.net/browse/OP-171)) that are already merged on develop branch since V1.10.0 ([ff5c1ed](https://github.com/informatici/openhospital-core/commit/ff5c1eddf431d76a1beea43268d6416fc5e824a5))
> 
> @fjevictoriano IMO you should start coverage improvements from a more recent version, if you rebase to ff5c1eddfw, then coverage changes would make more sense from that point and avoid conflicts.

Hi @emecas I did a rebase as you mentioned and resolved a conflict now I'm having an issue with the build cause one of the tests  is using a class from awt which requires access to  X server. I think with should use -Djava.awt.headless=true when running maven in the CI but I'm not sure.

emecas:
> 
> 
> > This branch is ignoring multiples changes/fixes ([OP-229](https://openhospital.atlassian.net/browse/OP-229) [OP-136](https://openhospital.atlassian.net/browse/OP-136) [OP-128](https://openhospital.atlassian.net/browse/OP-128) [OP-73](https://openhospital.atlassian.net/browse/OP-73) [OP-171](https://openhospital.atlassian.net/browse/OP-171)) that are already merged on develop branch since V1.10.0 ([ff5c1ed](https://github.com/informatici/openhospital-core/commit/ff5c1eddf431d76a1beea43268d6416fc5e824a5))
> > @fjevictoriano IMO you should start coverage improvements from a more recent version, if you rebase to ff5c1eddfw, then coverage changes would make more sense from that point and avoid conflicts.
> 
> Hi @emecas I did a rebase as you mentioned and resolved a conflict now I'm having an issue with the build cause one of the tests is using a class from awt which requires access to X server. I think with should use -Djava.awt.headless=true when running maven in the CI but I'm not sure.

Thank you @fjevictoriano 

The build now is faster and using in memory database, but unfortunately, I am having a local issue, it is getting stuck at same point,  after executing 307 tests of 396.

This is the message:

    [01/Aug/2020 14:57:45] [:] ERROR - HHH000099: an assertion failure occured (this may indicate a bug in Hibernate, but is more likely due to unsafe use of the session): org.hibernate.AssertionFailure: null id in org.isf.medicals.model.Medical entry (don't flush the Session after an exception occurs)

image

image

Are you experiencing a similar issue locally? I would like to discard first that it is not related to my environment .

fjevictoriano:
No it runs fine in my local. Maybe you need to rebuild the DB?

emecas:
> 
> No it runs fine in my local. Maybe you need to rebuild the DB or its not need it due to the in memory db??

Yes, with the in-memory db, db gets "rebuild" before starting testing, I am still having same issue, comparing to #156, that is completing the build successfully,  but I have not noted a particular change for fixing my local, but a combination of changes, about java version and dates, that does not make much sense to bring them  here.

In addition to the pom.xml, I have been mainly focused on those 2 files, but i did not note any particular thing that i can change for fixing it

```
$
openhospital/core ((36e3901...))
$  git diff ffda9dd --name-only
...
src/main/java/org/isf/medicalstock/service/LotIoOperationRepository.java
src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
...
```

The weird thing is that all tests  under `org.isf.medicalstock.test` pass if you run them independently. So the DB problem could be related to a bad state left for another package's test.  

@pviotti  @tomek39856 any advise how i could restart the DB during the test execution?

tomek39856:
Hi

Problem is probably caused because data polluted by other test executed earlier. We run everything on "dirty context" so db is fresh before each test execution but niot between the tests. I'll try to find out what is not being cleaned up before test exec tommorow or during weekend

tomek39856:
You can also experiment with @Before method and try to inject all necessary repositories and clean them

pviotti:
> So the DB problem could be related to a bad state left for another package's test.

Yes it could be that. I get errors of primary key and other constraints violations, and it hangs.

> any advise how i could restart the DB during the test execution?

From a quick search there appears to be a number of solutions: using [Beans](https://www.baeldung.com/spring-jpa-test-in-memory-database) or simply [truncating](https://stackoverflow.com/a/44604779) tables during methods marked as `@Before` or `@After`. I would have expected that's what those `jpa.flush()` and `jpa.close()` in the tearDown method did.

emecas:
Thanks @pviotti @tomek39856

@fjevictoriano I found the problem, or at least the one is making the build fail locally, you have to set this property to false: `AUTOMATICLOT_IN=false` on  `rsc/generalData.properties` 

This is the coverage summary:

image


```
Results :

Tests run: 394, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time:  41.267 s
[INFO] Finished at: 2020-08-05T23:03:06-04:00

```

pviotti:
> I found the problem, or at least the one is making the build fail locally, you have to set this property to false: AUTOMATICLOT_IN=false on rsc/generalData.properties

It works here too, good catch. (I love it when everything is sorted out by some undocumented flag :smile: )
(btw we should move the files in `/rsc/` to `/src/{main,test}/resources`)

mwithi:
Hi guys, I'm trying to understand the problem here, what I find is that:
- the code as is, it fails a number of tests
- if set AUTOMATICLOT_IN=no as found out by @emecas all tests pass
- the parameter AUTOMATICLOT_IN has been introduced recently and not yet documented (OP-153); it should be tested with both yes|no values, either by modifying the current tests and/or by writing new adhoc tests
- the OP-147 it's probably too big, but what is really needed to be covered are `manager` and `service` packages, tests can be run also with `usingSet` parameter which should cover `model` setters too.

@fjevictoriano feel free to open OP-147 sub-tasks on Jira, in order to address developments on each package, like this one for the DICOM package can be renamed into OP-239 (already created) and confirm this PR (draft->open)

fjevictoriano:
@mwithi should I rename my local branch OP-147  to OP-239 and make a new pull request ?

emecas:
> 
> @mwithi should I rename my local branch [OP-147](https://openhospital.atlassian.net/browse/OP-147) to [OP-239](https://openhospital.atlassian.net/browse/OP-239) and make a new pull request ?

@fjevictoriano  I don't think you need rename the branch or a new PR, you can keep working here with the new limited scope about improving testing coverage for 'dicom' package, what seems is already accomplished here, then as @mwithi  proposed we can create new subtasks for other particular modules under OP-147. thank you

mwithi:
> @mwithi should I rename my local branch [OP-147](https://openhospital.atlassian.net/browse/OP-147) to [OP-239](https://openhospital.atlassian.net/browse/OP-239) and make a new pull request ?

I confirm what @emecas says, no need to rename, but the work **seems** completed, because I'm still figuring out why on `develop` branch with don't have problems about that **AUTOMATICLOT_IN** parameters while in this PR yes (fourh point in my [bullet list above](https://github.com/informatici/openhospital-core/pull/138#issuecomment-675581603)). 

So I would first understand this before to merge it.

fjevictoriano:
hi @mwithi I was checking and it seems that **AUTOMATICLOT_IN** is **false** by default. I turned it on by calling the GeneralData.getGeneralData() on my tests and when I did that other tests started to fail or specifically tests from **org.isf.medicalstock.test** I think maybe because they were written without considering **AUTOMATICLOT_IN** = true?

mwithi:
Hi @fjevictoriano , that's it indeed! I didn't notice you was calling it [here](https://github.com/informatici/openhospital-core/pull/138/files#diff-2ffc3f2420df655fd4118b3590d3075aR336).

As I said in the [bullet list above](https://github.com/informatici/openhospital-core/pull/138#issuecomment-675581603) (third point) the AUTOMATICLOT_IN parameter is new and should be tested with both values. I'll open a new Jira issue for this. 

For this PR instead, do you really need to instanciate GeneralData?

fjevictoriano:
> Hi @fjevictoriano , that's it indeed! I didn't notice you was calling it [here](https://github.com/informatici/openhospital-core/pull/138/files#diff-2ffc3f2420df655fd4118b3590d3075aR336).
> 
> As I said in the [bullet list above](https://github.com/informatici/openhospital-core/pull/138#issuecomment-675581603) (third point) the AUTOMATICLOT_IN parameter is new and should be tested with both values. I'll open a new Jira issue for this.
> 
> For this PR instead, do you really need to instanciate GeneralData for the 'dicom' package?

Well I need it to test FileSystemDicomManager's loadDetails method. it uses it [here](https://github.com/fjevictoriano/openhospital-core/blob/b27cbf2b96ffa5e773f79db06afdbd7d626db865/src/main/java/org/isf/dicom/manager/FileSystemDicomManager.java#L356-L358)

mwithi:
> Well I need it to test FileSystemDicomManager's loadDetails method. it uses it [here](https://github.com/fjevictoriano/openhospital-core/blob/b27cbf2b96ffa5e773f79db06afdbd7d626db865/src/main/java/org/isf/dicom/manager/FileSystemDicomManager.java#L356-L358)

I see... maybe we can generalize that format, and move the locale into the presentation layer (gui)?

fjevictoriano:
> > Well I need it to test FileSystemDicomManager's loadDetails method. it uses it [here](https://github.com/fjevictoriano/openhospital-core/blob/b27cbf2b96ffa5e773f79db06afdbd7d626db865/src/main/java/org/isf/dicom/manager/FileSystemDicomManager.java#L356-L358)
> 
> I see... maybe we can generalize that format, and move that locale convertion at the presentation layer (gui)?

Yeah and just set the date fields directly from the property file without parsing them.

mwithi:
> Yeah and just set the date fields directly from the property file without parsing them.

right! would you try and test as well?

emecas (approved):
```
$ openhospital/core ((390f8b5...)) mvn clean test

....

Results :

Tests run: 394, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time:  01:51 min
[INFO] Finished at: 2020-09-03T10:26:34-04:00
```

It seems like a good progress about coverage for the module, we can always revisit later any missed parts.

thanks @fjevictoriano 

New `dicom ` numbers 390f8b5:

image

Previous `dicom` numbers 816bdce:

image

PR #135, opened 2020-07-20
Remove redundant public for interface methods

dbmalkovsky:
Removed redundant `public` modifier for interface methods.  Note some of the file already had some methods without the `public` so this makes things consistent.

PR #148, opened 2020-07-31
Add formatting and import order settings files for Eclipse and Intellij

dbmalkovsky:
I have some issues with my pull requests because of formatting issues.    See this [PR](https://github.com/informatici/openhospital-core/pull/135) for an example and the follow up discussion [here](https://github.com/informatici/openhospital-core/pull/135#discussion_r458128798).   The core code base is currently a mixture of tabs and spaces which makes updating difficult.  Using `.editorconfig` is just not sufficient in my humble opinion as it provides very few options and enforcement of styles..  In other projects I work on we provide IntelliJ and Eclipse formatting files that provide many advantages including ordering of imports (see this [comment](https://github.com/informatici/openhospital-core/pull/136#issuecomment-662049302) in another PR of mine).

I would like to contribute the formatting files for use by this project following the same standards used in my other work with the exception that tabs are set instead of spaces as the default (my other projects use spaces).   I understand tabs vs. spaces is a sensitive issue but I and others believe that it provides the least common denominator that works in all situations and editors.   I would be more than happy to provide new updated versions of the files if the decision to use space indentation is accepted.

I personally use IntelliJ so I know that profile is up to date.  The IntelliJ xml file includes both the formatting rules and the import ordering definition.  For Eclipse one needs a second file to specify import ordering in addition to the base formatting definition.  I have done minimal testing with Eclipse and it appears to work correctly.


pviotti:
Thanks - +1 vote for spaces
 - could you rename the folder to start with dot, as most configuration-related files and folders do?
 - will this allow to reformat the whole project in one go using Eclipse or IntelliJ?
 - does this override user's settings just for this project? For instance, if I have a convert-tabs-to-4-spaces setting in Eclipse, for this project it will be ignored and pure tabs will be used?

dbmalkovsky:
If it isn't obvious I am a big +1 for spaces too.

- Folder rename done and pushed.
- I can speak for Intellij and say yes this would allow for reformatting the whole project in one command as this is the IDE that I use daily.  I have attempted the same in a small Eclipse project and it appears to work too.
- Both IntelliJ allow for `project specific settings` for `Code Style` so each project can be different.   I cannot speak to the convert-tabs-to-4-spaces setting in Eclipse as I am not very familiar with Eclipse settings

.

PR #149, opened 2020-08-02
Update import settings for package imports

dbmalkovsky:
This is an update to this [PR](https://github.com/informatici/openhospital-core/pull/148l).   After I applied the formatting and import settings to the whole project I noted that some specific imports of packages were changed into wild card imports.  This corrects the oversight.

Sorry about the second push.


mwithi:
Hi guys, should we do the same change into other repositories? How to use the settings if not automatically adopted by the IDE?

PR #154, opened 2020-08-03
Remove redundant public for interface methods

dbmalkovsky:
This replaces this [PR](https://github.com/informatici/openhospital-core/pull/135).

Removed redundant public modifier for interface methods. Note some of the file already had some methods without the public so this makes things consistent.

Even though it looks like a lot of changes it really isn't as the files each were a mixture of spaces and tabs and I reformatted them all to tabs before applying the change.


pviotti:
LGTM. Thanks!

PR #156, opened 2020-08-04
Op-225 update to Java 8

pviotti:
Due to inactivity in #46 , I'm submitting this PR to have a minimal-impact switch to Java 8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,8 +25,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.6</source>
-					<target>1.6</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/src/main/java/org/isf/admission/service/AdmissionIoOperationRepositoryImpl.java
+++ b/src/main/java/org/isf/admission/service/AdmissionIoOperationRepositoryImpl.java
@@ -2,6 +2,7 @@ package org.isf.admission.service;
 
 
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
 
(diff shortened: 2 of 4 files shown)


pviotti:
I had to add explicit casts to `Date` and turn `GregorianCalendar` to `Date` as well in some JPA comparisons since apparently they changed for Java 8 (it can't use `Comparable` and `GregorianCalendar` anymore: it fails, see [here](https://github.com/informatici/openhospital-core/runs/944401630?check_suite_focus=true#step:4:1281)) ([related StackOverflow](https://stackoverflow.com/questions/9449003/compare-date-entities-in-jpa-criteria-api/9451600)).  
cc: @tomek39856

dbmalkovsky:
I am the author of those two PRs.  They are present because the current code base is a mixture of styles; most notably tabs and spaces in the same file and the order of imports is inconsistent across files.   These files are code styles to import into Eclipse and IntelliJ so the project can standardize on a single style and thus cut down on the number of issues when trying to update code as evidenced by some of my other PRs.

pviotti:
I approved those PRs because a) some style convention is better than no style convention and b) they are quite self-contained (they don't touch code files). But of course we can discuss and improve those settings, they're not set in stone. HTH

emecas:
Thank you @dbmalkovsky @pviotti, it makes totally sense focusing on standardize code style and conventions.  

My question was more why those commits were listed on my local as part of this branch, appeared like pending to be merged to develop branch, I noted later that was looking at an out to date version on my local repo, my mistake.

Also clarify a little how those changes could be linked (or not) to the goal of this particular PR: updating to java8. I mean, are the introduced styles independent from the Java version as well as independent or not of a particular IDEs' versions?

pviotti:
> Also clarify a little how those changes could be linked (or not) to the goal of this particular PR: updating to java8.

they are not related. This Pr is just about switching to java version 8, as per OP-225.  
Those settings are agnostic to java versions AFAIK, and I hope they're quite agnostic to IDE version as well.

pviotti:
> Should we also refactor old code by introducing new Java8 patterns at this stage? 

yes, eventually we should. For the moment let's merge this PR as small as it is (as it is more testable and less disruptive for other contributors).

PR #165, opened 2020-08-19
Add maven license plugin

dbmalkovsky:
Per comment in this [PR](https://github.com/informatici/openhospital-gui/pull/195#discussion_r472464238), here is an example of using the Maven license plugin with associated changes to the `pom.xml` file.

License

To modify the java source files run this command: `mvn process-sources`.

If this PR is OK I can add it to the other repositories.

We will most likely need to review each file to ensure there aren't other bogus headers from the original implementation but it is a start.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,35 @@
 					<finalName>OpenHospital20/bin/OH-core</finalName>
 				</configuration>
 			</plugin>
-
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>license-maven-plugin</artifactId>
+				<version>2.0.0</version>
+				<configuration>
+					<verbose>false</verbose>
+					<addSvnKeyWords>true</addSvnKeyWords>
+				</configuration>
+				<executions>
+					<execution>
+						<id>first</id>
+						<goals>
+							<goal>update-file-header</goal>
+						</goals>
+						<phase>process-sources</phase>
+						<configuration>
+							<licenseName>gpl_v3</licenseName>
+							<inceptionYear>2006</inceptionYear>
+							<organizationName>Informatici Senza Frontiere</organizationName>
+							<projectName>OpenHospital</projectName>
+							<addJavaLicenseAfterPackage>false</addJavaLicenseAfterPackage>
+							<roots>
+								<root>src/main/java</root>
+								<root>src/test</root>
+							</roots>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
 		</plugins>
 	</build>
 


mwithi:
It conflicts with Maven version that at the moment is 3.2.5 or lesser as in [README.md](https://github.com/informatici/openhospital-core/blob/develop/README.md) while the plugin requires version 3.5.4. 

I think that version was linked to Java6, but now we can bump also this.

dbmalkovsky:
Given the project is now using JDK8 the Maven version restriction is gone and the latest version can be used.  I removed the note in the README.md in this [PR](https://github.com/informatici/openhospital-core/pull/167).

mwithi:
> Given the project is now using JDK8 the Maven version restriction is gone and the latest version can be used. I removed the note in the README.md in this [PR](https://github.com/informatici/openhospital-core/pull/167).

Well done @dbmalkovsky !
Should the same plugin be adopted on "gui" and "api"?

mwithi:
Should we include the new generated headers in this PR too?

dbmalkovsky:
@mwithi @pviotti If we decide to include the new headers with this PR my suggestion would be to use the opportunity to reformat all the files as every file will needed to be touched anyway.

pviotti:
> If we decide to include the new headers with this PR my suggestion would be to use the opportunity to reformat all the files as every file will needed to be touched anyway.

makes sense. Then I'd vote for doing it in another PR - which we'll evaluate whether to merge now or when we have less outstanding PRs.

dbmalkovsky:
@mwithi @pviotti  Fine by me; it really doesn't matter.   Going the separate PR route means this PR can be committed, agreed?

mwithi:
Ok, to merge this and to make a new PR with the new headers

pviotti:
this plugin adds licenses to every file every time I run `mvn compile`. Is there an easy way to ignore it via pom.xml or command line? Shall we add those headers?
(also, the header template looks a bit ugly - what with those `$Id:$` and `$HeadURL:$` probably due to `addSvnKeyWords` conf flag...)

dbmalkovsky:
I would personally like to manually add the headers as we update files instead of using the plugin.  

@pviotti @mwithi Should I create another PR and remove the automated header generation?

mwithi:
> I would personally like to manually add the headers as we update files instead of using the plugin.

I thought the idea was to push the new headers all at one after having defined a proper template thanks to the plugin, and also cleaning the result removing previous bogus.

Otherwise what the plugin has been introduced for? 🤔

pviotti:
> Should I create another PR and remove the automated header generation?

yes that would be great - it's a useful plugin, but IMO it's better to have it do its job strictly on demand :)

> push the new headers all at once after having defined a proper template

+1 (maybe we can open a decision log on Jira with the license alternatives and their templates, if there's still something to be decided (?), and discuss it in the next meeting)

PR #166, opened 2020-08-19
Replaces try-finally with try with resources

dbmalkovsky:
Now that Java 8 is the base version, replace `try-finally` with `try` with `resources`.  

Note that most of the changes are the result of reformatting the file to project standards are just white space changes.

If you want to hide the white space changes:
HideWhiteSpaceChanges


mwithi (approved):
LGTM

PR #199, opened 2020-09-06
[OP-229] OPD patient search fix

tomek39856:


mwithi (changes requested):
that might be something wrong, searchin "al":
image

Searching "pas" (it is shown within previous results) returns empty:
image

Even with case "Pas":
image

tomek39856:
did you rebuild with my change? Your examples looks exactly like missing % at the end which I fixed, I'm testing it now and it's working for examples you provided

mwithi:
> did you rebuild with my change? Your examples looks exactly like missing % at the end which I fixed, I'm testing it now and it's working for examples you provided

yeah, now I tried again and it works... it was working partially because before the list was always empty with any keyword

mwithi (approved):
Tested

PR #208, opened 2020-09-08
OP-265 do not depend on all values being the same with constructor method

dbmalkovsky:


src/main/java/org/isf/accounting/manager/BillBrowserManager.java
```diff
@@ -69,9 +69,14 @@ protected void validateBill(Bill bill,
         List<OHExceptionMessage> errors = new ArrayList<OHExceptionMessage>();
         
         GregorianCalendar today = new GregorianCalendar();
```
mwithi, line 71: Fix indentation please

mwithi (approved):
Approved.

src/main/java/org/isf/accounting/manager/BillBrowserManager.java
dbmalkovsky: Done; everything is inline now.

mwithi: Thanks a lot!!!

PR #211, opened 2020-09-10
OP-249 add Junit tests for accounting package

dbmalkovsky:
- Add tests for more coverage
- Use assertj
- Remove return statements in void methods

Coverage Before:
before

Coverage After:
after

### DIFF
--- a/src/test/java/org/isf/accounting/test/TestBill.java
+++ b/src/test/java/org/isf/accounting/test/TestBill.java
@@ -21,7 +21,7 @@
  */
 package org.isf.accounting.test;
 
-import static org.junit.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.GregorianCalendar;
 
@@ -79,6 +79,7 @@ public class TestBill
 		bill.setListName(listName);
 		bill.setPatient(isPatient);
 		bill.setPatient(patient);
+		bill.setPatName(patName);
 		bill.setStatus(status);
 		bill.setAmount(amount);
 		bill.setBalance(balance);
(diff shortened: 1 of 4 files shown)


mwithi (commented):
It works fine (40 tests more, wow!):
```
Results :

Tests run: 430, Failures: 0, Errors: 0, Skipped: 0
```

Anyway question: does it make sense to adopt ‘assertj’ only for one package diverting from other tests?

pviotti (approved):
Looks good, good job!

I'm a bit puzzled though: where does assertj come from? It looks like it's not in `pom.xml`. Is it some kind of transitive implicit dependency? If so, should we add it explicitly as dependency of the project?

> Anyway question: does it make sense to adopt ‘assertj’ only for one package diverting from other tests?

It seems more expressive than standard junit. I guess we change other tests gradually in other PRs. There's also a semi-automatic way of converting assertions from junit: https://joel-costigliola.github.io/assertj/assertj-core-converting-junit-assertions-to-assertj.html#automatic-conversion

dbmalkovsky:
@pviotti assertj is a transitive dependency of org.springframework.boot:spring-boot-starter-test jar.

I am 80% of the way of converting all of the other tests in the core project.   I hope to have a new PR with all the changes tomorrow.

mwithi:
> I'm a bit puzzled though: where does assertj come from? It looks like it's not in `pom.xml`. Is it some kind of transitive implicit dependency? If so, should we add it explicitly as dependency of the project?

yes, it is carried by spring-boot-test 😊

> It seems more expressive than standard junit. I guess we change other tests gradually in other PRs. There's also a semi-automatic way of converting assertions from junit: https://joel-costigliola.github.io/assertj/assertj-core-converting-junit-assertions-to-assertj.html#automatic-conversion

definetely it is! but I'm just wondering if doing it "by pieces" could bring confusion in other contributors, like it happened for other patterns in the past like the controller patterns in api or now the introduction of Spring Repository in tests by-passing the Spring Services (my fault I dind't check twice each PR, and sorry @dbmalkovsky 😅). When a broad refactoring has to be done it is better to do it at once so the code style is always coherent, like we did with the headers, wdyt?

So in this case well welcomed the automated script! 🤓

mwithi:
> I am 80% of the way of converting all of the other tests in the core project. I hope to have a new PR with all the changes tomorrow.

Cool! Should we do the same in "api" as well? Are you using that automated script?

dbmalkovsky:
@mwithi I don't have an automated script but rather there is a plugin in IntelliJ that does most of the work and then I run some other error checkers that are also plugins.  It isn't perfect but it does a lot of the work.

mwithi:
@dbmalkovsky it's ok, should we also document this in CONTRIBUTING.md?

dbmalkovsky:
@mwithi sounds like a reasonable path forward.   BTW, I was further along than I thought and there is now a PR that converts all the remaining tests in CORE.

dbmalkovsky:
FWIW, the conversion of all core tests to assertj is done in [PR 212](https://github.com/informatici/openhospital-core/pull/212)

mwithi (approved):
LGTM
```
Results :

Tests run: 399, Failures: 0, Errors: 0, Skipped: 0
```

PR #213, opened 2020-09-12
OP-271 | Fix findVisit name in VisitManager

emecas:
Signed-off-by: Emerson Castaneda <emecas@users.noreply.github.com>


mwithi (approved):
Seems has no effect on the rest of the application stack

PR #216, opened 2020-09-20
OP-263 more tests on dicom package

mwithi:
As per OP-263 Some tests in ‘dicom' package depend by the client’s Locale and SqlDicomManager setting has not been tested yet.


mwithi:
OP-263 Concerning changes already on `develop`

PR #227, opened 2020-10-09
Changes to get contextLoads() method to work in API repo

dbmalkovsky:
**Summary of Changes**

The root of the changes were to get Spring happy when doing reflection and generating queries.   It is very particular about names matching as it looks for properties, setters, and getters from examining query method names for example.   Most (if not all) changes were to make the component scanning happy.

**The changes by file are:**

_pom.xml_
	Use the hibernate validator instead of the `javax.validation` jar; this fixes a bean instantiation issue
	
_src/main/java/org/isf/accounting/service/AccountingBillIoOperationRepository.java_
	Changed methods names to remove the underscore and capitalize correctly to match the Swing JPA standards (not exactly sure if this was needed)
	
_src/main/java/org/isf/accounting/service/AccountingIoOperations.java_
	Changed method names to match above changes (sorry about the white space issues)
	
_src/main/java/org/isf/exa/service/ExamIoOperationRepository.java_
	Changed method names to git rid of `ExamType` as when Spring does reflection it is looking for exact names and the field is really `Description`
	
_src/main/java/org/isf/therapy/manager/TherapyManager.java_
	Changed `patID` to `patient` because in TherapyRow the field was called `patID` but the datatype was `Patient`
	
_src/main/java/org/isf/therapy/service/TherapyIoOperationRepository.java_
	Changed `PatID` to `Patient` in the first two method names
	Changed the delete method likewise and changed the parameter from Integer to a Patient
	
_src/main/java/org/isf/therapy/service/TherapyIoOperations.java_
	Changes made to match new names and parameter above
	
_src/main/java/org/isf/therapy/service/TherapyPatientMergedEventListener.java_
	Change to use correct set method
	
_src/test/java/org/isf/therapy/test/TestTherapy.java_
	Change to use correct set method

_src/test/java/org/isf/therapy/test/Tests.java_
	Again more name changes to match other changes
	

**What is left to do if this path forward is acceptable:**

_src/main/java/org/isf/therapy/manager/TherapyManager.java_
	There are two commented out lines that have `FIXME FIXME` labels.  I did this to get a clean compile knowing that neither of the two routines currently have any tests associated with them.


dbmalkovsky:
The white space differences can be hidden using HideWhiteSpaceChanges

pom.xml
```diff
@@ -357,6 +357,11 @@
 			<artifactId>hibernate-jpa-2.0-api</artifactId>
 			<version>1.0.1.Final</version>
 		</dependency>
+		<dependency>
```
tomek39856, line 360: shouldn't this dependency be placed in openhospital-api? I assume that it's used to validate  endpoint input parameters. And if we need javax.validation annotations somewhere in code using validation-api was better option than concrete implementation

tomek39856 (approved):

mwithi (commented):
Some questions for @dbmalkovsky and @tomek39856

mwithi (changes requested):
Sorry, my bad, I've wrongly changed base and resolved conflicts with branch https://github.com/informatici/openhospital-core/tree/OP-282-101_Spring5_and_Java8_time, which was not the case.

Current PR should be closed and reopened from a https://github.com/dbmalkovsky/openhospital-core/commit/d2345695094dc435fd53089055462b4ccddf0281 rebase onto current develop.

mwithi:
Merged as single commit on 'develop' branch (**cherry-pick** from https://github.com/dbmalkovsky/openhospital-core/tree/OP-6-ContextLoad). Ref https://github.com/informatici/openhospital-core/commit/1ac7a2db67f6d677ea3b33017d468dee1c0544e4

pom.xml
mwithi: I've tried to comment it in `core` and annotations like `@NotNull` into models are no longer resolved

PR #231, opened 2020-10-19
[OP-282] excluded transient dependency to itext in jesperreports

tomek39856:
I had to exclude transient dep to com.lowagie it was in version which was not found in maven central repo

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -208,6 +208,12 @@
 		    <groupId>net.sf.jasperreports</groupId>
 		    <artifactId>jasperreports</artifactId>
 		    <version>6.14.0</version>
+			<exclusions>
+				<exclusion>
+					<groupId>com.lowagie</groupId>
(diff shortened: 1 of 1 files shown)


mwithi:
https://github.com/TIBCOSoftware/jasperreports/issues/148#issuecomment-702049447

> ... iText 2.1.7 is no longer maintained. We had to stay with it because after 2.1.7, iText changed its license.
The current version of iText is 7+, but since it has different package names inside, it acts like a total different library compared to 2.1.7, so you cannot use it as a replacement ... you can use the original iText 2.1.7 from Maven Central, without our modifications, which are marginal and you probably don't need them. In order to do that, you would have to use "exclude" in your pom.xml when you bring JasperReports.
You can also use a different fork of the iText 2.1.7, which is called OpenPDF:
https://github.com/LibrePDF/OpenPDF

PR #232, opened 2020-10-22
OP-282_OP-101 Spring5 and Java8 time

mwithi:
Draft PR to buffer developments about OP-101 (Spring5) and OP-282 (java.time.*) by the moment that such change will need adoption into "gui" component and this will take time.


pviotti:
for more context on OP-101 and OP-282: https://github.com/informatici/openhospital-core/pull/120 
Also, as I was noting [here](https://openhospital.atlassian.net/browse/OP-319), in #120 I forgot to migrate the `Auditable` class to java.time API. If anyone wants to fix that, please do.

mwithi:
> If anyone wants to fix that, please do.

Thanks @pviotti ! I would create an "reminder" issue for that, so we don't forget

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -22,9 +22,8 @@
 package org.isf.opd.manager;
 
 import java.util.ArrayList;
-import java.util.Date;
-import java.util.GregorianCalendar;
 import java.util.List;
+import java.time.*;
```
dbmalkovsky, line 26: There are 8 occurrences of package imports; the standard is to enumerate all imports.  Thansk

mwithi: Correct: the standard is [here](https://github.com/informatici/openhospital-core#code-style).

PR #234, opened 2020-10-25
[OP-282] reporting fix: workaround for LocalDateTime to Date conversion

tomek39856:
fixed erros caused by java time upgrade in reports module

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -208,6 +208,12 @@
 		    <groupId>net.sf.jasperreports</groupId>
 		    <artifactId>jasperreports</artifactId>
 		    <version>6.14.0</version>
+			<exclusions>
+				<exclusion>
+					<groupId>com.lowagie</groupId>
+					<artifactId>itext</artifactId>
+				</exclusion>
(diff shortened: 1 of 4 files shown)


mwithi:
queried Jaspersoft for this
https://github.com/TIBCOSoftware/jasperreports/issues/151

tomek39856:
Ok but maybe merge it now and than remove this conversion when they'll fix the issue? I also fixed within this PR one more bug related strictly to java.time way of working not to jasperreports (change in TimeTools class)

soc:
I think the JasperReports API mirrors the JDBC `Statement` one – so `setTime` means `sql.Time` and `setDate` means `sql.Date`.

If you want to pass other things, you might try `setStatementParameter`, which falls through to `setObject` (which is how `time.*` classes are passed, as they didn't add new methods to the JDBC API for those types.

Maybe this helps: https://stackoverflow.com/questions/43039614/insert-fetch-java-time-localdate-objects-to-from-an-sql-database-such-as-h2.

PR #214, opened 2020-09-13
OP-272 Remove contexts and jpaUtil from tests

tomek39856:
Introduced common test base with cleaning db and get rid of test contexts

Created to track progress not ready yet

PR #239, opened 2020-11-10
OP-244 AUTOMATICLOT parameterized tests WIP

dbmalkovsky:
This is WIP and I'm looking for feedback.

What I did was change the testing type to `Parameterized.class`.  Then I defined the values for the parameters 
````
	@Parameterized.Parameters(name ="Test with AUTOMATICLOT_IN={0}, AUTOMATICLOT_OUT={1}, AUTOMATICLOTWARD_TOWARD={2}")
	public static Collection automaticlot() {
		return Arrays.asList(new Object[][] {
				{ false, false, false },
				{ false, false, true },
				{ false, true, false },
				{ false, true, true },
				{ true, false, false },
				{ true, false, true },
				{ true, true, false },
				{ true, true, true }
		});
	}
````
Next I added a constructor method that takes the parameter values and assigns them to the three GeneralData variables: 
```
    public Tests(boolean in, boolean out, boolean toward) {
	    GeneralData.AUTOMATICLOT_IN = in;
	    GeneralData.AUTOMATICLOT_OUT = out;
	    GeneralData.AUTOMATICLOTWARD_TOWARD = toward;
    }
```
Because I had to `SpringRunner` to use the `Parameterized.class` (JUnit doesn't allow multiple runners in one JUnit test class. This means we can't take advantage of `SpringRunner`) so I had to manually do that with something recommended by Spring:
````
	@ClassRule
	public static final SpringClassRule scr = new SpringClassRule();

	@Rule
	public final SpringMethodRule smr = new SpringMethodRule();
````
And then I fired off the test file and it seems to hang after passing 78 tests and failing on 10. 

The output in my IDE looks _something_ like this where each test is annotated with the parameter values used:

11-10-2020 2-28-14 PM

Is this what OP-244 was looking for?

### DIFF
--- a/src/test/java/org/isf/medicalstock/test/Tests.java
+++ b/src/test/java/org/isf/medicalstock/test/Tests.java
@@ -31,6 +31,8 @@ import java.util.Collection;
 import java.util.GregorianCalendar;
 import java.util.List;
 
+import javax.persistence.EntityManager;
+
 import org.isf.generaldata.GeneralData;
 import org.isf.medicals.model.Medical;
 import org.isf.medicals.test.TestMedical;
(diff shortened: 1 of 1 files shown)


dbmalkovsky:
The CI job was hanging and my local running of the code showed that exceptions were appearing during test cleanups (among other places) so I *borrowed* some code from this [PR](https://github.com/informatici/openhospital-core/pull/214) entitled `OP-272 Remove contexts and jpaUtil from tests` and now the base 19 tests get run for a clean 152 test suite.

mwithi:
@dbmalkovsky can we close this PR?

dbmalkovsky:
Closing this PR as the method has been successfully used in various test suites.

PR #241, opened 2020-11-12
OP-223 Upgrade demo data script

mwithi:
Trying to move the demo data loading to the end, as an additional and optional step at the end of the DB creation, but to do so the original script needs changes in order to avoid dataset duplications (that turn into errors and warnings) and at the same time to include all changes after step_04.

- updated `DeleteAllData.sql` -> **delete_all_data.sql**
- created **load_demo_data.sql** with all self contained data
- removed old `step_04_*` scripts (old demo data)
- renamed `from_step_05.sql` -> `step_04_all_following_steps.sql` (**temporary name**)
- modified **create_all_demo.sql** script in order to 1) create the DB 2) delete_all_data.sql 3) load_demo_data.sql
- modified all other `create_all_*.sql` scripts accordingly
- updated `Dockerfile-ohdb` (but no changes in the build)

**TODO**:
- [x] clean up scripts
- [x] anonymize data
- [x] update dates (manually or with a reusable script)
- [x] add missing demo data (manually or with a structured solution)

NOT CONCERNING THIS PR (**TO DO NOT**):
x - not reorganizing other scripts (OP-344)
x - not adding missing tranlsations
x - other things not related to demo data


mizzioisf:
Hi mwithi, 
I can't find the delete_all_data.sql file, what I'm missing ?

mizzioisf:
I propose to completely drop the step_00_create_user.sql for the following reasons:

- it's the final missing piece of code (should be :-) ) where we have hard coded credential :-(
- syntax is old and not supported after mysql 5.7 (creating user by granting)
- we have already a more complete and centralized set of commands (drop if exist, create, grant permission) in oh.sh

mwithi:
> I propose to completely drop the step_00_create_user.sql for the following reasons:
> 
> * it's the final missing piece of code (should be :-) ) where we have hard coded credential :-(
> * syntax is old and not supported after mysql 5.7 (creating user by granting 👎 )
> * we have already a more complete and centralized set of commands (drop if exist, create, grant permission) in oh.sh
> * the docker buidl file already contains the variables in ENV format

As stated in the issue's description, all these (very important things) are not part of this issue, but rather OP-344, coming soon anyway. 

In order to speed up this PR, contributions on TODOs are welcome!

mizzioisf (approved):

pviotti (approved):
I suggest not renaming those three files to `.old`: if it's not used anymore, just delete it. It's in git history anyway, if we need to restore it.

mizzioisf (approved):
Ok to merge

PR #244, opened 2020-11-17
Remove dup dependency for jai_imageio and jai-core

dbmalkovsky:
Remove duplicate dependency for `jai_imageio` and `jai-core`.   First definition was at lines 233-248 and the second one was at lines 254-269.

This was the result of a push on March 28, 2020.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,22 +231,6 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-		    <groupId>com.sun.media</groupId>
-		    <artifactId>jai_imageio</artifactId>
-		    <version>1.1</version>
-		    <exclusions>
-		    	<exclusion>
-		    		<groupId>javax.media</groupId>
-		    		<artifactId>jai_core</artifactId>
-		    	</exclusion>
-		    </exclusions>
-		</dependency>
-		<dependency>
-		    <groupId>javax.media</groupId>
-		    <artifactId>jai-core</artifactId>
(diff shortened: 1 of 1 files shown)


dbmalkovsky:
The CI builds are failing because they are unable resolve `jai-core:jar:1.1.3`; if one looks at maven (specifically at [https://mvnrepository.com/artifact/javax.media/jai-core/1.1.3](https://mvnrepository.com/artifact/javax.media/jai-core/1.1.3)) there is a note at the bottom that reads:  `Note: this artifact is located at Atlassian 3rdParty repository (https://packages.atlassian.com/maven-3rdparty/)`.  And sure enough when you look at the raw file structure there is no jar there: https://repo.maven.apache.org/maven2/javax/media/jai_core/1.1.3/

It appears to be working locally as at some point (at least in my local maven repo) there is a `jai-core-1.1.3.jar` present that is being used.

dbmalkovsky:
So it appears that the last commit to add the JFrog repo made it possible to resolve the dependency and everything runs clean again.  @mwithi I think this needs to be committed soon so others don't run into the problem.

mwithi:
It works fine, nevertheless can we get rid of non-central repositories like in https://github.com/informatici/openhospital-core/pull/248?

mwithi:
Superseded by https://github.com/informatici/openhospital-core/pull/248.

mwithi:
Superseded by #248.

PR #248, opened 2020-11-18
Removing repo-spring-io repository

mwithi:
Moving unaccessible library into local repo/ folder

handling build error:
https://github.com/informatici/openhospital-core/runs/1405837170?check_suite_focus=true#step:4:985

following Spring notice:
https://spring.io/blog/2020/10/29/notice-of-permissions-changes-to-repo-spring-io-fall-and-winter-2020


dbmalkovsky:
@mwithi This sort of overlaps with PR #244.  You could go ahead and remove the spring repo but and the Atlassian JFrog repo like in PR #244 and then there is no need to add a repo folder to the project.

mwithi:
@dbmalkovsky you are right, I completely forgot to review your #244 for DICOM usage (and I did it now, JFrog works fine).

The problem here has come exactly by the fact of not using a central repository (like Maven Central) and "Atlassian 3rd party" repository could just turn unavailable in the future as well (as occurred with repo-spring-io) so I don't know what is the best solution at this point.

We already have a local repo for other libraries... could it be a reminder for everybody that those "provided" libraries need to be removed/replaced ASAP?

Thanks a lot!

dbmalkovsky:
OK; if this is the option chosen please don't forget about the duplicate definition of dependencies that were removed as part of #244

mwithi:
yes, 

> OK; if this is the option chosen please don't forget about the duplicate definition of dependencies that were removed as part of #244

Yes, I've removed also https://github.com/informatici/openhospital-core/pull/248/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L233

Thanks!

pviotti:
+1 on getting rid of of those checked in binaries ASAP... ideally you shouldn't check-in binaries to git like it's early 2000 again.
If that library is actively maintained, then it should be hosted by some stable repository out there (and probably Atlassian's is fairly stable..). If it's just a legacy unmaintained library, then maybe we should get rid of it.

mwithi:
> If it's just a legacy unmaintained library, then maybe we should get rid of it.

We hope this one :)

PR #249, opened 2020-11-19
Chore: Private method cannot be overridden at runtime

dbmalkovsky:
Fix a case where code prevents a class from being subclassed (because of the `private` modifier) by some frameworks (e.g. Spring and Hibernate) at runtime.


dbmalkovsky:
This was found by running IntelliJ's Analyze tool specifically the inspection: `Final declaration can't be overridden at runtime`.   The subtext reads: Methods annotated with `@Transactional` must be overridable.

PR #253, opened 2020-11-22
OP-340 Updating Ward fix

dnlebgwa:
Changing Ward hibernate core code to return saved object instead of boolean to fix OP-340


src/test/java/org/isf/ward/test/Tests.java
```diff
@@ -148,11 +172,10 @@ public void testIoUpdateWard() {
 			foundWard.setDescription("Update");
 
 			// when:
-			boolean result = wardIoOperation.updateWard(foundWard);
+			wardIoOperation.updateWard(foundWard);
 			Ward updateWard = wardIoOperationRepository.findOne(code);
 
 			// then:
-			assertTrue(result);
```
dbmalkovsky, line 155: FWIW, by the fact that `assertTrue()` and `catch-try-fail` are present it appears that the `develop` branch this was based on is significantly out of date.

dnlebgwa: @dbmalkovsky  sorry about that. I have updated my branch and re-merged the changes. Is this ok?

src/main/java/org/isf/ward/service/WardIoOperations.java
```diff
@@ -11,13 +11,36 @@
...
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+>>>>>>> 1e7c42900a486f154cf95232f7189b88b929e885
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
+<<<<<<< HEAD
```
dbmalkovsky, line 33: This conflict needs to be resolved.

src/main/java/org/isf/ward/service/WardIoOperations.java
```diff
@@ -11,13 +11,36 @@
...
  *
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
+=======
```
dbmalkovsky, line 37: This conflict needs to be resolved.

src/main/java/org/isf/ward/manager/WardBrowserManager.java
```diff
@@ -142,10 +142,14 @@ protected void validateWard(Ward ward, boolean insert) throws OHServiceException
 	 * Stores the specified {@link Ward}. 
 	 * In case of error a message error is shown and a <code>false</code> value is returned.
 	 * @param ward the ward to store.
-	 * @return <code>true</code> if the ward has been stored, <code>false</code> otherwise.
+<<<<<<< HEAD
```
dbmalkovsky, line 145: This is still a conflict; needs a resolution.  There are other occurrences too.

src/test/java/org/isf/ward/test/Tests.java
dbmalkovsky: There are still issues; see my comments in the source files.

dnlebgwa:
Thanks @dbmalkovsky. I believe I have addressed your comments in my latest commit
https://github.com/informatici/openhospital-core/pull/253/commits/2547a4dfb06712cef71208445578716c96f45da0

PR #254, opened 2020-11-25
Op 262 contextualize weight and height

mwithi:
Found out that the `@Transient weight` property in `Patient` entity was used only (1 occurrence) in `WardPharmacyNew`, while `@Transient height` property never used.

Going to simplify Patient’s entity and to retrieve his/her weight only when needed.

### DIFF
--- a/src/main/java/org/isf/patient/manager/PatientBrowserManager.java
+++ b/src/main/java/org/isf/patient/manager/PatientBrowserManager.java
@@ -104,7 +104,7 @@ public class PatientBrowserManager {
 	 * @return the Patient that match specified name (could be null)
 	 * @throws OHServiceException 
 	 * @deprecated use getPatient(Integer code) for one patient or 
-	 * getPatientWithHeightAndWeight(regex) for a list
+	 * getPatientsByOneOfFieldsLike(String regex) for a list
 	 */
 	public Patient getPatientByName(String name) throws OHServiceException {
         return ioOperations.getPatient(name);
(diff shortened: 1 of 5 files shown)


mwithi:
@dbmalkovsky I would like your quick review also if possible. Thanks!

src/main/java/org/isf/patient/manager/PatientBrowserManager.java
```diff
@@ -104,7 +104,7 @@ public Patient savePatient(Patient patient) throws OHServiceException {
 	 * @return the Patient that match specified name (could be null)
 	 * @throws OHServiceException 
 	 * @deprecated use getPatient(Integer code) for one patient or 
-	 * getPatientWithHeightAndWeight(regex) for a list
+	 * getPatientsByOneOfFieldsLike(String regex) for a list
 	 */
 	public Patient getPatientByName(String name) throws OHServiceException {
```
dbmalkovsky, line 109: Missing `@Deprecated` annotation

dbmalkovsky:
@mwithi LGTM (noticed one little change that can be ignored if desired)

src/main/java/org/isf/patient/manager/PatientBrowserManager.java
mwithi: Yep, it seems not needed any more by the majority of the compilers, but for homogeneity of the code I would add it.

Maybe we can investigate more in order to see if to stay with `@deprecated` in Javadoc only and slim down the code, but it's not a priority of course.

Thanks!

dbmalkovsky: I'm not sure but does the Java compiler "read" the JavaDoc?  If you don't add the annotation then the compiler can't flag uses of deprecated methods I believe.

mwithi: well there's a little bit debate on it... https://stackoverflow.com/a/5039811/5309541

on my IDE it is shown as deprecated in any way:

image

mwithi: Commited anyway: https://github.com/informatici/openhospital-core/pull/254/commits/aa7eb54519d481bd54ea013ae71954a98e00b399

Note: the commit's comment was correct (Uppercase) but GitHub changed it 😄

PR #257, opened 2020-11-27
OP-374 increase coverage for admtype and agetype packages

dbmalkovsky:
Coverage before:
image

Coverage after:
image

Updated `assertj` to something newer to use newer features

Also removed from `AgeTypeBrowserManager` two methods that are not needed given autowiring.  I also grepped across all the repos and found no uses:
`````
	public AdmissionTypeIoOperation getIoOperations() {
		return ioOperations;
	}

	public void setIoOperations(AdmissionTypeIoOperation ioOperations) {
		this.ioOperations = ioOperations;
	}

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -403,12 +403,24 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<version>1.5.22.RELEASE</version>
 			<scope>test</scope>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by mwithi — APPROVED** [review]
LGTM

```
Results :

Tests run: 443, Failures: 0, Errors: 0, Skipped: 1
```

PR #263, opened 2020-12-02
OP-381 increase coverage for dlvrestype and dlvrtype packages

dbmalkovsky:
Add tests to increase coverage and formatter `Manager` files to match standards.

Coverage before:
image

Coverage after:
image

All tests pass locally.

### DIFF
--- a/src/main/java/org/isf/dlvrrestype/manager/DeliveryResultTypeBrowserManager.java
+++ b/src/main/java/org/isf/dlvrrestype/manager/DeliveryResultTypeBrowserManager.java
@@ -32,8 +32,6 @@ import org.isf.utils.exception.OHDataValidationException;
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
 import org.isf.utils.exception.model.OHSeverityLevel;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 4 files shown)


emecas (approved):
LGTM

Thank you @dbmalkovsky

emecas:
This is not a  blocking comment to merge this PR, but I noticed that the way Eclipse measures coverage is more precise than IntelliJ, and this is why:

image

This is the portion of code for the 2 missed instructions in `org.isf.dlvrtype.service.DeliveryTypeIoOperation`  because them are being evaluate in different way, since Eclipse considers coverage separately for all possible paths, for this particular cases (lines 67 & 85) , the cases for `true` and `false`:

image

dbmalkovsky:
@emecas  Please see the comments in [OP-391](https://openhospital.atlassian.net/browse/OP-391) where I discuss why I believe the value will never be null.   The `save()` routine returns an object or throws an exception.

emecas:
> 
> 
> @emecas Please see the comments in [OP-391](https://openhospital.atlassian.net/browse/OP-391) where I discuss why I believe the value will never be null. The `save()` routine returns an object or throws an exception.

@dbmalkovsky I just wanted to bring here the subject about differences on coverage measures reported in each IDE, I guess I did not use the ideal lines to exemplify it.

I'm mostly agree about the particular case for OP-391, but also it is true that we can write unit tests mocking the repositories methods in order to return and test all the potential expected values (true, false, null, exception). You work has been exceptional for the scope. We can handle this subject and related somewhere else.

PR #264, opened 2020-12-04
OP-382 Increase coverage and format Manager files

dbmalkovsky:
Increase test coverage for `exa`, `examination` and `exatype` packages.
Also reformat `Manager` classes to meet project standards and remove unused `logger` variables.

All tests pass locally.

Coverage before:
image

Coverage after:
image


emecas (approved):
LGTM

Great job @dbmalkovsky , thanks for adding that huge amount of testing that was in pending for these modules

PR #265, opened 2020-12-05
Chore: Remove NPE in logging statement

dbmalkovsky:
In the logging statement the else clause is reached only when `defaultPrinterService` is null.   Attempting to `toString()` on a null value results in a NPE.  Just removed the statement as the other logging statement indicates the problem.

Found with SpotBugs.


mwithi (approved):
Good catch! Thanks a lot!

PR #269, opened 2020-12-17
OP-383 increase coverage Hospital and Lab; reformat managers

dbmalkovsky:
Coverage before:
image

Coverage after:
image

I could not get 100% coverage of `LabManager` because the private field `materialHashMap` was not accessible through Java reflection as I've done in other packages.   The reason is that the `@Transactional` annotation is used on two methods causing the class to generate a proxy class to handle the rollback.  So when I tried to `getDefinedFields` through reflection all I got was the proxy class methods and not the protected fields in the base class.  I tried a few options like `getSuperClass()` where I could get the variable but not set it to any new value.  Thus only 98% coverage for that class.


emecas (approved):
Great Job @dbmalkovsky 

thanks!

PR #277, opened 2020-12-23
OP-372 Operation results list return code instead of translations

gildasdongmo:


src/main/java/org/isf/operation/manager/OperationBrowserManager.java
```diff
@@ -144,11 +144,11 @@ public boolean descriptionControl(String description, String typeCode) throws OH
...
 	 */
-	public ArrayList<String> getResultsList() { //TODO (OP-372): deal with codes instead of translations!!!
+	public ArrayList<String> getResultsList() { 
 		ArrayList<String> resultsList = new ArrayList<String>();
-		resultsList.add(MessageBundle.getMessage("angal.operation.result.sucess"));
-		resultsList.add(MessageBundle.getMessage("angal.operation.result.failure"));
-		resultsList.add(MessageBundle.getMessage("angal.operation.result.undefined"));
+		resultsList.add("angal.operation.result.sucess");
+		resultsList.add("angal.operation.result.failure");
+		resultsList.add("angal.operation.result.undefined");
```
mwithi, line 151: fix: sucess -> success

Anyway, with this method, in the DB will finish the whole string "angal.operation.result.sucess":

image

What about using a mapping mechanism as done in (e.g.) [ExaminationBrowserManager.java:92](https://github.com/informatici/openhospital-core/blob/develop/src/main/java/org/isf/examination/manager/ExaminationBrowserManager.java#L92) ?

mwithi (changes requested):
to fix bundle name and to revise the mapping to the DB.

in any case the migration script [step_57_multiple_operations_in_admission.sql](https://github.com/informatici/openhospital-core/blob/develop/mysql/db/step_57_multiple_operations_in_admission.sql) needs to be changed accordingly

mwithi:
@gildasdongmo +1

src/main/java/org/isf/operation/manager/OperationBrowserManager.java
gildasdongmo: > What about using a mapping mechanism as done in (e.g.) [ExaminationBrowserManager.java:92](https://github.com/informatici/openhospital-core/blob/develop/src/main/java/org/isf/examination/manager/ExaminationBrowserManager.java#L92) ?

It was just a choice, I can do it like that if necessary.

mwithi: I think it's better as pattern; also step_57_multiple_operations_in_admission.sql should be changed after.
Thanks!

gildasdongmo: Hi @mwithi, I made a new commit that uses the mapping mechanism as requested. and for what needs to be saved in bd, I used the keys already present in step_57_multiple_operations_in_admission.sql file and I think it won't be necessary to update it anymore.

src/main/java/org/isf/operation/manager/OperationBrowserManager.java
```diff
@@ -144,11 +150,39 @@ public boolean descriptionControl(String description, String typeCode) throws OH
...
+	public LinkedHashMap<String, String> getResultsList() { 
+		if (resultsListHashMap == null) buildResultHashMap();
+		return resultsListHashMap;
+	}
+	
+	private void buildResultHashMap() {
+		resultsListHashMap = new LinkedHashMap<>();
+		resultsListHashMap.put("Success", MessageBundle.getMessage("angal.operation.result.success"));
+		resultsListHashMap.put("Failure", MessageBundle.getMessage("angal.operation.result.failure"));
+		resultsListHashMap.put("Unknown", MessageBundle.getMessage("angal.operation.result.undefined"));
```
mwithi, line 162: as "codes" in the DB I would put them lowercase

mwithi (changes requested):
It looks fine, but:
- let's use lowercase "keys" in the db
- please resolve the conflicts
- please apply imports standards (see README.md)

src/main/java/org/isf/operation/manager/OperationBrowserManager.java
gildasdongmo: > * let's use lowercase "keys" in the db
> * please resolve the conflicts
> * please apply imports standards (see README.md)

**All done.**

> in any case the migration script [step_57_multiple_operations_in_admission.sql](https://github.com/informatici/openhospital-core/blob/develop/mysql/db/step_57_multiple_operations_in_admission.sql) needs to be changed accordingly

script has been updated using lowercase **keys**

mwithi (approved):

mwithi:
LGTM.
```
Results :

Tests run: 1105, Failures: 0, Errors: 0, Skipped: 1
```

PR #281, opened 2020-12-31
OP-395 fix dateFrom and dateTo values

dbmalkovsky:
See  https://openhospital.atlassian.net/browse/OP-395

### DIFF
--- a/src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
+++ b/src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
@@ -102,15 +102,6 @@ public class OpdIoOperationRepositoryImpl implements OpdIoOperationRepositoryCus
 				cb.equal(opd.get("newPatient"), newPatient)
 			);
 		}
-		dateFrom.clear(Calendar.HOUR);
-		dateFrom.clear(Calendar.MINUTE);
-		dateFrom.clear(Calendar.SECOND);
-		dateFrom.clear(Calendar.MILLISECOND);
-		dateTo.clear(Calendar.HOUR);
-		dateTo.clear(Calendar.MINUTE);
-		dateTo.clear(Calendar.SECOND);
-		dateTo.clear(Calendar.MILLISECOND);
-		dateTo.add(Calendar.DAY_OF_WEEK, 1);
 		predicates.add(
 			cb.between(opd.<Date>get("visitDate"), dateFrom.getTime(), dateTo.getTime())
 		);
--- a/src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
+++ b/src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
@@ -22,6 +22,7 @@
 package org.isf.opd.service;
 
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


mwithi (changes requested):
are you sure about the empty return values with false? the criteriabuilder was already [fixed](https://github.com/informatici/openhospital-core/commit/c1fe72df6c84cd70b15ca103992728bfde37abc7) to compare [only dates](https://github.com/informatici/openhospital-core/blob/ea06ac2f2eaf1c31f96af1a525599c967ff48bba/src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java#L106) in the BETWEEN sql function.

Anyway, to clear hour, minute, second, millisecond values, we should already have:

- org.isf.utils.time.TimeTools.getDateToday0()
- org.isf.utils.time.TimeTools.getDateToday24()
- org.isf.utils.time.TimeTools.getBeginningOfDay(GregorianCalendar)
- org.isf.utils.time.TimeTools.getBeginningOfNextDay(GregorianCalendar)

dbmalkovsky:
When I tried to write test cases they all failed until I implemented the changes in this PR.

mwithi:
> When I tried to write test cases they all failed until I implemented the changes in this PR.

I'm sure of that, but the point is: if the linked fix is not working as expected we have a problem in many other points where it has been used

dbmalkovsky:
In the `getOpdList()` method in `OpdIoOperations.java` has:

````
		GregorianCalendar dateFrom=new GregorianCalendar();
		GregorianCalendar dateTo=new GregorianCalendar();
````

The query builder than has: 
`````
		predicates.add(
			cb.between(opd.<Date>get("visitDate"), dateFrom.getTime(), dateTo.getTime())
		);
`````

There is no value that will ever be between those two values because they are set to the same value.

FWIW, I went and looked in the OP-282-101_Spring5_and_Java8_time branch and it is not correct as it ignores the `boolean oneWeek` parameter and assumes that the value is `true`:

````
	public ArrayList<Opd> getOpdList(boolean oneWeek) throws OHServiceException	{
		LocalDate dateTo = LocalDate.now();
		LocalDate dateFrom = dateTo.minusWeeks(1);
		
		return getOpdList(MessageBundle.getMessage("angal.opd.alltype"),MessageBundle.getMessage("angal.opd.alldisease"),dateFrom,dateTo,0,0,'A','A');
	}
````

Guess that will be found when the track gets merged and the JUnit tests are run against it.

I will update this PR to use the utility methods to clear the date value.

dbmalkovsky:
I made the changes to use the utility class.  As I continue with the JUnit test coverage I will be on the look out for other related issues.

mwithi:
> There is no value that will ever be between those two values because they are set to the same value.

My understanding was that the typing `<Date>` was providing to the `BETWEEN` function only the date part (excluding the time) so the resulting query would be equivalent to `BETWEEN DATE(dateFrom) AND DATE(dateTo)` but now I realize it is not like that.

We should verify and fix other places where `javax.persistence.criteria.CriteriaBuilder.between` has been used bacause it may conceal the same problem.

> Guess that will be found when the track gets merged and the JUnit tests are run against it.

True, but we could put a comment on the PR#232 just to be sure to not forget?


Thanks for changes in https://github.com/informatici/openhospital-core/pull/281/commits/8deaccbe695a6d1aaaba429d31f0d6186f8234aa!

mwithi (approved):

emecas (approved):
LGTM

Thanks @dbmalkovsky 

is  coverage for these modules what you are going to work on later?

image

dbmalkovsky:
@emecas Yes those are some of the packages left on my list.

mwithi:
Thanks a lot guys!

PR #282, opened 2021-01-01
OP-396 increase coverage in operation; format Manager files

dbmalkovsky:
Coverage before:
image

Coverage after:
image

### DIFF
--- a/src/test/java/org/isf/operation/test/Tests.java
+++ b/src/test/java/org/isf/operation/test/Tests.java
@@ -24,14 +24,56 @@ package org.isf.operation.test;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.ArrayList;
+import java.util.GregorianCalendar;
+import java.util.List;
 
 import org.isf.OHCoreTestCase;
+import org.isf.admission.manager.AdmissionBrowserManager;
+import org.isf.admission.model.Admission;
+import org.isf.admission.service.AdmissionIoOperationRepository;
+import org.isf.admission.service.AdmissionIoOperations;
+import org.isf.admission.test.TestAdmission;
+import org.isf.admtype.model.AdmissionType;
+import org.isf.admtype.service.AdmissionTypeIoOperationRepository;
+import org.isf.admtype.test.TestAdmissionType;
+import org.isf.disctype.model.DischargeType;
+import org.isf.disctype.service.DischargeTypeIoOperationRepository;
+import org.isf.disctype.test.TestDischargeType;
+import org.isf.disease.model.Disease;
+import org.isf.disease.service.DiseaseIoOperationRepository;
+import org.isf.disease.test.TestDisease;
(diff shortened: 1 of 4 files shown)


emecas (commented):
Thanks @dbmalkovsky 

LGTM

19e25685

```
Results :

Tests run: 716, Failures: 0, Errors: 0, Skipped: 2

[INFO] BUILD SUCCESS
[INFO] Total time:  01:24 min
[INFO] Finished at: 2021-01-09T00:46:30-05:00
```

emecas (approved):

src/test/java/org/isf/operation/test/Tests.java
```diff
@@ -119,23 +295,693 @@ public void testIoUpdateOperation() throws Exception {
...
+		DeliveryType deliveryType = testDeliveryType.setup(false);
+		DeliveryResultType deliveryResult = testDeliveryResultType.setup(false);
+
+		Admission admission = testAdmission.setup(ward, patient, admissionType, diseaseIn, diseaseOut1,
+				diseaseOut2, diseaseOut3, operation, dischargeType, pregTreatmentType,
+				deliveryType, deliveryResult, false);
+		operationRow.setAdmission(admission);
+
+		assertThat(operationRow).hasToString("some new description TestUserId");
 	}
```
emecas, line 985: just wonder , why this big test just for testing  `OperationRow.toString` method, could it not be implicitly covered in any other test?

or maybe  a simplified version of the test, only loading the compromised values like:

	@Test
	public void testOperationRowToStringSimplified() throws Exception {
		OperationType operationType = testOperationType.setup(true);
		Operation operation = testOperation.setup(operationType, false);
		
		operation.setDescription("aDescription");
		OperationRow operationRow = testOperationRow.setup(operation, true);

		Patient patient = testPatient.setup(false);
		Admission admission = testAdmission.setup(null, patient, null, null, null,
				null, null, operation, null, null,
				null, null, false);
		admission.setUserID("UserID");
		operationRow.setAdmission(admission);

		assertThat(operationRow).hasToString("aDescription UserID");
	}

dbmalkovsky: Thanks; used your simplified version.

I was under the impression from other reviewers that a single test should test a single method;  that combining things was frowned upon and thus the reason for this test structure.

mwithi (approved):
LGTM

```
Results :

Tests run: 804, Failures: 0, Errors: 0, Skipped: 1
```

src/test/java/org/isf/operation/test/Tests.java
emecas: Great.

I am agree about following single responsibility principle by test, but I don't typically add  tests for coverage of setters, getters, equals, or hash, mainly because all of them typically get tested implicitly. 

toString usually fits in same category, but as the case here you can be expecting something else about the returning string.

PR #300, opened 2021-01-20
Update PR-293 by replacing missing line

dbmalkovsky:
When making the change to correct the compile error it appears that a line setting the font was deleted by mistake.


mwithi:
Good catch @dbmalkovsky ! Thanks a lot! (verified also)

PR #303, opened 2021-01-21
OP-404 update: remove duplicated tests and fix spelling

dbmalkovsky:
I was reviewing the PR for OP-404 and noticed a spelling mistake and that two tests were duplicated.  This PR fixes the spelling and removes the duplicated tests.

### DIFF
--- a/src/test/java/org/isf/visits/test/Tests.java
+++ b/src/test/java/org/isf/visits/test/Tests.java
@@ -174,7 +174,7 @@ public class Tests extends OHCoreTestCase {
 	}
 
 	@Test
-	public void testMgrGetVisiNoPatientCode() throws Exception {
+	public void testMgrGetVisitNoPatientCode() throws Exception {
 		int id = _setupTestVisit(false);
 		Visit visit = visitsIoOperationRepository.findOne(id);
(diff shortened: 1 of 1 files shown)


src/test/java/org/isf/visits/test/Tests.java
```diff
@@ -202,20 +202,6 @@ public void testMgrGetVisitsEmpty() throws Exception {
...
-		assertThat(visitManager.getVisits(0)).hasSize(1);
-	}
-
-	@Test
-	public void testMgrGetVisitsbyID() throws Exception {
-		int id = _setupTestVisit(false);
-		Visit visit = visitsIoOperationRepository.findOne(id);
-		assertThat(visitManager.getVisits(id)).hasSize(1);
-	}
-
```
mwithi, line 218: is correct eliminating both?

mwithi (commented):
twop tests were duplicating each other but we have to keep one at least?

src/test/java/org/isf/visits/test/Tests.java
dbmalkovsky: Yes; each one is duplicating another test elsewhere in the file.   Sorry I didn't make it more clear.

PR #304, opened 2021-01-21
OP-411 correct deleteAllTherapies() method

dbmalkovsky:
I was writing tests for the Therapy package and noticed that the method `deleteAllTherapies(Integer code)` was not working correctly.

I updated the method body without changing the method signature and my local tests now run cleanly.

This is a separate PR so the change doesn't get lost in the larger JUnit test suite to come.

### DIFF
--- a/src/main/java/org/isf/therapy/manager/TherapyManager.java
+++ b/src/main/java/org/isf/therapy/manager/TherapyManager.java
@@ -255,9 +255,8 @@ public class TherapyManager {
 	 */
 	@Transactional(rollbackFor=OHServiceException.class)
 	public boolean deleteAllTherapies(Integer code) throws OHServiceException {
-		// FIXME FIXME	ioOperations.deleteAllTherapies(code);
-		smsOp.deleteByModuleModuleID("therapy", String.valueOf(code));
-		return true;
+		Patient patient = patientManager.getPatientById(code);
+		return ioOperations.deleteAllTherapies(patient);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


mwithi (changes requested):
Maybe we can remove also the copy-and-paste comment on [line 195](https://github.com/informatici/openhospital-core/blob/24c4d4d541b1c85c03328e585eb3650dedebf07b/src/main/java/org/isf/therapy/manager/TherapyManager.java#L195)

dbmalkovsky:
Good catch; made the same type of change as was done originally

src/main/java/org/isf/therapy/manager/TherapyManager.java
```diff
@@ -188,12 +187,10 @@ public TherapyRow newTherapy(TherapyRow thRow) throws OHServiceException {
...
 	public boolean newTherapies(ArrayList<TherapyRow> thRows) throws OHServiceException {
 		if (!thRows.isEmpty()) {
-			DateTime now = new DateTime();
-
 
 			int patID = thRows.get(0).getPatient().getCode();
-			// FIXME FIXME			ioOperations.deleteAllTherapies(patID);
-			smsOp.deleteByModuleModuleID("therapy", String.valueOf(patID));
+			Patient patient = patientManager.getPatientById(patID);
+			ioOperations.deleteAllTherapies(patient);
```
mwithi, line 193: Only the comment was wrong, the operation was correct! 😅

mwithi (changes requested):
Only the comment was wrong...

src/main/java/org/isf/therapy/manager/TherapyManager.java
dbmalkovsky: OK; let us try it one more time.

mwithi (approved):

mwithi:
LGTM
```
Results :

Tests run: 1059, Failures: 0, Errors: 0, Skipped: 1
```

PR #311, opened 2021-01-30
OP-422 use containsKey()

dbmalkovsky:
Instead of enumerating the key set use `containsKey()` instead.

See OP-422.

Found with IntelliJ's Analyze plugin looking for redundant code.


mwithi (approved):
Thanks a lot!

PR #312, opened 2021-01-30
OP-423 remove unneeded toString()

dbmalkovsky:
There is no need to call `toString()` on a field/variable that is already a `String` type.

See OP-423.

Found with IntelliJ's Analyze plugin looking for redundant code.


mwithi (approved):
Thanks!

PR #313, opened 2021-02-02
Chore: Use StandardCharsets.UFT_8

dbmalkovsky:
Replace constant charset String literal  (like "UTF-8")  with a predefined Charset object like StandardCharsets.UTF_8. This may work a little bit faster, because charset lookup becomes unnecessary.

Found with IntelliJ's Analyze plugin.


mwithi (approved):
Tested

image

PR #315, opened 2021-02-02
Chore: Logger placeholder not available for `cause`

dbmalkovsky:
Just use the ` void error(String var1, Throwable var2);` method of the `org.slf4j.Logger` class.

Found with IntelliJ's Analyze plugin


mwithi (approved):
Tested 

image

PR #320, opened 2021-02-08
OP-434 use parameterized tests for MATERNITYRESTARTINJUNE

dbmalkovsky:
Use parameterized testing method for MATERNITYRESTARTINJUNE with `true` and `false` values.

### DIFF
--- a/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
+++ b/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
@@ -277,6 +277,7 @@ public class AdmissionIoOperations
 	 * The variables, {@code testing} and {@code afterJune}, are here only for testing purposes and are **NOT** to be used
 	 * in production code.   The default path ({@code testsing == false}) ensures that the code performs as it
 	 * always has in the past.
+	 * This code permits the unit testing of maternity wards with dates before and after June.
 	 * TODO: once the GregorianCalendar object is replaced by Java 8+ date/time objects this can be revisited
 	 * as there is more flexibility in modifying the new objects in Java 8+.
 	 */
--- a/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
+++ b/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
@@ -273,6 +273,13 @@ public class AdmissionIoOperations
 		return next;
 	}
 
+	/**
+	 * The variables, {@code testing} and {@code afterJune}, are here only for testing purposes and are **NOT** to be used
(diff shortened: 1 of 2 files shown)


src/test/java/org/isf/admission/test/Tests.java
```diff
@@ -147,6 +166,14 @@ public void setUp() {
 		cleanH2InMemoryDb();
 	}
 
+	@Parameterized.Parameters(name ="Test with MATERNITYRESTARTINJUNE={0}")
+	public static Collection<Object[]> automaticlot() {
```
mwithi, line 170: wrong variable name

mwithi (changes requested):
One variable has a wrong name.

Anyway I expect that some tests will fail because it is not always true that getNextYProg() returns YProg + 1.

Based on the MATERNINTYSTARTINJUNE flag it will if the admission date is before june, otherwise it will restart by 1 if it is the first admission in june.

I don't think we were testing in this way before but we should.

Thanks!

dbmalkovsky:
I believe the current code will work for both values of MATERNINTYSTARTINJUNE now through the end of May of this year and then it will fail.

In AdmissionIoOperations.java the code is comparing **now** to **June**:
````
if (wardId.equalsIgnoreCase("M") && GeneralData.MATERNITYRESTARTINJUNE) 
   {
	if (now.get(Calendar.MONTH) < 6) 
````

I'll work on resetting the calendar in the test 😄

**Comment by dbmalkovsky** [comment]
I tried multiple ways to manipulate the dates for MATERNITYRESTARTINJUNE testing but the only way that seemed to work is by adding a new method and `public static` variables specific for testing in `AdmissionIoOperations`.     In the process I found that the original tests for `getNextYProg()` never found any records (because of date issues) and always just returned `1`.

It might not be elegant but it works.

File: src/main/java/org/isf/admission/service/AdmissionIoOperations.java [hunk 2]
```diff
@@ -273,6 +273,21 @@ public int getNextYProg(
...
+			return now;
+		}
+		// testing date June or later
+		if (afterJune) {
+			return new GregorianCalendar(now.get(Calendar.YEAR), 8, 1);
+		}
+		// testing data before June
+		return new GregorianCalendar(now.get(Calendar.YEAR), 0, 1);
+	}
+
```
mwithi, line 290: we could add Javadoc to explain the testing purpose

mwithi (approved):
maybe some Javadoc could help.
it's ok otherwise.
```
Results :

Tests run: 1164, Failures: 0, Errors: 0, Skipped: 2
```

PR #324, opened 2021-02-11
OP-435 tests with LABEXTENDED GeneralData

dbmalkovsky:
Use parameterized testing method for LABEXTENDED and LABMULTIPLEINSERT  with all combinations of `true` and `false` values.

### DIFF
--- a/src/test/java/org/isf/lab/test/TestLaboratory.java
+++ b/src/test/java/org/isf/lab/test/TestLaboratory.java
@@ -27,6 +27,7 @@ import java.util.Calendar;
 import java.util.GregorianCalendar;
 
 import org.isf.exa.model.Exam;
+import org.isf.generaldata.GeneralData;
 import org.isf.lab.model.Laboratory;
 import org.isf.patient.model.Patient;
 import org.isf.utils.exception.OHException;
@@ -79,13 +80,21 @@ public class TestLaboratory {
 	}
 
 	public void check(Laboratory laboratory) {
-		assertThat(laboratory.getAge()).isEqualTo(age);
+		// If GeneralData.LABEXTENDED is true then the age found in the patient record is
+		// copied into the Laboratory record and as the age in the patient record changes
+		// based on when (what day and year) the test is run then a comparison to a fixed
+		// value almost always fails (except for late in 2021 and early in 2022).
+		// Likewise the patient name is copied from the Patient record which does not
+		// match the default in the Laboratory record.
+		if (!GeneralData.LABEXTENDED) {
+			assertThat(laboratory.getAge()).isEqualTo(age);
+			assertThat(laboratory.getPatName()).isEqualTo(patName);
+		}
 		assertThat(laboratory.getDate()).isEqualTo(registrationDate);
 		assertThat(laboratory.getExamDate()).isEqualTo(examDate);
 		assertThat(laboratory.getInOutPatient()).isEqualTo(InOutPatient);
 		assertThat(laboratory.getMaterial()).isEqualTo(material);
 		assertThat(laboratory.getNote()).isEqualTo(note);
-		assertThat(laboratory.getPatName()).isEqualTo(patName);
 		assertThat(laboratory.getResult()).isEqualTo(result);
 		assertThat(laboratory.getSex()).isEqualTo(sex);
 	}
--- a/src/test/java/org/isf/lab/test/Tests.java
+++ b/src/test/java/org/isf/lab/test/Tests.java
@@ -23,9 +23,12 @@ package org.isf.lab.test;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assumptions.assumeThat;
 
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.GregorianCalendar;
 import java.util.List;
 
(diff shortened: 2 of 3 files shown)


dbmalkovsky:
@mwithi The failure in the JDK 8 build was not in the code but rather it looks like a communications issue when trying to build/deploy the jar.   I am going to force a rebuild.

src/test/java/org/isf/lab/test/TestLaboratory.java
```diff
@@ -79,13 +80,21 @@ public void _setParameters(Laboratory laboratory, Exam exam, Patient patient) {
...
+		// If GeneralData.LABEXTENDED is true then the age found in the patient record is
+		// copied into the Laboratory record and as the age in the patient record changes
+		// based on when (what day and year) the test is run then a comparison to a fixed
+		// value almost always fails (except for late in 2021 and early in 2022).
+		// Likewise the patient name is copied from the Patient record which does not
+		// match the default in the Laboratory record.
+		if (!GeneralData.LABEXTENDED) {
+			assertThat(laboratory.getAge()).isEqualTo(age);
+			assertThat(laboratory.getPatName()).isEqualTo(patName);
+		}
```
mwithi, line 92: I think a good test should be that if LABEXTENDED=false, setPatient() on [Line77](https://github.com/informatici/openhospital-core/pull/324/files#diff-8b38acca6010318ede2951d248b94f0f33e2cb88a01787f94eb8d8f5f930dc53R77) should not be called, hence the only assertions about values should rely on the properties defined above in this class (as it is done here). Sex field should also be included.

Otherwise, if the Patient is set (LABEXTENDED=true), then those properties should be overridden when persisting the Laboratory, beside maintaining the link with the Patient's entity (through the field LAB_PAT_ID).

**Note**: overriding with Patient's record should happen only when persisting a new Laboratory record, but it should not happen when updating an old laboratory exam, this because for statistical purposes we need a snapshot of "age" and "sex" at the time of the exam recording.

src/test/java/org/isf/lab/test/Tests.java
```diff
@@ -84,6 +100,11 @@
 	@Autowired
 	private ApplicationEventPublisher applicationEventPublisher;
 
+	public Tests(boolean labExtended, boolean labMultipleInsert) {
+		GeneralData.LABEXTENDED = labExtended;
+		GeneralData.LABMULTIPLEINSERT = labMultipleInsert;
```
mwithi, line 105: `labMultipleInsert` is defined but never used. Maybe this property is not really needed in "core" component (but only in "gui" to drive the interface selection).

Moreover I found that LabManager needs some code review and optimization that makes again LABMULTIPLEINSERT  quite useless in the "core" component (saving a list or saving a single element list should be the same operation)

mwithi (commented):
Pleae find some comments in the PR, thanks a lot!

src/test/java/org/isf/lab/test/Tests.java
dbmalkovsky: Removed `LABMULTIPLEINSERT` and added a note that it is only used in the GUI.

I will leave the code review in `LabManager` to those that know much more than I.

src/test/java/org/isf/lab/test/TestLaboratory.java
dbmalkovsky: I am not sure I understand all the options here but I did add an `insert` flag to the validation code so that the laboratory values are only overridden with the patient information when doing an insert and not an update.

dbmalkovsky:
@mwithi I made some changes (to the best that I understood) based on your comments.   Please re-review.  Thanks.

src/test/java/org/isf/lab/test/TestLaboratory.java
mwithi: Ok, I understand now that the main problem lies in the `validate()` method, because it should only validate the object, not setting its properties... what do you think if we move that logic outside the method (but still in the manager, to not affect the "gui" at this stage) and so removing the flag? In this way also the tests would be simplified, wdyt?

If so, same changes could be applied for Opd module.

dbmalkovsky: If I understand; you want to move the setting out of the validation routine to some other routine but wouldn't it still have to be called from the `insert` method to ensure consistency with the Lab model object and the corresponding Patient model object?   How does moving it really fix the problem?  Or are you saying the GUI would be the only one calling the new routine?  And if so who is the GUI going to persist the "possible" changes in value back to the database?

dbmalkovsky: And yes a method with the name `validation` shouldn't be enforcing some `consistency`.

mwithi: I mean, that the setting out could be moved from `validation` method to `newLaboratory`s methods (that are called by GUI so it will not see any change) and so the consistency of validation method is about checking only and not setting anything. In this way, the passed `laboratory` object is exactly the one to be persisted (after validation); in this sense, tests should just take care about laboratory object construction and expect its correct persistence. Make sense?

dbmalkovsky: @mwithi OK; I created a new method, `setPatientConsistency(Laboratory laboratory)`, by moving the setting of name, age, and sex out of the `validateLaboratory()` routine.  The new method is added to the `newLaboratory()` and `newLaboratory2()`.    So, I think this covers all the remaining concerns.

mwithi: Good! I think now is more linear. Thanks a lot!

src/test/java/org/isf/lab/test/Tests.java
mwithi: Thanks!

mwithi:
All good. Thanks a lot!
```
Results :

Tests run: 1245, Failures: 0, Errors: 0, Skipped: 4
```

PR #325, opened 2021-02-12
Removed not needed native libs

mwithi:
Attempt to remove no longer needed native libraries


mizzioisf:
Yest ! you can pull this alltogether, thanks,

PR #326, opened 2021-02-12
Chore: use parameterized logging message instead of concatenation

dbmalkovsky:
Use more efficient parameterized logging message instead of concatenation; this was just introduced with a recent push.


mwithi (approved):

mwithi:
Thanks!

PR #327, opened 2021-02-14
OP-439 Increase coverage and format Model file

dbmalkovsky:
Coverage before:
image

Coverage after:
image

Also improvements to `equals()` routine to handle more `null` cases.

### DIFF
--- a/src/test/java/org/isf/patvac/test/Tests.java
+++ b/src/test/java/org/isf/patvac/test/Tests.java
@@ -193,7 +193,7 @@ public class Tests extends OHCoreTestCase {
 	public void testIoGetProgYear() throws Exception {
 		int prog_year = 0;
 		int found_prog_year = 0;
-		_setupTestPatientVaccine(false);
+		_setupTestPatientVaccine(true);
 		List<PatientVaccine> patientVaccineList = patvacIoOperation.getPatientVaccine(null, null, null, null, 'A', 0, 0);
 		for (PatientVaccine patVac : patientVaccineList) {
 			if (patVac.getProgr() > found_prog_year)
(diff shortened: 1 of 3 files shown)


src/test/java/org/isf/patvac/test/Tests.java
```diff
@@ -215,6 +221,305 @@ public void testListenerShouldUpdatePatientToMergedWhenPatientMergedEventArrive(
...
+		patientVaccine.setVaccineDate(new GregorianCalendar());
+
+		vaccineTypeIoOperationRepository.saveAndFlush(vaccineType);
+		vaccineIoOperationRepository.saveAndFlush(vaccine);
+		patientIoOperationRepository.saveAndFlush(patient);
+		patVacIoOperationRepository.saveAndFlush(patientVaccine);
+
+		ArrayList<PatientVaccine> patientVaccines = patVacManager.getPatientVaccine(false);
+		assertThat(patientVaccines.get(patientVaccines.size() - 1).getPatName()).isEqualTo(patientVaccine.getPatName());
+	}
```
mwithi, line 240: Could be bettere to enforce the "Today" word in the method's signature by persisting at least two objects in different dates?

src/test/java/org/isf/patvac/test/Tests.java
```diff
@@ -215,6 +221,305 @@ public void testListenerShouldUpdatePatientToMergedWhenPatientMergedEventArrive(
...
+		patientVaccine.setVaccineDate(date);
+
+		vaccineTypeIoOperationRepository.saveAndFlush(vaccineType);
+		vaccineIoOperationRepository.saveAndFlush(vaccine);
+		patientIoOperationRepository.saveAndFlush(patient);
+		patVacIoOperationRepository.saveAndFlush(patientVaccine);
+
+		ArrayList<PatientVaccine> patientVaccines = patVacManager.getPatientVaccine(true);
+		assertThat(patientVaccines.get(patientVaccines.size() - 1).getPatName()).isEqualTo(patientVaccine.getPatName());
+	}
```
mwithi, line 260: Same here, could it be better to have one object inside and an other one outside the range to be retrieved?

src/test/java/org/isf/patvac/test/Tests.java
dbmalkovsky: Good idea; added second object.

src/test/java/org/isf/patvac/test/Tests.java
dbmalkovsky: Good idea; added second object.

mwithi:
LGTM.
```
Results :

Tests run: 1703, Failures: 0, Errors: 0, Skipped: 3
```

PR #329, opened 2021-02-18
OP-443 fix NPE in utils.FileTools

dbmalkovsky:
See description in OP-443.

In addition added a test file for some coverage of the class and made minor code updates/improvements.

### DIFF
--- a/src/main/java/org/isf/utils/file/FileTools.java
+++ b/src/main/java/org/isf/utils/file/FileTools.java
@@ -41,7 +41,7 @@ public class FileTools {
 
 	private static final String[] BINARY_UNITS = { "B", "M", "G" }; //Byte, Megabyte, Gigabyte 
 
-	protected static final String[][] dateTimeFormats = new String[][] {
+	private static final String[][] dateTimeFormats = new String[][] {
 			{ "yyyy-MM-dd", "\\d{4}-\\d{2}-\\d{2}" },
 			{ "dd-MM-yyyy", "\\d{2}-\\d{2}-\\d{4}" },
 			{ "dd-MM-yy", "\\d{2}-\\d{2}-\\d{2}" },
@@ -55,7 +55,7 @@ public class FileTools {
 			{ "yyyy-MM-dd_HHmm", "\\d{4}-\\d{2}-\\d{2}_\\d{4}" },
 	};
 
-	public FileTools() {
+	private FileTools() {
 	}
 
 	/**
@@ -97,7 +97,7 @@ public class FileTools {
 	 * @return the list of retrieved date (first null)
 	 */
 	public static List<Date> getTimestampFromName(String formattedString) {
-		List<Date> datesFound = new ArrayList<Date>();
+		List<Date> datesFound = new ArrayList<>();
 		return getTimestampFromName(formattedString, datesFound);
 	}
 
@@ -119,7 +119,6 @@ public class FileTools {
 					datesFound.add(date);
 				}
 			} catch (ParseException e) {
-				date = null;
 			}
 		}
 
@@ -135,7 +134,7 @@ public class FileTools {
 		}
 
 		final int exponent = (int) (Math.log(bytes) / Math.log(base));
-		final String unit = BINARY_UNITS[exponent];
+		final String unit = BINARY_UNITS[exponent-1];
 		return String.format(locale, "%.1f %s", bytes / Math.pow(base, exponent), unit);
 	}
 
@@ -163,27 +162,4 @@ public class FileTools {
 		return size;
 	}
 
-	/**
-	 * @param args
-	 */
-	public static void main(String[] args) {
-		System.out.println("1. 09-03-2020      -> " + FileTools.getTimestampFromName("09-03-2020"));
-		System.out.println("2. 01-04-2020 1238 -> " + FileTools.getTimestampFromName("01-04-2020 1238"));
-		System.out.println("3. 02-05-2020_1122 -> " + FileTools.getTimestampFromName("02-05-2020_1122"));
-		System.out.println("4. 03-06-2020      -> " + FileTools.getTimestampFromName("03-06-2020"));
-		System.out.println("5. 04/03/2020      -> " + FileTools.getTimestampFromName("04/03/2020"));
-		System.out.println("6. 05/05/20        -> " + FileTools.getTimestampFromName("05/05/20"));
-		System.out.println("7. 2021-12-22 1100 -> " + FileTools.getTimestampFromName("2021-12-22 1100"));
-		System.out.println("8. 21-11-21_0922   -> " + FileTools.getTimestampFromName("21-11-21_0922"));
-
-		try {
-			System.out.println("1. 1024B -> " + FileTools.humanReadableByteCountParse("1024B"));
-			System.out.println("2. 4M   -> " + FileTools.humanReadableByteCountParse("4M"));
-			System.out.println("2. 16M   -> " + FileTools.humanReadableByteCountParse("16M"));
-			System.out.println("3. 256M  -> " + FileTools.humanReadableByteCountParse("256M"));
-			System.out.println("4. 1G    -> " + FileTools.humanReadableByteCountParse("1G"));
-		} catch (OHException e) {
-			e.printStackTrace();
-		}
-	}
 }
--- a/src/main/java/org/isf/utils/file/FileTools.java
+++ b/src/main/java/org/isf/utils/file/FileTools.java
@@ -31,6 +31,7 @@ import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.StringUtils;
 import org.isf.generaldata.MessageBundle;
 import org.isf.utils.exception.OHException;
 import org.slf4j.Logger;
@@ -104,7 +105,7 @@ public class FileTools {
 	 * @param file
 	 * @return the list of retrieved date or <code>null</code> if nothing is found
 	 */
-	public static Date getTimestampFromName(File file) {
+	public static List<Date> getTimestampFromName(File file) {
 		return getTimestampFromName(file.getName());
 	}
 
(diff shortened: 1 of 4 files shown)


dbmalkovsky:
This passes locally in my IDE and using mvn

Results :

Tests run: 1873, Failures: 0, Errors: 0, Skipped: 5

[INFO] BUILD SUCCESS
[INFO] Total time:  35.555 s
[INFO] Finished at: 2021-02-18T06:50:21-05:00

I will need to investigate more.

dbmalkovsky:
`TestTimeTools` and `TestBCrypt` were added by mistake to this PR.    There is a bug in `TimeTools` that I will fix with another PR once this one is merged.

dbmalkovsky:
FWIW, someone who understands the intent of ` FileTools.getTimestampFromName()` needs to carefully review the tests.  I got them all to pass but I'm not always sure the results that are generated are the answers that are truly expected.

src/main/java/org/isf/utils/file/FileTools.java
```diff
@@ -119,7 +119,6 @@ public static Date getTimestamp(File file) {
 					datesFound.add(date);
 				}
 			} catch (ParseException e) {
-				date = null;
```
mwithi, line 122: maybe we can at least log that no dates patterns have been found in the filename

src/test/java/org/isf/utils/file/TestFileTools.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.TimeZone;
+
+import org.junit.Test;
+
+public class TestFileTools {
+
+	@Test
+	public void testGetTimestampFromName() throws Exception {
```
mwithi, line 37: the aim of the method is to correctly parse a date from tipical patterns in filenames (like from cameras, screenshots, downloads, ecc...). To test if the date also is correct is good, thus obvious.

mwithi (commented):
Just few comments

src/test/java/org/isf/utils/file/TestFileTools.java
dbmalkovsky: I think I understand the purpose but not so the implementation.  For example, parsing "2021-12-22 1100" returns four dates: 
````
Wed Dec 22 00:00:00 EST 2021
Wed Dec 21 00:00:00 EST 2022
Wed Dec 22 11:00:00 EST 2021
Wed Dec 22 11:00:00 EST 2021
````
I am not quite sure how that second entry matches to original string.

File: src/main/java/org/isf/utils/file/FileTools.java (hunk 1 above)
Comment (dbmalkovsky) [reply src/main/java/org/isf/utils/file/FileTools.java:122]: Added logging message.

Comment (mwithi) [reply src/main/java/org/isf/utils/file/FileTools.java:122]: Thanks!

File: src/test/java/org/isf/utils/file/TestFileTools.java (hunk 2 above)
Comment (mwithi) [reply src/test/java/org/isf/utils/file/TestFileTools.java:37]: It look like a bug... the entries are driven by the `dateTimeFormats` matrix, where I can see now there is a duplicated pattern `yyyy-MM-dd HHmm`, although this does not explain why the second match appear...

The idea is to prompt to the user a set of parsed data in order to choose one...

Comment (dbmalkovsky) [reply src/test/java/org/isf/utils/file/TestFileTools.java:37]: Removing the duplicated pattern reduces the output to:
````
Wed Dec 22 00:00:00 EST 2021
Wed Dec 21 00:00:00 EST 2022
Wed Dec 22 11:00:00 EST 2021
````
which is better but still confusing.

So the first entry is returned because it matches:
````
{ "yyyy-MM-dd", "\\d{4}-\\d{2}-\\d{2}" },
````
The third entry is returned because it matches:
````
{ "yyyy-MM-dd HHmm", "\\d{4}-\\d{2}-\\d{2} \\d{4}" },
````
The second entry is returned be it matches:
````
{ "dd-MM-yy", "\\d{2}-\\d{2}-\\d{2}" },
````

It appears that `match` is matching substrings.  If we add `\\z" (which means the end of the input string) to the end of all the regular expressions I think it will fix the problem.

Comment (dbmalkovsky) [reply src/test/java/org/isf/utils/file/TestFileTools.java:37]: I also removed the duplicated pattern and added a missing one: `dd-MM-yyyy HHmm`.

Comment (mwithi) [reply src/test/java/org/isf/utils/file/TestFileTools.java:37]: Actually (20)21-12-22 is also a correct date! Good catch!
Nevertheless, adding end of line is not correct because most of the time the timestamp is at the beginning of the filename, followed by some other sequence of characters, thus not always true... definitely the date could be at any position. 

I don't master regex but maybe we can exclude any other digit [1-9] before and after each regex?

File: src/main/java/org/isf/utils/file/FileTools.java [hunk 3]
```diff
@@ -46,17 +46,17 @@
...
+			{ "dd-MM-yyyy", "\\d{2}-\\d{2}-\\d{4}\\z" },
+			{ "dd-MM-yyyy HHmm", "\\d{2}-\\d{2}-\\d{4} \\d{4}\\z" },
+			{ "dd-MM-yy", "\\d{2}-\\d{2}-\\d{2}\\z" },
+			{ "dd/MM/yyyy", "\\d{2}/\\d{2}/\\d{4}\\z" },
+			{ "dd/MM/yy", "\\d{2}/\\d{2}/\\d{2}\\z" },
+			{ "yyyy-MM-dd HHmm", "\\d{4}-\\d{2}-\\d{2} \\d{4}\\z" },
+			{ "yyyy-MM-dd HHmmss", "\\d{4}-\\d{2}-\\d{2} \\d{6}\\z" },
+			{ "yyyy-MM-dd_HHmmss", "\\d{4}-\\d{2}-\\d{2}_\\d{6}\\z" },
+			{ "dd-MM-yy_HHmm", "\\d{2}-\\d{2}-\\d{2}_\\d{4}\\z" },
+			{ "yyyy-MM-dd_HHmm", "\\d{4}-\\d{2}-\\d{2}_\\d{4}\\z" },
```
mwithi, line 59: As I said in the comments I don't think it's a good idea because we cannot know where the date is placed in the filename (if as a suffix or a postfix) but maybe we can predict that no other digits [1-9] need to be found on a pattern's hips

dbmalkovsky: I made two changes:

1. Start a pattern at a word boundary; this allows the string to be in a larger context and eliminates things like "2021-12-22 1100" matching the string "dd-MM-yy" (which was real bad because it formatted the date so it wasn't even correct)
2. Only add the formatted date to the resulting array if it is not already in the list

mwithi: 1. the problem is not solved, because if I'm not wrong, it would skip cases like "_09-03-2020_" or "02-05-2020_1122". Unfortunately the underscore is a very common character and it would be taken as “non-word characters”. `\w` stands for `[A-Za-z0-9_]`, which includes also digits, so also cannot apply... I think we have to exclude just digits manually, with something like `\b^[1-9].....\b^[1-9]` but I could be wrong!! 
2. that is good!

mwithi: 1. sorry, I was totally wrong... found out that we could use the [look-around assertion](https://www.regular-expressions.info/lookaround.html) (match but not capture) so we could precede each regex with `(?<![1-9])` and end with `(?![1-9])`

make sense?

dbmalkovsky: I don't believe your solution works as if we try "2021-12-22" it matches:
````
{ "yyyy-MM-dd", "(?<![1-9])(\\d{4}-\\d{2}-\\d{2})(?![1-9])" },
{ "dd-MM-yy", "(?<![1-9])(\\d{2}-\\d{2}-\\d{2})(?![1-9])" },
````
which results in two dates:
````
  <[2021-12-22T00:00:00.000 (java.util.Date),
     2022-12-21T00:00:00.000 (java.util.Date)]>
````
Note that the second regex ignores the leading  "20" and treats it as "21-12-22".  So instead of 2021 the year is 2022.

Comment (mwithi) [reply src/main/java/org/isf/utils/file/FileTools.java]: it doesn't work but it should... maybe is not supported by `java.util.regex` like warned on https://regexr.com/ ?

anyway just trying...

image

Comment (mwithi) [reply src/main/java/org/isf/utils/file/FileTools.java]: 1. what about `(?<![0-9])` and end with `(?![0-9])` 😅

Test (using old `main()`):
```
7. 2021-12-22 1100 -> [Wed Dec 22 00:00:00 CET 2021, Wed Dec 22 11:00:00 CET 2021]
```

Comment (dbmalkovsky) [reply src/main/java/org/isf/utils/file/FileTools.java]: Thanks; that last change seems to do the trick. 👍

**Comment by dbmalkovsky** [comment]
Added more tests to cover all the expressions.

I think we should consider ordering the regular expressions from longest to shortest and just return the first match instead of an array.  If the caller wants to ignore some part of the date then they can do that themselves.   It makes the method signature simpler and the tests too 😄

File: src/main/java/org/isf/utils/file/FileTools.java (hunk 3 above)
Comment (mwithi) [reply src/main/java/org/isf/utils/file/FileTools.java]: nice to hear that!!! thanks a lot!

**Comment by mwithi** [comment]
> I think we should consider ordering the regular expressions from longest to shortest and just return the first match instead of an array. If the caller wants to ignore some part of the date then they can do that themselves. It makes the method signature simpler and the tests too 😄

Also, sounds fine... we should add anyway the following:
```
{ "yyyyMMdd HHmm", "(?<![0-9])(\\d{4}\\d{2}\\d{2} \\d{4})(?![0-9])" },
{ "yyyyMMdd_HHmm", "(?<![0-9])(\\d{4}\\d{2}\\d{2}_\\d{4})(?![0-9])" },
{ "yyyyMMdd HHmmss", "(?<![0-9])(\\d{4}\\d{2}\\d{2} \\d{6})(?![0-9])" },
{ "yyyyMMdd_HHmmss", "(?<![0-9])(\\d{4}\\d{2}\\d{2}_\\d{6})(?![0-9])" },
```

Tests are ok:
```
Results :

Tests run: 1931, Failures: 0, Errors: 0, Skipped: 7
```

dbmalkovsky:
Ordered the tests from longest to shortest; return the first match instead of an array, add the 4 new patterns that were suggested, and improved the tests.

mwithi (approved):
Approved!

PR #330, opened 2021-02-19
OP-368 Changes Operations list order in Prices Browser from Descending to Ascending

GraninNikita:
Linked Bug in Jira: https://openhospital.atlassian.net/browse/OP-368


mwithi (approved):
Thanks a lot!

PR #332, opened 2021-02-20
OP-444 reorder code to avoid NPE

dbmalkovsky:
See OP-444 for description of the issue.

Also took the opportunity to reformat and improve the method.

This was found using SpotBugs.


mwithi (approved):
Good catch! Thanks!

PR #333, opened 2021-02-20
Chore instanceOf returns false for null

dbmalkovsky:
No need to test for `null` because `instanceOf` always returns `false` for a `null` value.

Found with SpotBugs.


mwithi (approved):
Thanks!

PR #340, opened 2021-02-25
Chore: remove unused private methods

dbmalkovsky:
What started out as just removing private methods that existed in the base class that differed only by case: `cleanH2InMemoryDB` turned into a file cleanup exercise using SpotBugs, SonarLint, DeadCode, and IntelliJ Analyze.  

There really wasn't any substantive code changes.  The biggest changes are formatting and adding a test for the number of validation messages returned on Exceptions.    No all `assertThatThrownBy` have the test because some methods return a different number of errors based on GeneralData settings and this test is parametrized for various values of `AUTOMATICLOT`.

### DIFF
--- a/src/test/java/org/isf/medicalstock/test/Tests.java
+++ b/src/test/java/org/isf/medicalstock/test/Tests.java
@@ -33,6 +33,7 @@ import java.util.List;
 
 import javax.persistence.EntityManager;
 
+import org.assertj.core.api.Condition;
 import org.isf.OHCoreTestCase;
 import org.isf.generaldata.GeneralData;
 import org.isf.medicals.model.Medical;
@@ -61,6 +62,7 @@ import org.isf.supplier.test.TestSupplier;
 import org.isf.utils.db.DbJpaUtil;
 import org.isf.utils.exception.OHDataValidationException;
 import org.isf.utils.exception.OHException;
+import org.isf.utils.exception.OHServiceException;
 import org.isf.ward.model.Ward;
 import org.isf.ward.service.WardIoOperationRepository;
 import org.isf.ward.test.TestWard;
@@ -80,7 +82,6 @@ import org.springframework.test.context.junit4.rules.SpringClassRule;
 import org.springframework.test.context.junit4.rules.SpringMethodRule;
 import org.springframework.transaction.annotation.Transactional;
 
-
 @Transactional
 @RunWith(Parameterized.class)
 @ContextConfiguration(locations = { "classpath:applicationContext.xml" })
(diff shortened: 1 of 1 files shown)


src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1027,22 +1083,22 @@ public void testMovementToString() throws Exception {
...
 	}
 
 	@Test
 	public void testMovementEquals() throws Exception {
 		int code = _setupTestMovement(false);
 		Movement movement = movementIoOperationRepository.findOne(code);
 		assertThat(movement.equals(movement)).isTrue();
-		assertThat(movement.equals(null)).isFalse();
-		assertThat(movement.equals("someString")).isFalse();
+		assertThat(movement).isNotEqualTo(null);
```
emecas, line 1094: Just wondering in case I missed a previous discussion, have you agreed anytime before that this line (and equivalents) supposed to be more readable than `assertNotNull(movement)` ?

emecas (approved):
thanks @dbmalkovsky LGTM

```
Results :

Tests run: 1953, Failures: 0, Errors: 0, Skipped: 7

[INFO] BUILD SUCCESS
[INFO] Total time:  01:58 min
[INFO] Finished at: 2021-03-01T10:45:57-06:00

```

src/test/java/org/isf/medicalstock/test/Tests.java
dbmalkovsky: I was trying to test the `equals()` method and I wasn't sure what `assertNotNull()` used under the covers but I did know what `isNotEqualTo()` used and that was the method under test.

emecas: OK. I can see your point and for sure:  

     assertThat(movement).isNotEqualTo(null);

looks much better than:

    assertThat(movement.equals(null)).isFalse();

but, remembering old discussions about why the adoption of assertThat from assertj and how it would suppose to be beneficial, this is one of the cases where I don't think any of both beat the simplicity/readability of a:

     assertNotNull(movement) 

It is just my personal opinion :)

mwithi: am I wrong or `assertNotNull` has never been used up to now?

dbmalkovsky: Probably not but it is a very valid method

mwithi: I'm sure of that, so we want to introduce it or we merge as is this PR?

dbmalkovsky: I would merge it

mwithi: Alright, that's why I was verify the change's consistency...

PR #344, opened 2021-02-28
OP-454 non-static non-final upper-case names

dbmalkovsky:
See OP-454.
If a variable is a constant make it upper-case with static final.
If a variable is not a constant make it lower-case with Camel casing.

Not all occurrences were changed/fixed as classes like `GeneralData` defined `properties` as `public static <type> <UPPERCASE-NAME>`.   Any changes to the variable names bleeds into the GUI.


mwithi (approved):
Thanks!

PR #345, opened 2021-03-01
OP-456 Cannot fill database with demo data when using oh-core_database docker image

GraninNikita:
Adds file _my.cnf_ with mysql configuration to set **lower_case_table_names** to make database case-insensitive in docker environment.


mizzioisf:
Hi Nikita,
I propose https://github.com/informatici/openhospital-core/pull/347
to fix OP-456, so we don't need to add another configuration file.
What to you think ?

GraninNikita:
Hi mizzioisf, I'm totally fine with this PR.

mwithi:
I'm closing this PR, superseded by #347

PR #346, opened 2021-03-02
OP-457 use camel case names

dbmalkovsky:
See OP-457.

After this is committed to `develop` the corresponding [PR](https://github.com/informatici/openhospital-gui/pull/427) in the GUI project needs to be committed.


mwithi (approved):
Thanks!

PR #350, opened 2021-03-05
Delete lib/native/solaris directory

mizzioisf:
I don't really think there is anyone that might still be using Open Hospital on Solaris...if we ever had solaris users :-)


mwithi:
> I don't really think there is anyone that might still be using Open Hospital on Solaris...if we ever had solaris users :-)

Can you confirm is same as #325 ?

mwithi:
Closing this because included in #325

PR #357, opened 2021-03-09
OP-464 use isEmpty()

dbmalkovsky:
See OP-464


mwithi (approved):
Thanks! Let's wait a bit for the checks to be available

PR #358, opened 2021-03-09
chore: rename 'mysql' dir to 'sql' in scripts

mizzioisf:
To work with https://github.com/informatici/openhospital/pull/156


mizzioisf:
You should move all the scripts in the sql/ directory, not into the mysql/db directory.

"mysql/db" -> "sql"

PR #361, opened 2021-03-11
Chore: Remove old 'gsvideo_isf' and 'sal_isf' video libs

mwithi:
Removed old libraries for video module (webcam) as per OP-139 (**DONE**).

Kept only the ones for DICOM as per OP-347 (**TO DO**).


mizzioisf (approved):

mwithi:
Can I merge this?

PR #364, opened 2021-03-15
Update demo data 2021

mizzioisf:
Update demo data to 2021
Update sql file format
Data fixes and corrections


mwithi (approved):
Thanks a lot

PR #369, opened 2021-03-21
Op 473 email validation regex

dbmalkovsky:
See OP-473

The current validation routine simplifies the regex by making the matcher case-insensitive but changes nothing else (than formatting).  Given this setup all the new accompanying tests pass as the system currently exists.

Given the note in OP-473, there is a suggested alternative regex that make the tests in `testQuestionablePatterns` fail (as I think they should).

### DIFF
--- a/src/main/java/org/isf/utils/validator/EmailValidator.java
+++ b/src/main/java/org/isf/utils/validator/EmailValidator.java
@@ -27,11 +27,12 @@ import java.util.regex.Pattern;
 public class EmailValidator {
 
 	// Java email validation permitted by RFC 5322
-	// Current regex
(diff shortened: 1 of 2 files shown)


dbmalkovsky:
This PR could be merged as is but I'd really appreciate a ruling on changing the validation regex.

mwithi:
> This PR could be merged as is but I'd really appreciate a ruling on changing the validation regex.

I had a draft comment on OP-437 that has been lost, sorry... I was pointing out that email regex should be quite stable nowadays, can we specify any RFC reference to mention as it was done with the current? otherwise is fine for me.

mwithi:
Up!

mwithi (approved):
LGTM. Thanks a lot!
```
Results :

Tests run: 1957, Failures: 0, Errors: 0, Skipped: 5
```

PR #363, opened 2021-03-12
Chore: use more natural assertj syntax

dbmalkovsky:
Minor changes to use consistent and more natural **assertj** syntax across the tests.   Some of this style was adopted with the later tests and this PR goes back and makes the style consistent.

PR #373, opened 2021-03-24
Chore: use more natural assertj syntax (part 2)

dbmalkovsky:
This is the second (and final) part (previous PR #363) of minor changes to use consistent and more natural assertj syntax across the tests. Some of this style was adopted with the later tests and this PR goes back and makes the style consistent.


mwithi (approved):
LGTM. Thanks!

PR #375, opened 2021-03-31
Chore: update license plugin version to match gui and api projects

dbmalkovsky:
Same version change as dependabot PRs in [GUI](https://github.com/informatici/openhospital-gui/pull/454) and [API](https://github.com/informatici/openhospital-api/pull/114) projects.


mwithi (approved):
Thanks!

PR #377, opened 2021-04-08
OP-488 Implement patient search in oh-core

npasquetto:
The aim of this implementation is to add **basic search capabilities** on patient entity to support OH 20 UI requirements (https://openhospital.atlassian.net/browse/OP-323).

At the moment, supported search parameters are:

- birthdate
- address
- firstName
- secondName

This is [the first of two PR](https://openhospital.atlassian.net/browse/OP-487): the second one will be [on the oh-api project](https://openhospital.atlassian.net/browse/OP-489), but it _depends_ on this one.

### DIFF
--- a/src/main/java/org/isf/patient/manager/PatientBrowserManager.java
+++ b/src/main/java/org/isf/patient/manager/PatientBrowserManager.java
@@ -21,6 +21,12 @@
  */
 package org.isf.patient.manager;
 
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.isf.accounting.manager.BillBrowserManager;
 import org.isf.accounting.model.Bill;
 import org.isf.admission.manager.AdmissionBrowserManager;
@@ -35,9 +41,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.stereotype.Component;
 import org.springframework.util.StringUtils;
-
-import java.util.*;
-
 
 @Component
 public class PatientBrowserManager {
--- a/src/main/java/org/isf/patient/service/PatientIoOperationRepositoryImpl.java
+++ b/src/main/java/org/isf/patient/service/PatientIoOperationRepositoryImpl.java
@@ -125,7 +125,7 @@ public class PatientIoOperationRepositoryImpl implements PatientIoOperationRepos
 				predicates.add(cb.between(keyPath.as(Date.class), birthDateFrom, birthDateTo));
 			} else {
 				if (entry.getValue() instanceof String) {
-					predicates.add(cb.like(keyPath, (String) entry.getValue() + "%"));
+					predicates.add(cb.like(cb.lower(keyPath), like(((String) entry.getValue()).toLowerCase())));
 				}
 			}
 		}
--- a/src/main/java/org/isf/patient/service/PatientIoOperationRepositoryImpl.java
+++ b/src/main/java/org/isf/patient/service/PatientIoOperationRepositoryImpl.java
@@ -21,23 +21,27 @@
  */
 package org.isf.patient.service;
 
-import org.isf.patient.model.Patient;
-import org.springframework.transaction.annotation.Transactional;
-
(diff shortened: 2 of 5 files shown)


src/main/java/org/isf/patient/manager/PatientBrowserManager.java
```diff
@@ -21,11 +21,6 @@
  */
 package org.isf.patient.manager;
 
-import java.util.ArrayList;
```
dbmalkovsky, line 24: Please see the **Code style** section of the repo README file to enable the project's code style rules.  For example, package imports are not supported and import ordering is

- statics
- java
- javax
- org
- com

Thanks.

src/test/java/org/isf/patient/test/Tests.java
```diff
@@ -216,6 +223,17 @@ public void testIoGetPatientAll() throws Exception {
...
 
+	@Test
+	public void testIoGetPatientsByParams() throws Exception {
+		_setupTestPatient(false);
+		Map<String, Object> params = new HashMap<String, Object>();
+		params.put("firstName", "TestFirstN");
+		params.put("birthDate", new GregorianCalendar(1984, Calendar.AUGUST, 14).getTime());
+		params.put("address", "TestAddress");
+		ArrayList<Patient> patients = patientIoOperation.getPatients(params);
+		assertThat(patients.size() > 0).isTrue();
```
dbmalkovsky, line 234: A more idiomatic version would be:  **assertThat(patients.size()).isPositive();**

src/test/java/org/isf/patient/test/Tests.java
```diff
@@ -48,6 +48,13 @@
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 
+import java.io.File;
+import java.lang.reflect.Field;
+import java.util.*;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
dbmalkovsky, line 56: I don't believe all the imports added are actually used in the file, are they?  For example, where is **assertThatThrownBy** or **java.io.File**?

src/main/java/org/isf/patient/manager/PatientBrowserManager.java
npasquetto: Sorry, I thought I had the correct settings.
Now it should be ok.

src/test/java/org/isf/patient/test/Tests.java
npasquetto: Yes! I've modified it.

src/test/java/org/isf/patient/test/Tests.java
npasquetto: It seems to me that they are used.

src/main/java/org/isf/patient/service/PatientIoOperationRepositoryImpl.java
```diff
@@ -90,15 +91,48 @@
...
+		for (Map.Entry<String, Object> entry : params.entrySet()) {
+			Path<String> keyPath = patient.get(entry.getKey());
+
+			if (entry.getKey().equals("birthDate")) {
+				Date birthDateFrom = (Date) entry.getValue();
+				Date birthDateTo = new Date(birthDateFrom.getTime() + (1000 * 60 * 60 * 24));
+				predicates.add(cb.between(keyPath.as(Date.class), birthDateFrom, birthDateTo));
+			} else {
+				if (entry.getValue() instanceof String) {
+					predicates.add(cb.like(keyPath, (String) entry.getValue() + "%"));
```
mwithi, line 128: I think we should implement case insensitive `like` as done in `wordExistsInOneOfPatientFields` method, eventually reusing the same code

mwithi (changes requested):
To implement case insensitive `like`

mwithi (approved):
In the end:

`getPatientsByOneOfFieldsLike(String keyword)` searches a keyword in any field of a "hard-coded" set of fields

`getPatientsByParams(Map<String, Object> params)` searches a specific set of fields-values in AND relation each other

I think both methods make sense. Thanks a lot!

PR #379, opened 2021-04-12
Chore: use diamond notation

dbmalkovsky:
Replace type specification and use diamond notation that is available starting with Java 7.

Note:  this is in code pushed earlier today.


mwithi (approved):
Thanks!

PR #380, opened 2021-04-12
Chore: use List.sort() instead of Collections.sort()

dbmalkovsky:
Starting with JDK 8 `Collections.sort()` internally calls `List.sort()`.  To save an extra method call just use `List.sort()` directly.

Found with IntelliJ Analyze plugin.


mwithi (approved):
Thanks!

PR #383, opened 2021-04-17
Make configuration classes final as security pattern

mwithi:
In order to prevent configuration methods and properties ovverride, all configuration classes are modified as final


mwithi:
@dbmalkovsky wdyt?

dbmalkovsky:
LGTM.   I'm was surprised to learn that overriding the methods/properties was an issue.

mwithi:
not an issue really, only potentially... the aim is just better design

PR #391, opened 2021-05-01
OP-501 add getMnemonic()

dbmalkovsky:
See OP-501.


mwithi (approved):
Thanks a lot!

mwithi:
Merged!

Of course, if the .key bundle is defined properly, one could set the mnemonic directly, correct?

PR #392, opened 2021-05-03
OP-502 implement formatMessage() method

dbmalkovsky:
See OP-502 and javadoc in new method

### DIFF
--- a/src/main/java/org/isf/generaldata/MessageBundle.java
+++ b/src/main/java/org/isf/generaldata/MessageBundle.java
@@ -25,8 +25,6 @@ import java.text.MessageFormat;
 import java.util.Locale;
 import java.util.MissingResourceException;
 import java.util.ResourceBundle;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.swing.JOptionPane;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/generaldata/MessageBundle.java
```diff
@@ -42,6 +45,9 @@
 
 	private static ResourceBundle defaultResourceBundle = null;
 
+	private static String formatSpecifier = "%(\\d+\\$)?([-#+ 0,(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])";
```
mwithi, line 48: if the aim is to count the number of arg, wouldn't be simpler this kind of pattern?
image

Anyway it seems not working with Arabic...

mwithi (commented):
question on the regex

src/main/java/org/isf/generaldata/MessageBundle.java
dbmalkovsky: I argued with myself adding the check because (1) it made the routine slower to execute and (2) it would fail when the same substitutable was used more than once in the string [fake example:  `The variable "{0}" is wrong.  "{0}" cannot be blank.`]

Therefore I just removed it.   If they don't match up then the string will look wrong and it will be reported as a bug.

mwithi: simplicity has its convenience ;)
thank a lot @dbmalkovsky

PR #394, opened 2021-05-05
OP-504 Core changes associated with GUI changes

dbmalkovsky:
See OP-504 and associated GUI [PR ](https://github.com/informatici/openhospital-gui/pull/485).

Changes:
- Change ordering of where the Exit menu is located
- Use new `MessageBundle.formatMessage()` with new bundle key


sql/step_02_dump_menu.sql
```diff
@@ -34,7 +34,7 @@ INSERT INTO MENUITEM (MNI_ID_A,MNI_BTN_LABEL,MNI_LABEL,MNI_TOOLTIP,MNI_SHORTCUT,
  ('examlist1','Exams List','ExamsList','x','E','printing','org.isf.stat.gui.ExamsList1Launcher','N',1),
  ('exams','Exams','Exams','x','E','generaldata','org.isf.exa.gui.ExamBrowser','N',4),
  ('examtype','Exam Type','Exam Type','x','E','types','org.isf.exatype.gui.ExamTypeBrowser','N',5),
- ('exit','Exit','Exit','x','E','file','none','N',0),
+ ('exit','Exit','Exit','x','E','file','none','N',1),
```
mwithi, line 37: I would try to totally remove Exit button (not used much) and to move Users under "General Data", which I would rename to "Settings"; it should be possible to do it by only changing the script.
Just a note, we could do it in another taks

mwithi (approved):
LTGM, just one proposal in comment

sql/step_02_dump_menu.sql
dbmalkovsky: I think we can change the look via the scripts and a few tweaks; I'll see how big of a project it is.

Changing the "GeneralData" class name to "Settings" is something I'd consider off the table.   In the GUI project there are 1475 references to "GeneralData" and at more than 166 in the CORE project.

mwithi: Yeah sure, classes refactoring should be eventually done in another time;
I've tested new changes and look fine for me! Thanks!

PR #396, opened 2021-05-08
OP-505 handle NPE

dbmalkovsky:
See OP-505.

Also added try-with-resources so the `close()` is always handled even if there is an error.


mwithi (approved):
Very good! Thanks!

PR #399, opened 2021-05-19
OP-438 New HTTP Client for SMS Gateway

AndreiDodu:
refactoring (added feign) + added textbelt gateway


sql/step_68_upgrade_pat_photo_size.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE PATIENT_PROFILE_PHOTO CHANGE COLUMN PAT_PHOTO PAT_PHOTO MEDIUMBLOB NULL DEFAULT NULL;
```
dbmalkovsky, line 1: There is already a step_68; see the step_04_all_following_steps script which is also a conflict the way it is currently coded.

src/main/java/org/isf/sms/providers/common/CustomCommonDecoder.java
```diff
@@ -0,0 +1,32 @@
+package org.isf.sms.providers.common;
```
dbmalkovsky, line 1: Please include the standard license header to all new files.   Thanks.

src/main/java/org/isf/sms/providers/skebby/model/SckebbySmsRequest.java
```diff
@@ -0,0 +1,101 @@
+package org.isf.sms.providers.skebby.model;
+
+import java.io.Serializable;
```
dbmalkovsky, line 3: Imports should be placed after the license header.  Thanks

sql/step_68_upgrade_pat_photo_size.sql
mwithi: yeah I think there fork was not aligned, @AndreiDodu you should merge the current develop.
anyway why this change here? is not in the OP-438 scope I think...

src/main/java/org/isf/sms/providers/gsm/GSMGatewayService.java
```diff
@@ -125,97 +131,88 @@ public boolean initialize() {
...
 		} else {
 			LOGGER.error("COM PORT not found ({})!!!", port);
- 		}
+		}
 		return connected;
 	}
 
 	@Override
-	public boolean sendSMS(Sms sms, boolean debug) {
+	public boolean sendSMS(Sms sms) {
```
mwithi, line 139: the `debug` flag was useful for testing mocking the actual SMS sending, is there another way to do it that I don't see in this PR's code?

sql/step_68_upgrade_pat_photo_size.sql
AndreiDodu: Pardon, I made a mistake here with commits. I will solve it on Saturday/Sunday. I will correct also the other code reviews warnings made by @dbmalkovsky.

pom.xml
```diff
@@ -409,39 +400,91 @@
...
+					<artifactId>log4j</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
+					<artifactId>slf4j-log4j12</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
+					<artifactId>log4j-over-slf4j</artifactId>
+				</exclusion>
```
mwithi, line 446: something misterious here, feign seems only depending on slf4j in its [pom.xml](https://github.com/OpenFeign/feign/blob/master/pom.xml)... I will check on my local

mwithi: found out it would be sufficient to comment out `slf4j-log4j12` dependency and leave feign alone without exclusions, could you try?

sql/step_68_upgrade_pat_photo_size.sql
AndreiDodu: I returned to the old version. The new commit has only the SMS "module".

pom.xml
AndreiDodu: Hi @mwithi, as you suggested I tried to exclude only slf4j-log4j12 dependency and removed the exclusion of other libraries from feign, but I still have an exception when I try to un OH-GUI (_Caused by: java.lang.IllegalStateException: Detected both log4j-over-slf4j.jar AND bound slf4j-log4j12.jar on the class path, preempting StackOverflowError. See also http://www.slf4j.org/codes.html#log4jDelegationLoop for more details._). So that I returned to my version.

src/main/java/org/isf/sms/providers/gsm/GSMGatewayService.java
AndreiDodu: I restored the previous code.

src/main/java/org/isf/sms/providers/skebby/model/SckebbySmsRequest.java
AndreiDodu: I corrected as you suggested.

src/main/java/org/isf/sms/providers/common/CustomCommonDecoder.java
AndreiDodu: I corrected as you suggested.

mwithi:
Superseded by https://github.com/informatici/openhospital-core/pull/412

PR #400, opened 2021-05-20
OP-510 Core changes associated with GUI changes

dbmalkovsky:
See OP-510.

### DIFF
--- a/src/main/java/org/isf/lab/manager/LabManager.java
+++ b/src/main/java/org/isf/lab/manager/LabManager.java
@@ -85,50 +85,50 @@ public class LabManager {
 
 		// Check Exam Date
 		if (laboratory.getExamDate() == null) {
-			errors.add(new OHExceptionMessage("noExamDateError",
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
 					MessageBundle.getMessage("angal.lab.pleaseinsertavalidexamdate"),
 					OHSeverityLevel.ERROR));
 		}
 		// Check Patient
 		if (GeneralData.LABEXTENDED && laboratory.getPatient() == null) {
-			errors.add(new OHExceptionMessage("patientNullError",
-					MessageBundle.getMessage("angal.common.select.patient.msg"),
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
+					MessageBundle.getMessage("angal.common.pleaseselectapatient.msg"),
 					OHSeverityLevel.ERROR));
 		} else if (laboratory.getPatient() == null) {
 			String sex = laboratory.getSex().toUpperCase();
 			if (!(sex.equals("M") || sex.equals("F"))) {
-				errors.add(new OHExceptionMessage("invalidSexError",
+				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
 						MessageBundle.getMessage("angal.lab.pleaseinsertmformaleorfforfemale"),
 						OHSeverityLevel.ERROR));
 			}
 			if (laboratory.getAge() < 0) {
-				errors.add(new OHExceptionMessage("invalidAgeError",
+				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
 						MessageBundle.getMessage("angal.lab.insertvalidage"),
 						OHSeverityLevel.ERROR));
 			}
 		}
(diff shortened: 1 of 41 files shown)


src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -92,7 +92,7 @@ protected void validateLaboratory(Laboratory laboratory) throws OHDataValidation
 		// Check Patient
 		if (GeneralData.LABEXTENDED && laboratory.getPatient() == null) {
 			errors.add(new OHExceptionMessage("patientNullError",
```
mwithi, line 94: why in this case we left the `"patientNullError"` while in other cases we replaced with `MessageBundle.getMessage("angal.common.error.title")`?  According to my understanding the "title" would be useful as univocal "error code" in order to identifiy the exception, not like a common title for every error

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -63,7 +63,7 @@ protected void validateMovementWard(MovementWard mov) throws OHDataValidationExc
 		List<OHExceptionMessage> errors = new ArrayList<>();
 		if (description.isEmpty() && mov.isPatient()) {
 			errors.add(new OHExceptionMessage("descriptionPatientEmptyError",
-					MessageBundle.getMessage("angal.medicalstockwardedit.pleaseselectapatient"),
+					MessageBundle.getMessage("angal.common.select.patient.msg"),
```
mwithi, line 66: wouldn't be more consistent with the past something like "`angal.common.pleaseselectapatient.msg`" like other similar like "`angal.opd.pleaseselectadisease.msg`"?

mwithi (commented):
some comments

src/main/java/org/isf/lab/manager/LabManager.java
dbmalkovsky: Yes it should be but it wasn't my focus when I was changing to a common "select patient message"; then I do the lab dialog this will be accounted for.

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
dbmalkovsky: I was trying to reduce to the length of the keys; it seems to me that "please" in the key is superfluous, redundant, and unneeded.    My preference is to remove the "please" for **all** the keys.

mwithi: the key is the sentence itself, so it can give a exact idea of the value, this is useful for the translators in absence of context so they have not to jump to English back and forth to see the original value.
In other words, according to the key only, someone could translate it as "select patient" only.
Moreover, also in the code would be more clear the message given to the patient without go check the bundles.

> My preference is to remove the "please" for all the keys.

I think these are too many changes to be tracked in one single PR, we agree we can revise in further steps, no worries

dbmalkovsky: I will put the "please" back and remove the extra dots in the keys.

mwithi: Thanks @dbmalkovsky !

dbmalkovsky: I am doing this because of the dictates of the the project owners but I still believe this is not needed as having worked with translators for 35+ years as they know and understand the context better then one may think.

src/main/java/org/isf/lab/manager/LabManager.java
dbmalkovsky: I made all the title changes in the CORE project.

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
dbmalkovsky: Added the "please", removed the dots, and expanded the key value to better match the actual English message.

dbmalkovsky:
@mwithi I believe I've addressed your concerns in the review comments with this commit.

src/main/java/org/isf/lab/manager/LabManager.java
mwithi: Thanks!

mwithi:
Thanks a lot!

PR #406, opened 2021-05-31
OP-525 comment out "hibernate.hbm2ddl.auto"

dbmalkovsky:
See OP-525.

This removes the warning message from the startup log.


dbmalkovsky:
It appears that the CORE project requires this line and sets the value to **update** in the database properties file when running the tests.   Will need to re-evaluate and find another way to suppress the message when running the GUI client.

PR #434, opened 2021-06-30
Chore: set date to next day

dbmalkovsky:
Set the end of the day range to the next day as the SQL code checks all parts of the date/time; currently it is set to June30, 0 minutes, 0 seconds thus anything occurring on June 30 is not retrieved in the search.  Setting it to July 1 makes it look like midnight on June 30 instead of 1 second into the day.


mwithi (approved):
Thanks a lot!

PR #442, opened 2021-07-19
OP-582 Fix bills comparator when getting bills associated to payments

mwithi:
See OP-582.

Cc: @dbmalkovsky


dbmalkovsky:
LGTM.   BTW, I reported the problem back in February; see OP-451

PR #446, opened 2021-07-20
OP-215 examination module note info

mwithi:
See OP-215.

### DIFF
--- a/src/main/java/org/isf/examination/manager/ExaminationBrowserManager.java
+++ b/src/main/java/org/isf/examination/manager/ExaminationBrowserManager.java
@@ -100,11 +100,9 @@ public class ExaminationBrowserManager {
 	private String formatLastNote(PatientExamination lastPatientExamination) {
 		String note = lastPatientExamination.getPex_note();
 		if (!note.isEmpty()) {
-			StringBuilder sbNote = new StringBuilder();
-			sbNote.append(MessageBundle.getMessage("angal.examination.lastnote")).append(" ");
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/examination/manager/ExaminationBrowserManager.java
```diff
@@ -77,17 +78,37 @@ public PatientExamination getDefaultPatientExamination(Patient patient) {
...
-				lastPatientExamination.getPex_auscultation(), lastPatientExamination.getPex_note());
+				lastPatientExamination.getPex_auscultation(), 
+				formatLastNote(lastPatientExamination));
 		return newPatientExamination;
 	}
 
+	private String formatLastNote(PatientExamination lastPatientExamination) {
+		String note = lastPatientExamination.getPex_note();
+		if (!note.isEmpty()) {
+			StringBuilder sbNote = new StringBuilder();
```
dbmalkovsky, line 103: I see you are adding new keys to the resource bundle why not use something like

`
MessageBundle.formatMessage("angal.examination.lastnote.fmt.msg", 
           TimeTools.formatDateTime(lastPatientExamination.getPex_date(), null), lastPatientExamination.getPex_note());
`

where angal.examination.lastnote.fmt.msg = "Last Note {0}:\n{1}\n\n".

Then you don't have to worry about word order and concatenation in other languages.

mwithi: you're right, I was just doing in the old fashined way... thanks!

dbmalkovsky:
LGTM

PR #448, opened 2021-07-21
OP-587 Add data so "Choose a Medical" window has codes

dbmalkovsky:
See OP-587


mwithi (commented):
LGTM, maybe we could set codes in the form Type|Number (e.g., K202, D204)

dbmalkovsky:
I actually did that for some of them so the screen shots looked better; there were just too many to do by hand and I figured other things were more important.

PR #451, opened 2021-07-22
Chore: escape single quotes

dbmalkovsky:
For MessageFormat to produce a single quote it must be escaped in the original pattern.   So as not to have everybody remember to escape it, just do it in code.


mwithi (approved):
Thanks!

PR #453, opened 2021-07-23
OP-540-docker-image-sql-mode

mwithi:
See OP-540.

Cc: @dbmalkovsky

### DIFF
--- a/Dockerfile-ohdb
+++ b/Dockerfile-ohdb
@@ -4,7 +4,6 @@ ENV MYSQL_DATABASE=oh
 ENV MYSQL_ROOT_PASSWORD=root
 ENV MYSQL_USER=isf
 ENV MYSQL_PASSWORD=isf123
-ENV MYSQL_LOWER_CASE_TABLE_NAMES=1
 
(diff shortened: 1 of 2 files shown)


dbmalkovsky:
LGTM.   

BTW, it would be great to also set the option to ignore the case of table names.   The create scripts have the table name in upper case but the load_demo_data script uses lower case table names.   Database and table names are not case sensitive in Windows, and case sensitive in most varieties of Unix. So on Windows this isn't a problem but causes issues on Linux.  I had to modify the script to use the common demo data.

mwithi:
that option actually is in the `Dockerfile-ohdb`, but at this point it seems uneffective like the SQL_MODE
```
ENV MYSQL_LOWER_CASE_TABLE_NAMES=1
```
Going to investigate more...

mwithi:
@dbmalkovsky could you try now?

dbmalkovsky:
I removed my docker image and started from scratch:  `docker-compose up` and then ran `load_demo_data.sql` and then started up the client and run `StockCard` all without issue.   So it looks good for me.

mwithi:
Thanks!
(PS: I use `docker-compose up --build` to test changes to docker files)

PR #454, opened 2021-07-26
OP-591 disclose accounting admin features

mwithi:
See OP-591.

Cc: @dbmalkovsky


dbmalkovsky:
LGTM given that you create the bill and then add some payment to it.

Thanks

PR #459, opened 2021-07-27
OP-546-inconsistent-search-behavior-in-patient-browser

mwithi:
See OP-546.

Cc: @dbmalkovsky

### DIFF
--- a/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
+++ b/src/main/java/org/isf/admission/service/AdmissionIoOperations.java
@@ -107,7 +107,6 @@ public class AdmissionIoOperations
 			GregorianCalendar[] dischargeRange) throws OHServiceException {
 		return patientRepository.findByFieldsContainingWordsFromLiteral(searchTerms).stream()
 			.map(patient -> new AdmittedPatient(patient, repository.findOneByPatientAndDateRanges(patient, admissionRange, dischargeRange).orElse(null)))
-			.filter(admittedPatient -> admittedPatient.getPatient()!= null && admittedPatient.getAdmission() != null)
 			.collect(Collectors.toList());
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/admission/service/AdmissionIoOperations.java
```diff
@@ -96,9 +96,9 @@
 
 	/**
 	 * Returns all patients based on the applied filters.
-	 * @param admissionRange the patient admission range
-	 * @param dischargeRange the patient discharge range
-	 * @param searchTerms the search terms to use for filter the patient list, <code>null</code> if no filter have to be applied.
+	 * @param admissionRange (two-dimensions array) the patient admission dates range, both {@code null} if no filter have to be applied.
+	 * @param dischargeRange (two-dimensions array) the patient discharge dates range, both {@code null} if no filter have to be applied.
```
dbmalkovsky, line 100: This is a small improvement; change:

 `* @param admissionRange (two-dimensions array) the patient admission dates range, both {@code null} if no filter have to be applied.`
  `* @param dischargeRange (two-dimensions array) the patient discharge dates range, both {@code null} if no filter have to be applied.`

to 

`* @param admissionRange (two-dimensional array) the patient admission dates range, both {@code null} if no filter is to be applied.`
  `* @param dischargeRange (two-dimensional array) the patient discharge dates range, both {@code null} if no filter is to be applied.`

dbmalkovsky:
LGTM if a small javadoc change.

src/main/java/org/isf/admission/service/AdmissionIoOperations.java
mwithi: Fixed, thanks!

PR #465, opened 2021-08-17
OP-625 demo data cause NPE in admissions

mwithi:
See OP-625.


sql/load_demo_data.sql
```diff
@@ -24,11 +24,11 @@ DROP TABLE IF EXISTS `admission`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `admission` (
   `ADM_ID` int(11) NOT NULL AUTO_INCREMENT,
-  `ADM_IN` int(11) NOT NULL DEFAULT 0,
+  `ADM_IN` int(11) NOT NULL DEFAULT '0',
```
dbmalkovsky, line 27: has the dump format changed?   `ADM_IN` is an `int` yet the default value looks like a `string`: '0' not as an integer zero as before.

mwithi: I noticed that, I think it's fine, but I'm still investigating it...

mizzioisf: mwhiti you should use mariadb if possible.

mwithi: [here](https://stackoverflow.com/a/35514653) there is a good explanation, nothing to worry about I think... ​

dbmalkovsky: IMHO it makes the project look less professional as someone reading the codebase is going to scratch their head and wonder what is going on....they shouldn't have to do research and understand that there is data conversion going on under the covers.   Again just my opinion.

mwithi: IMHO `MySQL dump 10.13  Distrib 5.7.17, for Win64 (x86_64)` explain everything...

we'll use another client like mariadb (although I don't know if on Windows has same exact behavior)

mwithi: Done again with MariaDB for Windows (taken from portable OH), the default values use same syntax

mwithi: Can we merge this or should we stick only to linux-forged mysql dumps?

mizzioisf: I think we can merge and get back on this after release; I will work on the demo data anyway.

PR #467, opened 2021-08-17
enable sql file execution which starts with 7*

AndreiDodu:
problem: when I generate DB, the scripts that starts with prefix 7* are not executed


mwithi (commented):
Thanks!

mwithi:
Closing because the change is already on https://github.com/informatici/openhospital-core

PR #474, opened 2021-08-18
Chore: nullcheck of movement when variable previously dereferenced

dbmalkovsky:
Remove null check of variable `movement` as it was previously dereferenced a few lines (line 126) above.

Found with SpotBugs.


mwithi (approved):
Thanks!

PR #480, opened 2021-09-02
OP-344 Remove step_00_create_user.sql

mwithi:
See OP-344.

Instructions on how to create users are now [here](https://github.com/informatici/openhospital-doc/blob/develop/doc_admin/AdminManual.adoc#3-1-4-open-hospital-database-creation)


mizzioisf (approved):
Everything looks good !

PR #481, opened 2021-09-06
OP-149 Aligned model with relational object and code cleaning/improvement

mwithi:
See OP-149.

### DIFF
--- a/src/main/java/org/isf/visits/manager/VisitManager.java
+++ b/src/main/java/org/isf/visits/manager/VisitManager.java
@@ -140,6 +140,7 @@ public class VisitManager {
 			smsOp.deleteByModuleModuleID("visit", String.valueOf(patID));
 
 			for (Visit visit : visits) {
+				validateVisit(visit);
 
 				visit.setVisitID(0); //reset ID in order to persist again (otherwise JPA think data is already persisted)
 				int visitID = ioOperations.newVisit(visit).getVisitID();
--- a/src/main/java/org/isf/visits/manager/VisitManager.java
+++ b/src/main/java/org/isf/visits/manager/VisitManager.java
@@ -135,8 +135,11 @@ public class VisitManager {
 	}
 
 	/**
-	 * Inserts or replaces all {@link Visit}s related to a patID
-	 *
+	 * Inserts or replaces all {@link Visit}s related to a patID<br>
+	 * <br>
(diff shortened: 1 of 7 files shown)


src/test/java/org/isf/visits/test/TestVisit.java
```diff
@@ -35,7 +35,7 @@
 	private GregorianCalendar date = new GregorianCalendar(10, 9, 8);
 	private String note = "TestNote";
 	private boolean sms = true;
-	private String duration = "10";
+	private Integer duration = new Integer(10);
```
dbmalkovsky, line 38: This is unnecessary boxing; just use `private Integer duration = 10;`

src/test/java/org/isf/visits/test/Tests.java
```diff
@@ -307,9 +293,9 @@ public void testVisitGetSet() throws Exception {
 		int id = _setupTestVisit(false);
 		Visit visit = visitManager.findVisit(id);
 
-		assertThat(visit.getDuration()).isEqualTo("10");
-		visit.setDuration("");
-		assertThat(visit.getDuration()).isEmpty();
+		assertThat(visit.getDuration()).isEqualTo(new Integer(10));
```
dbmalkovsky, line 296: same note as above; just use `10`,

dbmalkovsky:
Other than the two small comments, LGTM.

src/test/java/org/isf/visits/test/TestVisit.java
mwithi: Right!

src/main/java/org/isf/visits/manager/VisitManager.java
```diff
@@ -56,6 +60,38 @@
...
 	@Autowired
 	private ApplicationContext applicationContext;
+	
+	/**
+	 * Verify if the object is valid for CRUD and return a list of errors, if any
+	 *
+	 * @param visit
+	 * @throws OHServiceException
+	 */
+	private void validateVisit(Visit visit) throws OHServiceException {
```
dbmalkovsky, line 70: I believe this must be made `public` as there is a call to this method in the GUI file `InsertVisit.java` at line 328

mwithi: Yep, forgot to push, thanks!

PR #483, opened 2021-09-13
Chore: remove unneeded temporary variable

dbmalkovsky:
No need to store result  of method call in a variable so it can then next be tested.

Also did a little re-formatting.


mwithi (approved):
Thanks!

PR #485, opened 2021-09-13
Chore: use Java5 features

dbmalkovsky:
Use Java 5 features:

enhanced for loop
contains()
StringBuilder() instead of StringBuffer()


mwithi (approved):
LGTM. Thanks!

PR #487, opened 2021-09-13
Chore: fix if test

dbmalkovsky:
The `getSupplier()` and `getWard()` will always return on object of the correct type or the value `null`.  The test should be for `null` indicating that a supplier or ward must be selected.


mwithi (approved):
LGTM. Thanks!

PR #493, opened 2021-09-21
Improve connection errors handling

mwithi:
Aim of this PR is to better handling connections issues from outside the application (return exitcodes)


src/test/resources/sms.properties
```diff
@@ -1,11 +1,40 @@
-# Configuration file for SMS Sender
```
dbmalkovsky, line 1: I'm curious, what is the purpose of the empty `src/test/resources/language.properties` file just above this?

mwithi: I think garbage from previous tests

PR #508, opened 2021-11-10
Update demo data - 2021.11

mizzioisf:
- Updated demo data
- Fixed names / gender / birth dates


dbmalkovsky:
@mizzioisf It would be very nice if there was a version of this file that had tens of thousands of patients (say around 50,000) with corresponding number of exams, admissions, bills, etc. for testing :)

mizzioisf:
> @mizzioisf It would be very nice if there was a version of this file that had tens of thousands of patients (say around 50,000) with corresponding number of exams, admissions, bills, etc. for testing :)

I see and understand the need :-)
I think that can be done with a specific set of queries in order to create that amount of records; I'm absolutely not able to write that kind of SQL instructions, maybe @mwithi can help :-)

dbmalkovsky:
I've started looking at "generators" that might be of help; projects like [jFairy by Devskiller](https://github.com/Devskiller/jfairy) and [Random data generator for MySQL](https://github.com/Percona-Lab/mysql_random_data_load).   They tend to work well on a single table and not so well with a database like OH.

mwithi:
> They tend to work well on a single table and not so well with a database like OH.

That's the main reason why we mostly gave up low priority to this task, but we can create an issue/task for it.

mwithi (approved):
LGTM.

PR #513, opened 2021-11-22
OP-700 Fixing validation error (null ward means ODP)

mwithi:
See OP-700.

### DIFF
--- a/src/main/java/org/isf/visits/model/Visit.java
+++ b/src/main/java/org/isf/visits/model/Visit.java
@@ -39,6 +39,8 @@ import javax.persistence.Table;
 import javax.persistence.Transient;
 import javax.validation.constraints.NotNull;
 
+import org.apache.log4j.AppenderSkeleton;
+import org.isf.generaldata.MessageBundle;
 import org.isf.patient.model.Patient;
 import org.isf.utils.db.Auditable;
 import org.isf.ward.model.Ward;
@@ -76,7 +78,9 @@ public class Visit  extends Auditable<String>
 	@JoinColumn(name="VST_PAT_ID")
 	private Patient patient;
 	
-	@NotNull
+	/**
+	 * if {@code null} the visit is meant for OPD. 
+	 */
 	@ManyToOne
 	@JoinColumn(name="VST_WRD_ID_A")
 	private Ward ward;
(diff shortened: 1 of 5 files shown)


src/main/java/org/isf/visits/model/Visit.java
```diff
@@ -39,6 +39,8 @@
 import javax.persistence.Transient;
 import javax.validation.constraints.NotNull;
 
+import org.apache.log4j.AppenderSkeleton;
```
dbmalkovsky, line 42: What is this import used for?

src/main/java/org/isf/visits/service/VisitsIoOperations.java
```diff
@@ -52,7 +53,26 @@
...
 	}
+	
+	/**
+	 * Returns the list of all {@link Visit}s related to a patID in OPD (Ward is {@code null}).
+	 *
+	 * @param patID - the {@link Patient} ID. If <code>0</code> return the list of all OPD {@link Visit}s
+	 * @return the list of {@link Visit}s
+	 * @throws OHServiceException
+	 */
+	public ArrayList<Visit> getVisitsOPD(Integer patID) throws OHServiceException {
```
dbmalkovsky, line 64: Instead of `ArrayList` use `List`.

src/main/java/org/isf/visits/service/VisitsIoOperations.java
```diff
@@ -52,7 +53,26 @@
...
+		return patID != 0 ?
+				new ArrayList<>(repository.findAllByWardIsNullAndPatient_CodeOrderByPatient_CodeAscDateAsc(patID)) :
+				new ArrayList<>(repository.findAllByWardIsNullOrderByPatient_CodeAscDateAsc());
```
dbmalkovsky, line 67: And if you make the change noted above then you don't need `new ArrayList<>` because the repository method returns a `List`.

src/main/java/org/isf/visits/model/Visit.java
mwithi: really don't know why was there, fixed. Thanks

src/main/java/org/isf/visits/service/VisitsIoOperations.java
mwithi: Yep, thanks!

src/main/java/org/isf/visits/service/VisitsIoOperations.java
mwithi: correct! thanks!

PR #549, opened 2021-12-23
Chore: Update feign-gson as found by dependabot

dbmalkovsky:
Also changed the jar version from 1.11.99 -> 1.12.0

Will rebase dependabot when this is comitted.


mwithi (approved):
Thanks!

PR #550, opened 2021-12-23
Merge Hotfix after release

mwithi:


src/main/java/org/isf/dicom/service/DicomIoOperations.java
```diff
@@ -168,8 +169,12 @@ public boolean isCodePresent(Long code) throws OHServiceException {
 	 * @return <code>true</code> if the code is already in use, <code>false</code> otherwise
 	 * @throws OHServiceException
 	 */
-	public boolean isSeriePresent(String dicomSeriesNumber) throws OHServiceException {
-		List<FileDicom> result = repository.findAllWhereDicomSeriesNumber(dicomSeriesNumber);
-		return !result.isEmpty();
+	public boolean isSeriesPresent(String dicomSeriesNumber) throws OHServiceException
+	{
```
dbmalkovsky, line 173: Don't the braces usually go on the same line?

dbmalkovsky:
Other than a few formatting issues with braces; LGTM

src/main/java/org/isf/dicom/service/DicomIoOperations.java
mwithi: yep! done. Thank you!

mwithi:
> Other than a few formatting issues with braces; LGTM

Thanks!

> Please double check the version number in the POM to make sure it is what you want; thanks.

Yes, let's align all of them to the same *real* common version, then we see how to proceed. Thanks!

PR #554, opened 2021-12-25
Chore: add missing Override annotation

dbmalkovsky:
Found with IntelliJ Analyze


mwithi (approved):
LGTM. BTW maybe this method could be just removed since it doesn't add anything in respect Jpa... but we can keep the class (to be renamed `PricesListIoOperationRepository`) for simmetry with the rest of the code, wdyt?

dbmalkovsky:
How about this refactoring?

mwithi:
> How about this refactoring?

LGTM. So we keep the `deleteById` method?

dbmalkovsky:
For the time being until I am convinced it isn't needed 😄

mwithi:
> For the time being until I am convinced it isn't needed 😄

it should not, but in case I feel I would keep the class for simmetry... so let's keep all of it

PR #560, opened 2022-01-01
OP-332 replace iText with librepdf

dbmalkovsky:
See OP-332

JapserReports was defined both in CORE and GUI although only the fonts dependency was in GUI.  Moved them all together in this project and will remove from GUI pom.xml.


dbmalkovsky:
FWIW, the 4 unit tests failed because this is a new year 😢

dbmalkovsky:
To fix the tests had to set the time for the beginning of the range to zero and for the end of the range to 59:59

mwithi (approved):
Thanks a lot!

PR #564, opened 2022-01-07
Chore: remove redundant 'new' expression in constant array creation

dbmalkovsky:
Array initializers can omit the type because it is already specified in the left side of the assignment.


mwithi (approved):
Thanks!

PR #565, opened 2022-01-07
OP-742: fix method name to match the object in the package

dbmalkovsky:
See OP-742; correct a cut-n-paste error; there are no `vaccines` in the `dicom` package.


mwithi (approved):
Thanks!

PR #581, opened 2022-01-18
OP-746 Update default mysql driver to latest version

dbmalkovsky:
See OP-746.

Removes warning message from GUI startup log.


mwithi (approved):
Thanks!

PR #582, opened 2022-01-19
OP-747 remove extra message(s)

dbmalkovsky:
See OP-747

Simplify code and update tests for one less message.


mwithi (approved):
LGTM. Thanks a lot!

PR #586, opened 2022-01-20
OP-751 fix date conversions in reports

dbmalkovsky:
See OP-751

- add missing `convertToLegacyFormatForLibWithoutJavaTimeSupport()` method calls
- fix date patterns
- change method parameter from `false` to `true` when there was no time component


mwithi (approved):
LGTM. Thanks a lot!

PR #589, opened 2022-01-24
Chore Starting with Spring 5.3.0, isEmpty() was deprecated.

dbmalkovsky:
The easiest change is to use `isEmpty()` from `org.apache.commons.lang3` which is already included in the project and used in some of the code.

Also changed a few `!isEmpty()` to `isNotEmpty()`.


mwithi (approved):
LGTM. Thanks!

PR #594, opened 2022-01-26
Chore: update all outstanding dependabot jars

dbmalkovsky:
Combined all dependabot jar updates (except for H2) into one so all the unit tests in Core and Gui could be run and verified together incase there was some interaction.  Also did some interactive GUI testing after installing the jar.


mwithi (approved):
Thanks!

PR #595, opened 2022-01-30
OP-748: new method to only get operations with codes 1, 2, and 3

dbmalkovsky:
See OP-748


dbmalkovsky:
Closing as the wrong code base was used and missed a few cases.

PR #596, opened 2022-01-30
OP-748: new method to only get operations with codes 1, 2, and 3

dbmalkovsky:
See OP-748


mwithi (changes requested):
Filtering by OperationType fails
image

dbmalkovsky:
@mwithi Fix pushed

mwithi (approved):

mwithi:
> @mwithi Fix pushed

Thanks!

PR #597, opened 2022-02-02
OP-777 admission and discharge time

mwithi:
See OP-777.

Also re-enabled audit fields and updated demo data.


dbmalkovsky:
LGTM

PR #600, opened 2022-02-06
chore: bump h2 dependency to v2.x

pviotti:
[H2 v2.0 changelog ](https://github.com/h2database/h2database/releases/tag/version-2.0.202)- TLDR: they fixed lots of bugs

Additional changes to make the upgrade work:
 - add `NON_KEYWORDS=USER` - workaround to use `USER` as table name even if it's a keyword in H2 (see [this explanation](https://github.com/h2database/h2database/issues/3363#issuecomment-1010631797))
 - fixed `testTherapyRowToString` - that test (added in https://github.com/informatici/openhospital-core/commit/025b642f92ebdadd5f7cc78e435ab5af8967059a ) was failing also when executing it individually with H2 v1. When executing it with all the other tests it was passing, due, most probably, to some race condition bug in H2 v1.x that has been fixed in v2.x

replaces #580

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,7 +216,7 @@
 		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
-			<version>1.4.200</version>
+			<version>2.1.210</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


dbmalkovsky:
The real fix for the reserved table name is in OP-727.

dbmalkovsky:
LGTM as long as the workaround for the USER table is temporary.

dbmalkovsky (approved):

PR #609, opened 2022-02-20
Op 774 laboratory honoring date and time now

mwithi:
See OP-774.

Paired with https://github.com/informatici/openhospital-gui/pull/1138.

Will follow OP-781 after v1.11.3

### DIFF
--- a/src/main/java/org/isf/lab/model/Laboratory.java
+++ b/src/main/java/org/isf/lab/model/Laboratory.java
@@ -83,10 +83,11 @@ public class Laboratory extends Auditable<String>
 
 	@NotNull
 	@Column(name="LAB_DATE")
-	private GregorianCalendar registrationDate;
+	private GregorianCalendar labDate;
 	
 	@Column(name="LAB_EXAM_DATE")
 	@Temporal(TemporalType.DATE)
+	@Deprecated
 	private Calendar examDate;
 
 	@NotNull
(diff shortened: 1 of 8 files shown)


src/main/java/org/isf/lab/model/Laboratory.java
```diff
@@ -124,7 +125,7 @@ public Laboratory() { }
 	public Laboratory(Exam aExam,GregorianCalendar aDate,String aResult,
 			String aNote, Patient aPatId, String aPatName){
 		exam=aExam;
-		registrationDate=aDate;
+		labDate=aDate;
```
dbmalkovsky, line 128: Nitpick: the formatting in this and the next method could use some work.

src/main/java/org/isf/lab/model/Laboratory.java
```diff
@@ -165,14 +166,22 @@ public void setExam(Exam aExam){
...
+	/**
+	 * the field has been replaced by <code>createdDate()</code> and it's not meant to be managed by the user (Spring managed)
+	 */
+	@Deprecated
 	public void setExamDate(GregorianCalendar exDate) {
 		this.examDate = exDate;
 	}	
 	public void setDate(GregorianCalendar aDate){
-		registrationDate=aDate;
+		labDate=aDate;
```
dbmalkovsky, line 184: again more formatting here and the next routine

dbmalkovsky (approved):

src/main/java/org/isf/lab/model/Laboratory.java
mwithi: Fixed, thanks!

src/main/java/org/isf/lab/model/Laboratory.java
mwithi: Fixed, thanks!

PR #615, opened 2022-03-05
OP-818 fix subsetting by vaccine type value

dbmalkovsky:
See OP-818

Uses the hotfix branch as its base.


mwithi (approved):
LGTM. Thanks!

PR #619, opened 2022-03-07
Chore: combine and fix recent buildbot updates

dbmalkovsky:
Combined all the recent buildbot updates and tested them together

- Had to add `commons-lang3` dependency because jasperreports 6.19.0 removed it 
- Reorder some dependencies so the latest version of common jars where used and not excluded
- Change `slf4j-log4j12` to `slf4j-reload4j` which is the new name; see `Reload4j 1.2.18 as a replacement for log4j 1.2.17` at https://www.slf4j.org/log4shell.html

I will rebase/close the buildbot entries when this is committed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,7 +191,7 @@
 		<dependency>
 			<groupId>net.sf.jasperreports</groupId>
 			<artifactId>jasperreports</artifactId>
-			<version>6.18.1</version>
+			<version>6.19.0</version>
 			<exclusions>
 				<exclusion>
 					<groupId>com.lowagie</groupId>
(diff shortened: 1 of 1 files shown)


mwithi (commented):
Did you test also reports with new libraries?

dbmalkovsky:
No I did not.

pom.xml
```diff
@@ -304,7 +309,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
+			<artifactId>slf4j-reload4j</artifactId>
 			<version>1.7.36</version>
```
mwithi, line 313: Should this be done in gui also?

dbmalkovsky: yes but I was going to wait until this was this PR from buildbot was merged first: https://github.com/informatici/openhospital-gui/pull/1123

PR #620, opened 2022-03-07
OP-820 add visit parameters to settings

dbmalkovsky:
See OP-820


dbmalkovsky:
Closing for a suggested different implementation

PR #631, opened 2022-03-17
Resolve conflicts for #623

dbmalkovsky:
Paired with https://github.com/informatici/openhospital-gui/pull/1159


src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -240,10 +236,9 @@ protected void validateAdmission(Admission admission, boolean insert) throws OHS
 		List<Admission> admList = getAdmissions(patient);
 
 		/*
-		 * Today Gregorian Calendar
+		 * Today LocalDateTime
 		 */
-		GregorianCalendar today = TimeTools.getDateToday24();
-		DateFormat currentDateFormat = DateFormat.getDateInstance(DateFormat.SHORT, new Locale(GeneralData.LANGUAGE));
+		LocalDateTime today = LocalDateTime.now();
```
mwithi, line 241: The hotfix here was:
```
-		GregorianCalendar today = new GregorianCalendar();
+		GregorianCalendar today = TimeTools.getDateToday24();
```

So it seems the new `LocalDateTime today = LocalDateTime.now();` is rolling back the hotfix, isn't it?

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -382,15 +376,15 @@ else if (dateOut.before(ad.getAdmDate()) || dateOut.equals(ad.getAdmDate())) {
 						OHSeverityLevel.ERROR));
 			}
 
-			GregorianCalendar visitDate = admission.getVisitDate();
-			if (visitDate != null && ward.getCode().equalsIgnoreCase("M")) {
-				GregorianCalendar limit;
+			LocalDateTime visitDate = admission.getVisitDate();
+			if (operationDate != null && ward.getCode().equalsIgnoreCase("M")) {
+				LocalDateTime limit;
```
mwithi, line 381: The hotfix here was:
```
-			if (operationDate != null && ward.getCode().equalsIgnoreCase("M")) {
+			if (visitDate != null && ward.getCode().equalsIgnoreCase("M")) {
```
so it seems this PR is rolling back the hotfix, isn't it?

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -382,15 +376,15 @@ else if (dateOut.before(ad.getAdmDate()) || dateOut.equals(ad.getAdmDate())) {
...
 				if (admission.getDisDate() == null) {
 					limit = today;
 				} else {
 					limit = admission.getDisDate();
 				}
-				if (visitDate.before(dateIn) || visitDate.after(limit)) {
+				if (operationDate.isBefore(dateIn) || operationDate.isAfter(limit)) {
```
mwithi, line 387: Same as above... the hotfix was:
```
-				if (operationDate.before(dateIn) || operationDate.after(limit)) {
+				if (visitDate.before(dateIn) || visitDate.after(limit)) {
```

src/main/java/org/isf/utils/db/DbSingleConn.java
```diff
@@ -33,10 +33,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.mysql.jdbc.exceptions.jdbc4.CommunicationsException;
+import com.mysql.cj.jdbc.exceptions.CommunicationsException;
```
mwithi, line 36: Here the hotfix was:
```
-import com.mysql.cj.jdbc.exceptions.CommunicationsException;
+import com.mysql.jdbc.exceptions.jdbc4.CommunicationsException;
```

but it should temporary work in both ways I think...

mwithi (changes requested):
There are some hotfixex that seems overridden in this PR, please check!
Thanks a lot!

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: yep; fixed it; thanks

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: Good catch; fixed

src/main/java/org/isf/utils/db/DbSingleConn.java
dbmalkovsky: I'm going to leave it like it is; when we worry about JDBC jar versions we can visit if needed.

mwithi: ok

PR #657, opened 2022-05-02
OP-858 develop branch should use SNAPSHOT jar

dbmalkovsky:
See OP-858.

The GUI pom.xml file needs to be updated too.


mwithi (approved):
Thanks!

PR #678, opened 2022-06-06
OP-820 add visit start and end hour to hospital table

dbmalkovsky:
See OP-820. 

Once this is merged I will push the GUI changes that use the start/end hours for visits.
I defaulted the values to start at 6AM and end at 8PM to simplify the runtime checking.

_Note: added step_72_hospital_changes.sql_


src/main/java/org/isf/hospital/model/Hospital.java
```diff
@@ -81,6 +86,14 @@ public class Hospital extends Auditable<String> {
...
+	@Column(name="HOS_VISITS_START")
+	private Time visitStartTime;
+
+	@NotNull
+	@Column(name="HOS_VISITS_END")
+	private Time visitEndTime;
+
 	@Version
 	@Column(name="HOS_LOCK")
     private Integer lock;
```
mwithi, line 99: I know that only some lines are just unchanged, but it seems we have a problem of indentation in some files.

src/main/java/org/isf/hospital/model/Hospital.java
```diff
@@ -122,8 +137,27 @@ public Hospital(String aCode, String aDescription, String aAddress,
...
+		this.fax = aFax;
+		this.email = aEmail;
+		this.currencyCod = aCurrencyCod;
+		this.visitStartTime = visitStartTime;
+		this.visitEndTime = visitEndTime;
+	}
+
     public String getAddress() {
         return this.address;
     }
```
mwithi, line 163: same here

mwithi (approved):
LGTM. But is this PR still a WIP?

src/main/java/org/isf/hospital/model/Hospital.java
dbmalkovsky: I reformatted the whole file to eliminate any formatting issues

dbmalkovsky:
I consider this that it could be merged so we can move foreard.

mwithi:
> I consider this that it could be merged so we can move forward..

so we should maybe rephrase the OP-820 as it refers to duration and time picker

or

you just create multiple PRs with same prefix OP-820.

PR #681, opened 2022-06-14
OP-909 add visit duration time to ward table

dbmalkovsky:
See OP-909

Once this is merged I will push the GUI changes that will allow modification of the duration value.

I defaulted the current value to 30 minutes which is what is used in the new visit dialog currently.


mwithi (approved):
LGMT. Thanks!

PR #684, opened 2022-06-15
OP-820 add hospital level default visit duration to database

dbmalkovsky:
Continuing OP-820...


mwithi (approved):
LGTM. Will this PR be paired with another one on GUI (for the DURATION field only)?

PR #688, opened 2022-06-17
OP-138 - Save patient photo to file system

AndreiDodu:
Before merge I need to test integration with GUI.


src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -47,6 +52,13 @@ public class AdmissionBrowserManager {
 
 	@Autowired
 	private AdmissionIoOperations ioOperations;
+	
+	@Autowired
+	private FileSystemPatientPhotoManager fileSystemPatientPhotoManager;
+	
+	private static final Logger LOGGER = LoggerFactory.getLogger(AdmissionBrowserManager.class);
+	
+	private static final String PATIENT_PHOTO_FROM_DATABASE = "DB";
```
dbmalkovsky, line 61: why not make this PUBLIC in a class like PatientBrowserManager and then reference it in other classes instead of replicating it and making it private?

src/main/java/org/isf/patient/service/PatientIoOperations.java
```diff
@@ -133,11 +135,13 @@ public Patient getPatient(String name) throws OHServiceException {
 	 * @return the Patient
 	 * @throws OHServiceException
 	 */
-	public Patient getPatient(Integer code) throws OHServiceException {
```
dbmalkovsky, line 136: would it make sense to leave this method signature and make it look like:
<pre>
public Patient getPatient(Integer code) throws OHServiceException {
     return this(code, true);
}
</pre>

It seems like loading from the DB is the default and this would make it unnecessary to modify all those calls with a `true` parameter.

PR #695, opened 2022-07-07
OP-705 Add possibility to login by typing the username instead of selecting from a list

nlazzer1:
added USERSLISTLOGIN variable to GeneralData file to login by typing the username in a textbox (no) or selecting the user from a list (yes) (See OP-705)


mwithi (approved):
LGTM. Thanks a lot!

PR #706, opened 2022-07-18
Chore: combine all recent dependabot updates

dbmalkovsky:
Also add property for jaspperreports and make the openfeign property match the format of all the other properties (replaced `-` with `.`).


mwithi (approved):
Very good! Thanks @dbmalkovsky !

PR #708, opened 2022-07-20
OP-929 fix IllegalArgumentException

dbmalkovsky:
See OP-929

While I was in the file making changes I went ahead and made other improvements found by SonarLint and IntelliJ analyze.

Also was able to remove a file that after examination only had one method used by any other class and it was the `JasperReportsManager` class so I made it a private method in the class and removed `DataConverters.java`


src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -81,6 +81,13 @@ public class JasperReportsManager {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JasperReportsManager.class);
 
+	private static final String YYYY_MM_DD = "yyyy-MM-dd";
+	private static final String DD_MM_YYYY = "dd/MM/yyyy";
+	private static final String YYYY_M_MDD = "yyyyMMdd";
+	private static final String E_D_MMMM_YYYY = "E d, MMMM yyyy";
+	private static final String COMMON_ERROR_TITLE = "angal.common.error.title";
+	private static final String STAT_REPORTERROR_MSG = "angal.stat.reporterror.msg";
```
mwithi, line 89: Indentation problem?

mwithi (approved):
LGTM. Thanks!

PR #711, opened 2022-07-22
OP-946 Fix IllegalArgumentException in Pharmaceutical Stock

dbmalkovsky:
See OP-946

Corrects a cut-n-paste error as the fields tested are not the ones used in the query.


mwithi (approved):
LGTM. Thanks!

PR #717, opened 2022-07-29
OP-947 do not export the button object to Excel file

dbmalkovsky:
See OP-947

While I was in the file I removed deprecated and unused methods.


dbmalkovsky:
Need to reset this PR; sorry

PR #720, opened 2022-07-30
 OP-939 add additional validation logic for visit

dbmalkovsky:
See OP-939

Added validation for ward with no beds and for mismatch between patient's sex and ward type.

There is a corresponding GUI PR that implements the visual changes.

### DIFF
--- a/src/main/java/org/isf/visits/manager/VisitManager.java
+++ b/src/main/java/org/isf/visits/manager/VisitManager.java
@@ -41,6 +41,7 @@ import org.isf.utils.exception.model.OHSeverityLevel;
 import org.isf.utils.time.TimeTools;
 import org.isf.visits.model.Visit;
 import org.isf.visits.service.VisitsIoOperations;
+import org.isf.ward.model.Ward;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/visits/manager/VisitManager.java
```diff
@@ -88,10 +89,24 @@ public void validateVisit(Visit visit) throws OHServiceException {
 //							OHSeverityLevel.ERROR));
 //
 //		}
+		if (visit.getWard().getBeds() <= 0) {
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
+					MessageBundle.getMessage("angal.visit.thewardhasnobedsavailable.msgg"),
+					OHSeverityLevel.ERROR));
+		}
```
mwithi, line 96: this validation is unnecesary: the visit is not related to the beds occupation, anyway the .getBeds() method only returns the number of beds in the wards (not its state) so it cannot be negative

dbmalkovsky: The test is for zero or less beds; not just negative.

So how do I test "to show also wards with no beds" or do we care if there are no beds in the ward?   The Jira issue seems to indicate that there is some test.

mwithi: No, in this case we don't care if there are available free beds (that if they are zero cannot be available in any case :) )

The validation is only about sex. The wards to be shown (in gui) are all of them, full list ("to remove list filtering")

I've also updated the issue description because it had a wrong link (OP-704 -> OP-714)

dbmalkovsky: removed the test

mwithi: Thanks you @dbmalkovsky !

mwithi (approved):
LGMT. Thanks!

PR #723, opened 2022-07-31
OP-947 do not export the button object to Excel file (old format)

dbmalkovsky:
See OP-947

While working on OP-943 I realized that the OLD format needed the same column count parameter.  Originally I had deleted all the OLD methods until I noticed OP-943.   I forgot to go back and make the column change.


mwithi (approved):
LGTM. Thanks!

PR #726, opened 2022-08-06
OP-948 fix count of medical stock when moving to another ward

dbmalkovsky:
See OP-948


mwithi (approved):
LGTM and tested. Thanks!

PR #727, opened 2022-08-06
OP-948 fix count of medical stock when moving to another ward

dbmalkovsky:
This is the hotfix version

See OP-948

### DIFF
--- a/src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
+++ b/src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
@@ -231,28 +231,26 @@ public class MedicalStockWardIoOperations
                 }
 		Integer medical = movement.getMedical().getCode();		
 		boolean result = true;
-		
-		
-        if(wardTo != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -246,6 +246,7 @@ protected boolean updateStockWardQuantity(
 				medicalWard.setLot(movement.getLot());
 				repository.save(medicalWard);
             }
+	        repository.updateOutQuantity(Math.abs(qty), ward, medical, lot);
```
mwithi, line 249: maybe indentation?

dbmalkovsky: Fixed the indentation and while I was at it corrected some formatting in that area too.

mwithi: saw it, thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #728, opened 2022-08-08
OP-902 add STRONGLENGTH to GeneralData

dbmalkovsky:
See OP-902

Also added code so the value cannot be less than 6 characters (which was the value in the original reset password function).


mwithi (approved):
ALGTM and tested! Thanks a lot!

PR #734, opened 2022-08-26
Chore: combine all recent dependabot jar updates

dbmalkovsky:
Will rebase dependabot PRs when this is committed.

Tested with `mvn clean install`


mwithi (approved):
Hope major version change on slf4j will be safe! LTGM anyway!

PR #557, opened 2021-12-30
OP-727 prefix table names

dbmalkovsky:
See OP-727

All the CORE junit tests pass locally.

This is just the start of the changes; I know that all the report generation files need to be updated but I'm not familiar with the tool.

PR #745, opened 2022-09-07
OP-727 remove NON_KEYWORDS from db url once all tables have prefix

dbmalkovsky:
This can be merged after https://github.com/informatici/openhospital-core/pull/557 is merged into develop; the CI checks will fail until this is done.


dbmalkovsky:
Incorporated into PR https://github.com/informatici/openhospital-core/pull/557

PR #753, opened 2022-09-19
OP-967 DICOM dates local independent

dbmalkovsky:
See OP-967

Series date and Study date are back to including zone information and are formatted like the default for `Date` as was done in 1.11.4.  The date is written out in English.  In addition as `LocalDateTime` objects in java do not by default include zone information it is generally ignored but parsed and saved in the properties file.

The core junit tests run and interactive testing was successful.


mwithi (approved):
LGTM. Thanks a lot!

PR #766, opened 2022-09-24
OP-968 truncate DateTime values to seconds when storing in model and in service level

dbmalkovsky:
See OP-968

Some might consider these changes excessive and unneeded but given the many sources of where the values are coming from these changes ensure everything is consistent.

Added a new method `TimeTools.truncateToSeconds()` that does what the method name implies but also checks for null values.


mwithi (approved):
LGTM. Thanks!

PR #768, opened 2022-09-25
OP-727: add prefix to tables so create_all_XX.sql script works

dbmalkovsky:
Add "OH_" prefix to `vaccine` and `vaccinetype` tables so the `create_all_XX.sql` script doesn't fail


mwithi (commented):
LGMT. Thanks!

PR #770, opened 2022-09-27
OP-781 drop LAB_EXAM_DATE field

dbmalkovsky:
See OP-781

All CORE unit tests passed locally.

There is a corresponding GUI PR that fixes the JasperReports given the `LAB_EXAM_DATE` field is removed.


src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -79,8 +79,8 @@ protected void setPatientConsistency(Laboratory laboratory) {
 	 */
 	protected void validateLaboratory(Laboratory laboratory) throws OHDataValidationException {
 		List<OHExceptionMessage> errors = new ArrayList<>();
-		if (laboratory.getExamDate() == null) {
-			laboratory.setExamDate(LocalDate.now());
+		if (laboratory.getCreatedDate() == null) {
+			laboratory.setCreatedDate(TimeTools.getNow());
```
mwithi, line 83: This should be managed by JPA automatically

src/main/java/org/isf/lab/model/Laboratory.java
```diff
@@ -270,7 +249,7 @@ public int hashCode() {
 	@Override
 	public String toString() {
 		return "-------------------------------------------\nLaboratory{" + "code=" + code + ", material=" + material
-				+ ", exam=" + exam + ", registrationDate=" + labDate + ", examDate=" + examDate + ", result="
+				+ ", exam=" + exam + ", registrationDate=" + labDate + ", examDate=" + createdDate + ", result="
```
mwithi, line 252: this should be `labDate`, simply

src/test/java/org/isf/lab/test/TestLaboratory.java
```diff
@@ -39,7 +38,7 @@ public class TestLaboratory {
 
 	private String material = "TestMaterial";
 	private LocalDateTime labDate = TimeTools.getNow();
-	private LocalDate examDate = labDate.toLocalDate();
+	private LocalDateTime examDate = labDate;
```
mwithi, line 41: is this `examDate` still used?

src/test/java/org/isf/lab/test/TestLaboratory.java
```diff
@@ -94,7 +91,9 @@ public void check(Laboratory laboratory) {
 			assertThat(laboratory.getPatName()).isEqualTo(patName);
 		}
 		assertThat(laboratory.getDate()).isCloseTo(labDate, within(1, ChronoUnit.SECONDS));
-		assertThat(laboratory.getExamDate()).isEqualTo(examDate);
+		// because CreatedDate is set by the auditable class when the datbase is updated the value will differ from
+		// the static examDate value set above.
+		assertThat(laboratory.getCreatedDate()).isCloseTo(examDate, within(25, ChronoUnit.SECONDS));
```
mwithi, line 96: I think auditable fields can be ignored in tests

src/main/java/org/isf/lab/manager/LabManager.java
mwithi: where `setCreatedDate` is defined?

dbmalkovsky: Use you are correct it can be removed.

`setCreateDate` is found in `org.isf.utils.db.Auditable`

src/main/java/org/isf/lab/model/Laboratory.java
dbmalkovsky: Fixed.

src/test/java/org/isf/lab/test/TestLaboratory.java
dbmalkovsky: Agreed; good catch.

src/main/java/org/isf/lab/model/Laboratory.java
mwithi: Thanks!

src/test/java/org/isf/lab/test/TestLaboratory.java
mwithi: Thanks!

sql/step_78_adjust_lab_datetime.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE OH_LABORATORY SET LAB_CREATED_DATE = LAB_DATE;
+UPDATE OH_LABORATORY SET LAB_DATE = CAST(CONCAT(LAB_EXAM_DATE, ' ', TIME(LAB_DATE)) AS DATETIME);
+
+ALTER TABLE OH_LABORATORY
+DROP COLUMN LAB_EXAM_DATE;
```
mwithi, line 5: I would put it in a singole line
and I would rename the file in `step_78_drop_lab_exam_date.sql` (like step_71)

mwithi:
IDK if it is this the expected result, but now labDate is second-less
image
which will always differ from created_date
image

This might be true for all new `GoodTime*Chooser` though...

sql/step_78_adjust_lab_datetime.sql
dbmalkovsky: done; BTW, I just used the name that was there in the extra directory and changed XX to 78

mwithi: ok, thanks!

dbmalkovsky:
We are storing the seconds:  20:53:49 is 10PM, 53 minutes and 49 seconds.

mwithi:
> We are storing the seconds: 20:53:49 is 8 PM, 53 minutes and 49 seconds.

yes, but not in LAB_DATE

dbmalkovsky:
I just created a new lab and it looks like there are seconds:
image

mwithi:
> I just created a new lab and it looks like there are seconds: ![image](https://user-

let me try again...

mwithi:
> > I just created a new lab and it looks like there are seconds: image -> again JDK problem?
2. subsequent exams will get always the same time -> I think because of RememberDates, but it should keep only the date part

mwithi:
by debuggin it seems the value is taken directly like that 🤔

image

mwithi:
The same happens in OPD (and I suppose everywhere)

image

src/main/java/org/isf/lab/manager/LabManager.java
mwithi: is this removed or not yet?

src/test/java/org/isf/lab/test/Tests.java
```diff
@@ -696,7 +696,7 @@ public void testMgrValidationMissingExamDate() throws Exception {
 
 		ArrayList<String> labRow = new ArrayList<>();
 		Laboratory laboratory = testLaboratory.setup(exam, patient, false);
-		laboratory.setExamDate(null);
+		laboratory.setCreatedDate(null);
```
mwithi, line 699: This also I think can be removed

src/main/java/org/isf/lab/model/Laboratory.java
mwithi: sorry, spotted now... `registrationDate=" + createdDate + ", examDate=" + labDate`

PR #780, opened 2022-10-05
OP-977 changes to persist usergroup changes

dbmalkovsky:
See OP-977

Changes required to persist changes to both user description and group


mwithi (approved):
LGTM. Thanks a lot!

PR #781, opened 2022-10-05
OP-707_Demo-data

mwithi:
Demo data for OP-707.


dbmalkovsky (approved):
The diffs were too many so I ran the script and then did some interactive testing; LGTM

mwithi:
> The diffs were too many so I ran the script and then did some interactive testing; LGTM

I've reduced them to the minimum... curiously `oh_patient_profile_photo` table was still missing...

Thanks, I'll merge like that

PR #785, opened 2022-10-08
OP-985 remove results check for lab type 2; no labrows for negative results

dbmalkovsky:
See OP-985


mwithi (commented):
LGTM. Thanks a lot!

PR #808, opened 2022-11-11
OP-987 - OP-138 causes getPatientId to update at each read

goto-eof:
See OP-987.

### DIFF
--- a/src/main/java/org/isf/patient/service/PatientIoOperations.java
+++ b/src/main/java/org/isf/patient/service/PatientIoOperations.java
@@ -62,6 +62,8 @@ public class PatientIoOperations {
 
 	public static final String NOT_DELETED_STATUS = "N";
 	
+	public static boolean IS_LOAD_PROFILE_PHOTO_FROM_DB = LOAD_FROM_DB.equals(GeneralData.PATIENTPHOTOSTORAGE);
+	
 	@Autowired
 	private PatientIoOperationRepository repository;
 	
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/patient/service/PatientIoOperations.java
```diff
@@ -128,6 +136,7 @@ public Patient getPatient(Integer code) throws OHServiceException {
 			if (isLoadProfilePhotoFromDb) {
```
dbmalkovsky, line 136: Why is this loaded at the beginning of each file:
<pre>
boolean isLoadProfilePhotoFromDb = LOAD_FROM_DB.equals(GeneralData.PATIENTPHOTOSTORAGE);
</pre>
Can't it be a class static?   It doesn't change unless the system is restarted, correct?

src/main/java/org/isf/patient/service/PatientIoOperations.java
```diff
@@ -121,38 +131,37 @@ public List<Patient> getPatientsByOneOfFieldsLike(String keyword) throws OHServi
...
+			fileSystemPatientPhotoRepository.loadInPatient(patient, GeneralData.PATIENTPHOTOSTORAGE);
+		}
+	}
+
 	/**
 	 * Method that gets a Patient by his/her name
 	 * 
 	 * @param name
-	 * @param isLoadProfilePhotoFromDb
+	 * @param IS_LOAD_PROFILE_PHOTO_FROM_DB
```
mwithi, line 156: this is not a paran of the method `getPatient(String name)`

mwithi (approved):
LGTM. Thanks a lot!

PR #830, opened 2022-11-29
OP-980 - image file loader improvements

goto-eof:
See OP-980.

Paired with https://github.com/informatici/openhospital-gui/pull/1492.


src/main/java/org/isf/utils/gui/PatientGuiConst.java
```diff
@@ -0,0 +1,29 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.isf.utils.gui;
+
+public interface PatientGuiConst {
```
dbmalkovsky, line 24: I'm curious as to why these constants are in the CORE repo when they only appear to be used in GUI related code.....

goto-eof: @mwithi asked me to move them in core.

mwithi: well, it is part of the issue, but I didn't request to do it in this way...
only the constant itself is supposed to be moved, not the whole class

moreover the issue is talking about two constants: `IMAGE_THUMBNAIL_MAX_WIDTH` and `MAXPROFPICFILESIZEBYTES`
I would move both into GeneralData (only in the code) so in the future we can have them controlled by the user... wdyt?

src/main/java/org/isf/generaldata/GeneralData.java
```diff
@@ -144,6 +144,9 @@ public final class GeneralData extends ConfigurationProperties {
 	private static final boolean DEFAULT_STRONGPASSWORD = true;
 	private static final String DEFAULT_PATIENTPHOTOSTORAGE = "DB";
 	
+	public static final Integer IMAGE_THUMBNAIL_MAX_WIDTH = 140;
+	public static final int MAXPROFPICFILESIZEBYTES = 32768;
```
mwithi, line 148:
```suggestion
	public static final int MAX_PROFILE_IMAGE_FILE_SIZE_BYTES = 32768;
```

src/main/java/org/isf/generaldata/GeneralData.java
```diff
@@ -144,6 +144,9 @@ public final class GeneralData extends ConfigurationProperties {
 	private static final boolean DEFAULT_STRONGPASSWORD = true;
 	private static final String DEFAULT_PATIENTPHOTOSTORAGE = "DB";
 	
+	public static final Integer IMAGE_THUMBNAIL_MAX_WIDTH = 140;
```
mwithi, line 147: Is there a particular reason for `Integer` instead of `int`?

mwithi (changes requested):
find some code suggestions

mwithi (approved):
LGTM. Thanks a lot!

PR #835, opened 2022-12-08
Separate default OH database creation

mizzioisf:
Separate default OH database creation from the SQL script in order to have the database name parametric.
This allows to have multiple version databases, separated demo database,  etc. and might be useful for database management operations.
Also we separate the hardcoded "drop oh database" that could be risky in some situation.

Will update documentation accordingly (Admin Manual).


sql/step_01_create_structure.sql
```diff
@@ -14,13 +14,6 @@
...
 
-
-drop database if exists oh;
-create database oh CHARACTER SET utf8;
-
-use oh;
-
-
 --
 -- creazione TABELLE
```
dbmalkovsky, line 18: comments in English? 😏

mizzioisf: You are right @dbmalkovsky !
BTW sorry but I missed your comments, for some reason I didn't get any notification.

mwithi (approved):
LGTM. Thanks a lot!

PR #836, opened 2022-12-08
OP-97 - New Anamnesis module

goto-eof:
See OP-97.


src/main/java/org/isf/anamnesis/model/PatientPatientHistory.java
```diff
@@ -0,0 +1,56 @@
+/**
```
dbmalkovsky, line 1: Missing copyright block

src/main/java/org/isf/anamnesis/model/PatientHistory.java
```diff
@@ -0,0 +1,970 @@
...
+	private boolean familyOrto;
+
+	@Column(name="PAH_FAM_GYNO")
+	private boolean familyGyno;
+
+	@Column(name="PAH_FAM_OTHER")
+	private boolean familyOther;
+
+	@Column(name="PAH_FAM_NOTE")
+    private String familyNote;
```
dbmalkovsky, line 97: indentation?

src/main/java/org/isf/anamnesis/model/PatientHistory.java
```diff
@@ -0,0 +1,970 @@
...
+	private boolean patClosedOrto;
+
+	@Column(name="PAH_PAT_CLO_GYNO")
+	private boolean patClosedGyno;
+
+	@Column(name="PAH_PAT_CLO_OTHER")
+	private boolean patClosedOther;
+
+	@Column(name="PAH_PAT_CLO_NOTE")
+    private String patClosedNote;
```
dbmalkovsky, line 133: indentation?

src/main/java/org/isf/anamnesis/model/PatientHistory.java
```diff
@@ -0,0 +1,970 @@
...
+	private boolean patOpenOrto;
+
+	@Column(name="PAH_PAT_OPN_GYNO")
+	private boolean patOpenGyno;
+
+	@Column(name="PAH_PAT_OPN_OTHER")
+	private boolean patOpenOther;
+
+	@Column(name="PAH_PAT_OPN_NOTE")
+    private String patOpenNote;
```
dbmalkovsky, line 169: indentation here and following lines; is this tabs vs spaces ?

src/main/java/org/isf/anamnesis/model/PatientHistory.java
```diff
@@ -218,7 +218,7 @@ public class PatientHistory extends Auditable<String> implements Comparable<Pati
 	private int phyMenopauseYears;
 
 	@Column(name = "PAH_PHY_HRT_NOR")
-	private boolean phyHrtNormal;
+	private boolean phyHrtNormal = true;
```
mwithi, line 221: it's ok... should we follow same "default 1" values for all fields like in the step_82 script? (I didn't check if we do)
Also, is there a way to do it by annotation instead of by variable assignation? (just wondering what is the best practise)

goto-eof: I confirm that it is not following the default values like in the step_82 script. Should I follow the values of UI or those of the script_82? There is a way to do it with annotations, like @PrePersist, but it is cumbersome, there are to many fields and we should use wrappers not primitive types. @ColumnDefault instead is usefull when you generate the DB with hibernate, but we don't do that.

mwithi: Ok, let's stay this way, thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #851, opened 2023-01-18
OP-999 make jasper reports loadable from folder

nlazzer1:
See OP-999.


src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -977,22 +993,58 @@ private JasperReportResultDto generateJasperReport(String jasperFilename, String
...
+	}
+
+	private String compileFilename(String folderName, String jasperFileName, List<String> params, String ext) {
+		StringBuilder sbFilename = new StringBuilder();
+		sbFilename.append(folderName);
+		sbFilename.append(File.separator);
+		sbFilename.append(jasperFileName);
+		if(params != null) 
+		{
+			params.forEach(p -> {
```
dbmalkovsky, line 1012: I had to hide white space differences to better see the changes in the file.   It looks like there was a fair amount of reformatting done.    In addition like things like `if(` should be `if (` to match project standards.   And braces are not on their own line, so line 1010 should be `if (parms != null) {`

nlazzer1: Thanks David, I reformatted the code according to the OH style

pom.xml
```diff
@@ -131,7 +131,7 @@
 								<exclude>lib/**</exclude>
 								<exclude>sql/**</exclude>
 								<exclude>repo/**</exclude>
-								<exclude>rpt/**</exclude>
+								<exclude>rpt_base/**</exclude>
```
mwithi, line 134: should we add here also the other new folders?

mwithi (approved):
LGTM. Thanks a lot!

PR #857, opened 2023-01-24
OP-981 dicom series for jpegs

mwithi:
See OP-981.

Also fixed folder loading interruption when one or more file are not supported, instead it logs the error and continues the loading.


mwithi:
Discovered after PR checks fail, going to fix...

dbmalkovsky (approved):
LGTM

PR #863, opened 2023-01-27
OP-213 add search field for opd records

mwithi:
See OP-213.

Add also USER column and related filter (like in Wolisso's project).


dbmalkovsky (approved):
LGTM.    While not technically "wrong" I don't like using `Integer` objects for parameters when primitive `int` will work just as well.

mwithi:
> LGTM. While not technically "wrong" I don't like using `Integer` objects for parameters when primitive `int` will work just as well.

You're right, I just used copy and paste some code, then automatic method creation of Eclipse did the rest. I did the change also in "gui".

Thanks!

PR #865, opened 2023-01-27
OP-527 inconsistent column definition for XXX_DELETED fields

mwithi:
See OP-527.

### DIFF
--- a/src/main/java/org/isf/admission/model/Admission.java
+++ b/src/main/java/org/isf/admission/model/Admission.java
@@ -198,7 +198,7 @@ public class Admission extends Auditable<String> implements Comparable<Admission
 
 	@NotNull
 	@Column(name = "ADM_DELETED", columnDefinition = "char(1) default 'N'")
-	private Character deleted = 'N';                // flag record deleted ; values are 'Y' OR 'N' default is 'N'
+	private char deleted = 'N';                // flag record deleted ; values are 'Y' OR 'N' default is 'N'
 
 	@Transient
 	private volatile int hashCode = 0;
--- a/src/main/java/org/isf/admission/model/Admission.java
+++ b/src/main/java/org/isf/admission/model/Admission.java
@@ -197,7 +197,7 @@ public class Admission extends Auditable<String> implements Comparable<Admission
 
 	@NotNull
 	@Column(name = "ADM_DELETED")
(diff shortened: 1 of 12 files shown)


src/main/java/org/isf/admission/model/Admission.java
```diff
@@ -40,6 +40,7 @@
 import javax.persistence.Version;
 import javax.validation.constraints.NotNull;
 
+import org.hibernate.annotations.ColumnDefault;
```
dbmalkovsky, line 43: This import is not used.

src/main/java/org/isf/admission/model/Admission.java
```diff
@@ -196,8 +197,8 @@ public class Admission extends Auditable<String> implements Comparable<Admission
 	private int lock;                        // default 0
 
 	@NotNull
-	@Column(name = "ADM_DELETED")
-	private String deleted;                    // flag record deleted ; values are 'Y' OR 'N' default is 'N'
+	@Column(name = "ADM_DELETED", columnDefinition = "char(1) default 'N'")
+	private Character deleted = 'N';                // flag record deleted ; values are 'Y' OR 'N' default is 'N'
```
dbmalkovsky, line 201:
```suggestion
	private char deleted = 'N';                // flag record deleted ; values are 'Y' OR 'N' default is 'N'
```

This way there is no boxing/unboxing of the field.    All unit tests pass with this change.

src/main/java/org/isf/patient/model/Patient.java
```diff
@@ -174,8 +174,8 @@ public class Patient extends Auditable<String> {
 	private String profession;
 
 	@NotNull
-	@Column(name="PAT_DELETED")
-	private String deleted = "N";
+	@Column(name="PAT_DELETED", columnDefinition = "char(1) default 'N'")
+	private Character deleted = 'N';
```
dbmalkovsky, line 178:
```suggestion
	private char deleted = 'N';
```

Same change as noted above.

dbmalkovsky (changes requested):
Please see suggested changes.   Otherwise LGTM.

src/main/java/org/isf/admission/model/Admission.java
mwithi: Removed, thanks!

PR #886, opened 2023-02-14
OH2-172 add apis to enhance opd manipulation with operations rows as it is done for oh 1.11 (Updated)

dbmalkovsky:
This is based on the older OP2-172 PR but it adds the corrections for the unit tests.


dbmalkovsky:
Add PR on intesys repo in place of this

PR #889, opened 2023-02-18
OP-897 limit password retries

dbmalkovsky:
See OP-897


sql/step_04_all_following_steps.sql
```diff
@@ -83,4 +83,4 @@ source step_86_permissions_and_grouppermission.sql;
 source step_87_alter_table_opd.sql;
 source step_88_alter_table_patient.sql;
 source step_89_alter_tables_innodb.sql;
-
+source step_90_login_retries.sql;
```
mwithi, line 86: this step should be included also in Dockerfile-ohdb

src/main/java/org/isf/generaldata/GeneralData.java
```diff
@@ -94,6 +94,8 @@ public final class GeneralData extends ConfigurationProperties {
 
 	public static boolean STRONGPASSWORD;
 	public static int STRONGLENGTH;
+	public static int PASSWORDTRIES;
+	public static int PASSWORDLOCKTIME;
```
mwithi, line 98: These new parameters should be added also in settings.properties.dist in "gui"

sql/step_04_all_following_steps.sql
dbmalkovsky: Done

src/main/java/org/isf/generaldata/GeneralData.java
dbmalkovsky: Done; thanks

dbmalkovsky:
Good catch; the code was updating the record in the database with the number of failures but checking again the local object that wasn't refreshed from the value in the db.   Fixed.

mwithi (approved):

PR #891, opened 2023-02-20
OH2-185 update patient return 500 error

ArnaudFonzam:
See OH2-185.


src/main/java/org/isf/patient/service/PatientIoOperations.java
```diff
@@ -132,7 +132,9 @@ public Patient getPatient(Integer code) throws OHServiceException {
 		List<Patient> patients = repository.findAllWhereIdAndDeleted(code, NOT_DELETED_STATUS);
 		if (!patients.isEmpty()) {
 			Patient patient = patients.get(patients.size() - 1);
-			retrievePatientProfilePhoto(patient);
+			if (patient.getPatientProfilePhoto() != null) {
+				retrievePatientProfilePhoto(patient);
+			}
```
mwithi, line 137: It would never enter the if block because the `retrievePatientProfilePhoto(patient);` is needed exactly to retrieve `patient.getPatientProfilePhoto()`...

ArnaudFonzam: yes you have reason

mwithi: Ok, I close this PR

PR #892, opened 2023-02-22
OH2-187 search opd returns no content in all cases of the filter

ArnaudFonzam:
See OH2-187.


src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
```diff
@@ -82,22 +82,23 @@ private TypedQuery<Opd> getOpdQuery(
...
 			);
 		}
-		if (ageFrom != 0 || ageTo != 0) {
+		if (ageFrom > 0 && ageTo < 200) {
 			predicates.add(
 					cb.between(opd.<Integer>get("age"), ageFrom, ageTo)
 			);
 		}
 		if (sex != 'A') {
+			System.out.println("sexe");
```
mwithi, line 101: No println are allowed in the code

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
```diff
@@ -82,22 +82,23 @@ private TypedQuery<Opd> getOpdQuery(
...
 			);
 		}
-		if (diseaseCode != null) {
+		if (diseaseCode != null && !diseaseCode.equals("")) {
 			predicates.add(
 					cb.equal(opd.join("disease").get("code"), diseaseCode)
 			);
 		}
-		if (ageFrom != 0 || ageTo != 0) {
+		if (ageFrom > 0 && ageTo < 200) {
```
mwithi, line 95: AND operation would cause the input `0 - 199` not entering the if block.

If you don't like `!= 0` and you don't want negative values you may propose (with OR) `if (ageFrom > 0 || ageTo > 0)`
otherwise `ageTo < 200` can still be negative.

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
```diff
@@ -82,22 +82,23 @@ private TypedQuery<Opd> getOpdQuery(
 					cb.equal(opd.join("ward").get("code"), ward.getCode())
 			);
 		}
-		if (diseaseTypeCode != null) {
+		if (diseaseTypeCode != null && !diseaseTypeCode.equals("")) {
```
mwithi, line 85: Why to remove the empty string comparation? it is like passing a null value

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
ArnaudFonzam: ageFrom should be greater or equals than 0 and ageFrom should be less than ageTo , ageTo should be less than 200. that is the requiement

mwithi: we are talking about a filter, not the insertion of a new record....

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
mwithi: sorry, my mistake, it's ok

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
```diff
@@ -82,17 +82,17 @@ private TypedQuery<Opd> getOpdQuery(
...
 			);
 		}
-		if (diseaseCode != null) {
+		if (diseaseCode != null && !diseaseCode.equals("")) {
 			predicates.add(
 					cb.equal(opd.join("disease").get("code"), diseaseCode)
 			);
 		}
-		if (ageFrom != 0 || ageTo != 0) {
+		if (ageFrom != 0 && ageTo != 0 ) {
```
mwithi, line 95: && (AND) will cause the interval 0-x (from 0 to x) would not enter in the if block.
Please repleace it with || (OR)

src/main/java/org/isf/opd/service/OpdIoOperationRepositoryImpl.java
```diff
@@ -82,17 +82,17 @@ private TypedQuery<Opd> getOpdQuery(
...
 			);
 		}
-		if (diseaseCode != null) {
+		if (diseaseCode != null && !diseaseCode.equals("")) {
 			predicates.add(
 					cb.equal(opd.join("disease").get("code"), diseaseCode)
 			);
 		}
-		if (ageFrom != 0 || ageTo != 0) {
+		if (ageFrom != 0 || ageTo != 0 ) {
```
mwithi, line 95:
```suggestion
		if (ageFrom != 0 || ageTo != 0) {
```

PR #900, opened 2023-03-03
OH2-183 Create api of Exam Request

ArnaudFonzam:
See OH2-183.


src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +265,18 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	 * Inserts one Laboratory exam {@link Laboratory} (All Procedures)
+	 *
+	 * @param laboratory - the laboratory without result
+	 * @return <code>true</code> if the exam has been inserted, <code>false</code> otherwise
+	 * @throws OHServiceException
+	 */
+	public boolean newLaboratory3(Laboratory laboratory) throws OHServiceException {
+		setPatientConsistency(laboratory);
+		return ioOperations.newLabFirstProcedure(laboratory);
+	}
```
mwithi, line 278: why to add this method instead to modify and use the existing one `newLaboratory`?
This method is for FirstProcedure, so also the comment is wrong (All Procedures) and therefore is not clear what are you trying to do

mwithi: ok, as per discussion under OH2-183, the method should be named "newExamRequest", but the suggested status LABORATORY_ACTIVE=2 is not correct, LAB_ACTIVE (this the real field name) is for future use.
Please consider to better structure the model.

ArnaudFonzam: @mwithi  have create the new method because all other method need exam result to be set. For this requiement we just need to set exam and the information of patient set. for the field LAB_ACTIVE, we have 3 case: LAB_ACTIVE=0  mince exam is delete LAB_ACTIVE = 1 mince exam is done and LAB_ACTIVE = 2 mince exam is requested. that is the logic

mwithi: the logic using LAB_ACTIVE as status field is wrong. Please read my comments under both OH2-183 and OHCS-84. Thanks

sql/step_91_oh_update_table_laboratory.sql
```diff
@@ -0,0 +1,2 @@
+UNLOCK TABLES;
+ALTER TABLE LABORATORY ADD COLUMN LAB_STATUS VARCHAR(7) NULL DEFAULT NULL;
```
dbmalkovsky, line 2: Shouldn't there be an update to the demo data to go along with the adding of this column?

sql/step_91_oh_update_table_laboratory.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE LABORATORY ADD COLUMN LAB_STATUS VARCHAR(7) NOT NULL DEFAULT 'DONE';
```
mwithi, line 1: all statuses should be lowercase

sql/load_demo_data.sql
```diff
@@ -1952,6 +1952,7 @@ CREATE TABLE `oh_laboratory` (
   `LAB_SEX` char(1) DEFAULT NULL,
   `LAB_MATERIAL` varchar(25) DEFAULT NULL,
   `LAB_PAT_INOUT` char(1) DEFAULT NULL,
+  `LAB_STATUS` varchar(7) DEFAULT 'DONE',
```
mwithi, line 1955: statuses should be lowercase

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +265,28 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	/**
+	 * update one Laboratory exam {@link Laboratory} (All Procedures)
+	 *
+	 * @param laboratory - the laboratory without result
+	 * @return <code>true</code> if the exam has been update, <code>false</code> otherwise
+	 * @throws OHServiceException
+	 */
+	public boolean updateExamRequest(Laboratory laboratory) throws OHServiceException {
+		return ioOperations.updateLabFirstProcedure(laboratory);
+	}
```
mwithi, line 289: in terms of API, wouldn't be better to have a "new" method that forces the state to 'draft' and another method that just update the state?

If I want to update the laboratory (with results) there are already other methods.

Also the javadoc should be changed accordingly

ArnaudFonzam: yes if you want to update the laboratory with results we already have the method for that case.  about new method that force the state to 'draft' and another method that just update the state now if you want to update the state with another fields you will use other method while you can just use the updateExamRequest for all case.

mwithi: What I mean is that `newExamRequest` should call `laboratory.setStatus("draft")` before the ioOperations and the result should be blank even if passed, otherwise one could use the same method to save a laboratory first procedure with result directly in "done" and so duplicating the `newLabFirstProcedure` method.

Similarly, the method `updateExamRequest` should only change the laboratory status and the result should be unchanged, otherwise it may be used to update a laboratory first procedure and its result, duplicating the `updateLabFirstProcedure` method: in this case we could accept `code` and `status` as the only parameters.

In other words, a well-defined method should perform only the expected action, and the javadoc also should be more precise in this.

Also, a laboratoryStatus `enum` should be defined and used here.

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +266,40 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
 		}
 	}
 
+	/**
+	 * Inserts one Laboratory exam {@link Laboratory} (All Procedures)
```
mwithi, line 270:
```suggestion
	 * Inserts one Laboratory request {@link Laboratory} (All Procedures)
```

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +266,40 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	 */
+	public boolean newExamRequest(Laboratory laboratory) throws OHServiceException {
+		setPatientConsistency(laboratory);
+		laboratory.setStatus(LaboratoryStatus.draft.toString());
+		laboratory.setResult("");
+		return ioOperations.newLabFirstProcedure(laboratory);
+	}
+	
+	/**
+	 * update one Laboratory exam {@link Laboratory} (All Procedures)
```
mwithi, line 284:
```suggestion
	 * update one Laboratory request status (All Procedures)
```

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +266,40 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
 		}
 	}
 
+	/**
+	 * Inserts one Laboratory exam {@link Laboratory} (All Procedures)
+	 *
+	 * @param laboratory - the laboratory without result
+	 * @return <code>true</code> if the exam has been inserted, <code>false</code> otherwise
```
mwithi, line 273:
```suggestion
	 * @return <code>true</code> if the request has been inserted, <code>false</code> otherwise
```

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +266,40 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
 		}
 	}
 
+	/**
+	 * Inserts one Laboratory request {@link Laboratory} (All Procedures)
```
mwithi, line 270:
```suggestion
	 * Inserts one Laboratory request (All Procedures)
```

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -265,6 +266,40 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	 */
+	public boolean newExamRequest(Laboratory laboratory) throws OHServiceException {
+		setPatientConsistency(laboratory);
+		laboratory.setStatus(LaboratoryStatus.draft.toString());
+		laboratory.setResult("");
+		return ioOperations.newLabFirstProcedure(laboratory);
+	}
+	
+	/**
+	 * update one Laboratory request {@link Laboratory} (All Procedures)
```
mwithi, line 284:
```suggestion
	 * update one Laboratory request (All Procedures)
```

src/main/java/org/isf/lab/model/Laboratory.java
```diff
@@ -106,10 +106,21 @@ public class Laboratory extends Auditable<String> {
 
 	@Column(name="LAB_SEX")
 	private String sex;
+	
+	@Column(name="LAB_STATUS")
+	private String status;
```
mwithi, line 111: @dbmalkovsky for GUI should we assign a default value when creating a laboratory (directly in "done" I suppose)

src/main/java/org/isf/lab/model/Laboratory.java
```diff
@@ -106,10 +106,21 @@ public class Laboratory extends Auditable<String> {
...
 	private volatile int hashCode = 0;
 
+	public String getStatus() {
+		return status;
+	}
+
+	public void setStatus(String status) {
+		this.status = status;
+	}
+
```
mwithi, line 123: please move getters and setters along with others in the class

src/main/java/org/isf/lab/model/LaboratoryStatus.java
```diff
@@ -0,0 +1,5 @@
+package org.isf.lab.model;
```
dbmalkovsky, line 1: Missing copyright notice

src/main/java/org/isf/lab/model/LaboratoryStatus.java
```diff
@@ -0,0 +1,5 @@
...
+
+public enum LaboratoryStatus {
+	draft, open, done
```
dbmalkovsky, line 4:
```suggestion
	DRAFT, OPEN, DONE
```

sql/step_04_all_following_steps.sql
```diff
@@ -84,3 +84,4 @@ source step_87_alter_table_opd.sql;
 source step_88_alter_table_patient.sql;
 source step_89_alter_tables_innodb.sql;
 source step_90_login_retries.sql;
+source step_91_oh_update_table_laboratory.sql;
```
dbmalkovsky, line 87:
```suggestion
source step_91_alter_laboratory_with_status.sql;
```
There is no need to add "_oh_" to the file name and it is better to include the changes made in the file name.  `update_table_laboratory` is too generic.

src/main/java/org/isf/lab/model/LaboratoryStatus.java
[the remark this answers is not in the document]
mwithi: true... anyway in the table we should persist lowercases, how could be achieved this mapping?

dbmalkovsky (changes requested):
Left various comments with code improvements.

src/main/java/org/isf/lab/model/Laboratory.java
dbmalkovsky: why not just create the default for status when creating the object in the constructors in `Laboratory.java`?

mwithi: I mean, for GUI the field is not (yet) managed, so the default (that should be "draft" in the constructor) for GUI has to be set to "done". So I think a PR on GUI is needed to set this field, wdyt?

src/main/java/org/isf/lab/model/LaboratoryStatus.java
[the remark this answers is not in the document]
dbmalkovsky: Something like this?

<pre>
	public enum LaboratoryStatus {
		DRAFT("draft"), OPEN("open"), DONE("done");

		private String status;

		private LaboratoryStatus(String s) {
			status = s;
		}

		public String toString() {
			return this.status;
		}
	}
</pre>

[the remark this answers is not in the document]
mwithi: yes!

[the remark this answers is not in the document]
mwithi: @ArnaudFonzam please use that pattern from @dbmalkovsky for enum with String values. Thanks!

sql/step_04_all_following_steps.sql
mwithi: suggested: `step_91_add_laboratory_status.sql`

src/main/java/org/isf/lab/model/LaboratoryStatus.java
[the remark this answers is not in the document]
ArnaudFonzam: okay

ArnaudFonzam: okay @dbmalkovsky

src/main/java/org/isf/lab/model/Laboratory.java
ArnaudFonzam: okay

src/main/java/org/isf/lab/model/Laboratory.java
mwithi: Example here https://github.com/informatici/openhospital-gui/pull/1736

mwithi (dismissed):
LGTM. Thanks a lot!

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -266,6 +267,41 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	public boolean newExamRequest(Laboratory laboratory) throws OHServiceException {
+		setPatientConsistency(laboratory);
+		LaboratoryStatus laboratoryStatus = LaboratoryStatus.DRAFT;
+		laboratory.setStatus(laboratoryStatus.toString());
+		laboratory.setResult("");
+		return ioOperations.newLabFirstProcedure(laboratory);
+	}
+	
+	/**
+	 * update one Laboratory request {@link Laboratory} (All Procedures)
```
dbmalkovsky, line 286:
```suggestion
	 * Update one Laboratory request {@link Laboratory} (All Procedures)
```

src/main/java/org/isf/lab/manager/LabManager.java
```diff
@@ -266,6 +267,41 @@ public boolean newLaboratory2(Laboratory laboratory, List<LaboratoryRow> labRow)
...
+	 * @return <code>true</code> if the request has been update, <code>false</code> otherwise
+	 * @throws OHServiceException
+	 */
+	public boolean updateExamRequest(int code, LaboratoryStatus status) throws OHServiceException {
+		Optional<Laboratory> laboratory = ioOperations.getLaboratory(code);
+		if (laboratory.isPresent()) {
+			Laboratory lab = laboratory.get();
+			lab.setStatus(status.toString());
+			return ioOperations.updateLabFirstProcedure(lab);
+		} else {
```
dbmalkovsky, line 298: No need for the `else` when the previous statement is a `return`

dbmalkovsky (approved):
LGTM.  I made two small suggestions that can be ignored if it holds up committing this PR.

src/main/java/org/isf/lab/manager/LabManager.java
mwithi: Let's keep it this way for readability, thanks!

dbmalkovsky: FWIW, other methods have it without the `else`; just trying to get things consistent and without too much indentation 😏

PR #903, opened 2023-03-12
OP-1049 Fix error from OP-999

nlazzer1:
See OP-1049

### DIFF
--- a/src/main/java/org/isf/stat/manager/JasperReportsManager.java
+++ b/src/main/java/org/isf/stat/manager/JasperReportsManager.java
@@ -91,13 +91,15 @@ public class JasperReportsManager {
 	private static final String STAT_REPORTERROR_MSG = "angal.stat.reporterror.msg";
 
 	private static final String RPT_BASE = "rpt_base";
-
+	
+	
 	@Autowired
 	private HospitalBrowsingManager hospitalManager;
 
 	@Autowired
 	private DataSource dataSource;
 
+	
 	public JasperReportResultDto getExamsListPdf() throws OHServiceException {
 
 		try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -959,10 +959,9 @@ private void addSubReportsBundleParameters(String jasperFileFolder, String jaspe
 		}
 	}
 
-	private void addReportBundleParameter(String jasperParameter, String jasperFileFolder, String jasperFileName, Map<String, Object> parameters) {
+	private void addReportBundleParameter(String jasperParameter, String jasperFileName, Map<String, Object> parameters) {
 		try {
-			ResourceBundle resourceBundle = ResourceBundle.getBundle(jasperFileFolder + File.separator + jasperFileName, new Locale(GeneralData.LANGUAGE),
-							new UTF8Control());
+			ResourceBundle resourceBundle = ResourceBundle.getBundle(jasperFileName, new Locale(GeneralData.LANGUAGE), new UTF8Control());
```
dbmalkovsky, line 964: Can't `new Locale(GeneralData.LANGUAGE)` be a static class variable as you can't change the language without restarting OH.   That way it isn't reallocated for every report.   Just a thought ....

nlazzer1: Hi David, I restored that function to how it was before my last change, it was already like that. I'll move the creation to the beginning then as you suggest. Thank you very much

nlazzer1: The mainMenu set Locale.setDefault(new Locale(GeneralData.LANGUAGE)), so we can use the default locale

mwithi (commented):
LGTM. Thanks a lot!

mwithi (approved):

src/main/java/org/isf/stat/manager/JasperReportsManager.java
mwithi: is it ok now @dbmalkovsky? can we resolve the conversation?

dbmalkovsky (approved):
LGTM

src/main/java/org/isf/stat/manager/JasperReportsManager.java
dbmalkovsky: yes

PR #905, opened 2023-03-14
OP-1035 - Patient Consensus Management

goto-eof:
See OP-1035.

Please take a look and tell me if the table structure should be as I defined. Thank you.


sql/step_91_patient_consensuns_table.sql
```diff
@@ -0,0 +1,14 @@
...
+  PTC_ADMINISTRATIVE  TINYINT(1) NOT NULL DEFAULT 0,
+  PTC_SERVICE  TINYINT(1) NOT NULL DEFAULT 0,
+  PTC_CREATED_BY VARCHAR(50) NULL DEFAULT NULL,
+  PTC_CREATED_DATE datetime NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_DATE datetime NULL DEFAULT NULL,
+  SEA_ACTIVE TINYINT(1) NOT NULL DEFAULT 1,
+  PRIMARY KEY (PTC_ID),
+  FOREIGN KEY (PTC_US_ID_A) REFERENCES OH_USER(US_ID_A)
+) ENGINE = INNODB DEFAULT CHARACTER SET utf8;
```
dbmalkovsky, line 14: When adding a new table it is highly recommended that the `load_demo_data.sql` script is updated with sample data for the new table.

dbmalkovsky: And why is this table joined to the `user` table and not the `patient` table?     Users are the people who interact with OH, isn't it the patients who are giving their consensus?    Sorry if I misunderstand what is the purpose.

sql/step_91_patient_consensuns_table.sql
```diff
@@ -0,0 +1,14 @@
...
+  PTC_ID INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
+  PTC_US_ID_A VARCHAR(50) NOT NULL,
+  PTC_CONSENSUNS TINYINT(1) NOT NULL DEFAULT 0,
+  PTC_ADMINISTRATIVE  TINYINT(1) NOT NULL DEFAULT 0,
+  PTC_SERVICE  TINYINT(1) NOT NULL DEFAULT 0,
+  PTC_CREATED_BY VARCHAR(50) NULL DEFAULT NULL,
+  PTC_CREATED_DATE datetime NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_DATE datetime NULL DEFAULT NULL,
+  SEA_ACTIVE TINYINT(1) NOT NULL DEFAULT 1,
```
dbmalkovsky, line 11: Shouldn't this  be `PTC_ACTIVE`?

src/main/java/org/isf/patconsensus/service/PatientConsensusIoOperation.java
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Delete an {@link PatientConsensus}
+	 *
+	 * @param patientConsensus - the {@link PatientConsensus} to delete
+	 * @return <code>true</code> if the {@link PatientConsensus} has been delete, <code>false</code> otherwise.
+	 * @throws OHServiceException
+	 */
+	public boolean deletePatientConsensus(PatientConsensus patientConsensus) throws OHServiceException {
+		repository.delete(patientConsensus);
+		return true;
```
dbmalkovsky, line 83: I find this problematic.   I know the lots of the current code looks like this but we are slowly trying to clean it up.  The `delete` method has a `void` return.   Thus I feel this method should also return `void` and not force a `true` return value.

src/main/java/org/isf/patconsensus/model/PatientConsensus.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import javax.persistence.Transient;
+import org.isf.utils.db.Auditable;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
+
+@Entity
+@Table(name = "OH_SESSION_AUDIT")
```
dbmalkovsky, line 37: This table name is incorrect; it should be `OH_PATIENT_CONSENSUS`.

sql/step_91_patient_consensuns_table.sql
goto-eof: There are about 500 patients in the load_demo_data. Should I add a patient consensus for all of them?

dbmalkovsky: That is your call; but some number of them should have values.   Can't you write a small program to populate the records and then dump the db?

src/main/java/org/isf/patconsensus/model/PatientConsensus.java
```diff
@@ -0,0 +1,136 @@
...
+
+	@Column(name="PTC_ADMINISTRATIVE")
+	private boolean administrativeFlag;
+
+
+	@Column(name="PTC_SERVICE")
+	private boolean serviceFlag;
+
+
+    @OneToOne(cascade = CascadeType.ALL)
```
dbmalkovsky, line 69: Formatting looks out of line here; and there is lots of white space in the file.

src/main/java/org/isf/patient/model/Patient.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2022 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
dbmalkovsky, line 3: Why the back dating of the copyright year?  Changes are being made in 2023

src/main/java/org/isf/patient/model/Patient.java
dbmalkovsky, line 1: why is anamnesis and allergies removed from the Patient object; this seems like a step backwards.

src/main/java/org/isf/patient/model/Patient.java
```diff
@@ -615,21 +570,13 @@ public String getSearchString() {
...
-		if (getTelephone() != null) {
-			sbName.append(getTelephone()).append(' ');
-		}
-		if (getNote() != null) {
-			sbName.append(getNote().toLowerCase()).append(' ');
-		}
-		if (getTaxCode() != null) {
-			sbName.append(getTaxCode().toLowerCase()).append(' ');
-		}
+		if (getAddress() != null) sbName.append(getAddress().toLowerCase()).append(' ');
```
dbmalkovsky, line 573: why undoing the better/project standard formating?

goto-eof: Yes, you are true, I tried to restore the previous version 2 times, but seems that when I save, it indents automatically the code. I'll fix later. Thanks.

mwithi:
Structure looks good to me.

sql/step_91_patient_consensuns_table.sql
goto-eof: I have already fixed table structure:
image

sql/step_91_patient_consensuns_table.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE OH_PATIENT_CONSENSUS (
+  PTC_ID INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
+  PTC_PAT_ID int(11) NOT NULL,
+  PTC_CONSENSUNS TINYINT(1) NOT NULL DEFAULT 0,
```
mwithi, line 4:
```suggestion
  PTC_CONSENSUS TINYINT(1) NOT NULL DEFAULT 0,
```

sql/step_91_patient_consensuns_table.sql
mwithi, line 1: We also need a script to enable previous existing **real** data, something like:

`INSERT INTO OH_PATIENT_CONSENSUS (PTC_PAT_ID, PTC_CONSENSUNS) SELECT PAT_ID, 1 FROM OH_PATIENT WHERE PAT_DELETED = 'N';`

src/main/java/org/isf/patconsensus/service/PatientConsensusIoOperation.java
```diff
@@ -0,0 +1,106 @@
...
+	 * Delete an {@link PatientConsensus}
+	 *
+	 * @param patientConsensus
+	 *            - the {@link PatientConsensus} to delete
+	 * @return <code>true</code> if the {@link PatientConsensus} has been delete, <code>false</code> otherwise.
+	 * @throws OHServiceException
+	 */
+	public void deletePatientConsensus(PatientConsensus patientConsensus) throws OHServiceException {
+		repository.delete(patientConsensus);
+	}
```
mwithi, line 92: this should not actually delete the record, a patient without related consensus record returns an error:

```
Exception in thread "AWT-EventQueue-0" java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Unknown Source)
	at org.isf.patient.gui.PatientInsertExtended.getConsensus(PatientInsertExtended.java:2199)
	at org.isf.patient.gui.PatientInsertExtended.getJRightPanel(PatientInsertExtended.java:2190)
	at org.isf.patient.gui.PatientInsertExtended.getJDataPanel(PatientInsertExtended.java:391)
	at org.isf.patient.gui.PatientInsertExtended.getJContainPanel(PatientInsertExtended.java:375)
	at org.isf.patient.gui.PatientInsertExtended.initialize(PatientInsertExtended.java:354)
	at org.isf.patient.gui.PatientInsertExtended.<init>(PatientInsertExtended.java:338)
	at org.isf.admission.gui.AdmittedPatientBrowser.lambda$6(AdmittedPatientBrowser.java:829)
```

should it rather set all flags to "false"? in which case I would run a delete of a patient consensus record since patient itself cannot be hard deleted (but only soft deleted with `PAT_DELETED=Y|N`)

sql/step_04_all_following_steps.sql
```diff
@@ -84,3 +84,5 @@ source step_87_alter_table_opd.sql;
 source step_88_alter_table_patient.sql;
 source step_89_alter_tables_innodb.sql;
 source step_90_login_retries.sql;
+source step_91_patient_consensuns_table.sql;
+source step_92_consensus_old_records.sql;
```
mwithi, line 88: please merge these two scripts (there's no need to have them separated) as `step_92_patient_consensuns_table.sql` (because there is a new step_91 in core now. Thanks

mwithi (approved):

sql/step_04_all_following_steps.sql
```diff
@@ -84,5 +84,6 @@ source step_87_alter_table_opd.sql;
 source step_88_alter_table_patient.sql;
 source step_89_alter_tables_innodb.sql;
 source step_90_login_retries.sql;
-source step_91_alter_laboratory_with_status.sql;
-source step_92_add_dicom_data_table.sql;
\ No newline at end of file
+source step_91_patient_consensuns_table.sql;
+source step_92_add_dicom_data_table.sql;
+source step_93_alter_laboratory_with_status.sql;
```
mwithi, line 89:
```suggestion
source step_91_alter_laboratory_with_status.sql;
source step_92_add_dicom_data_table.sql;
source step_93_patient_consensus_table.sql;
```

sql/step_92_patient_consensuns_table.sql
```diff
@@ -0,0 +1,17 @@
...
+  PTC_CREATED_DATE datetime NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
+  PTC_LAST_MODIFIED_DATE datetime NULL DEFAULT NULL,
+  PTC_ACTIVE TINYINT(1) NOT NULL DEFAULT 1,
+  PRIMARY KEY (PTC_ID),
+  FOREIGN KEY (PTC_PAT_ID) REFERENCES OH_PATIENT(PAT_ID),
+  UNIQUE (PTC_PAT_ID)
+) ENGINE = INNODB DEFAULT CHARACTER SET utf8;
+INSERT INTO OH_PATIENT_CONSENSUS (PTC_PAT_ID, PTC_CONSENSUS, PTC_ADMINISTRATIVE, PTC_SERVICE, PTC_CREATED_BY, PTC_CREATED_DATE, PTC_LAST_MODIFIED_BY, PTC_LAST_MODIFIED_DATE)
+SELECT PAT_ID, 1, 0, 0, 'admin', NOW(), 'admin', NOW() FROM OH_PATIENT;
```
mwithi, line 17:
```suggestion
) ENGINE = INNODB DEFAULT CHARACTER SET utf8;

-- update previous data
INSERT INTO OH_PATIENT_CONSENSUS (PTC_PAT_ID, PTC_CONSENSUS, PTC_ADMINISTRATIVE, PTC_SERVICE, PTC_CREATED_BY, PTC_CREATED_DATE, PTC_LAST_MODIFIED_BY, PTC_LAST_MODIFIED_DATE)
SELECT PAT_ID, 1, 0, 0, 'admin', NOW(), 'admin', NOW() FROM OH_PATIENT;
```

mwithi (changes requested):
Please apply suggestions. 

In the PatientBrowserManager we should add the check we have now in "gui", so it can be leveraged by "api" too

src/main/java/org/isf/patconsensus/model/PatientConsensus.java
```diff
@@ -0,0 +1,124 @@
...
+	@OneToOne(cascade = CascadeType.DETACH)
+	@JoinColumn(name = "PTC_PAT_ID", referencedColumnName = "PAT_ID")
+	private Patient patient;
+
+	public PatientConsensus() {
+		super();
+	}
+
+
+
```
mwithi, line 73:
```suggestion

```

src/main/java/org/isf/patconsensus/model/PatientConsensus.java
```diff
@@ -0,0 +1,124 @@
...
+	public PatientConsensus(boolean consensusFlag, boolean administrativeFlag, boolean serviceFlag, Patient patient) {
+		super();
+		this.consensusFlag = consensusFlag;
+		this.administrativeFlag = administrativeFlag;
+		this.serviceFlag = serviceFlag;
+		this.patient = patient;
+	}
+
+
+
```
mwithi, line 83:
```suggestion

```

src/main/java/org/isf/patconsensus/model/PatientConsensus.java
```diff
@@ -0,0 +1,118 @@
...
+	public PatientConsensus(boolean consensusFlag, boolean administrativeFlag, boolean serviceFlag, Patient patient) {
+		super();
+		this.consensusFlag = consensusFlag;
+		this.administrativeFlag = administrativeFlag;
+		this.serviceFlag = serviceFlag;
+		this.patient = patient;
+	}
+	public int getId() {
+		return id;
+	}
```
mwithi, line 80:
```suggestion
		super();
	}
	
	public PatientConsensus(boolean consensusFlag, boolean administrativeFlag, boolean serviceFlag, Patient patient) {
		super();
		this.consensusFlag = consensusFlag;
		this.administrativeFlag = administrativeFlag;
		this.serviceFlag = serviceFlag;
		this.patient = patient;
	}
	
	public int getId() {
		return id;
	}
```

src/main/java/org/isf/patient/manager/PatientBrowserManager.java
```diff
@@ -406,11 +406,14 @@ protected void validatePatient(Patient patient) throws OHDataValidationException
 		if (' ' == patient.getSex()) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.patient.pleaseselectpatientssex.msg")));
 		}
+		if (!patient.getPatientConsensus().isConsensusFlag()) {
```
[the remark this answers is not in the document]
goto-eof: yeah, you are true. I'll fix them.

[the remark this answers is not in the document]
goto-eof: fixed, thanks

mwithi (approved):

mwithi (changes requested):
It works fine, but via API one can still `PUT` the `consensusFlag` to `false` using the `updatePatientConsensus` method...

We should either add a validation in the `updatePatientConsensus` flag or move that validation from `PatientBrowserManager` to the `PatientConsensusBrowserManager` and try to leverage it when saving a patient.

I think also that `newPatientConsensus` method is not needed anymore.

goto-eof:
Why validate PUT method? A patientConsensus have to have always the consensus flag=true? A user can't modify the consenus?

mwithi:
> Why validate PUT method? A patientConsensus have to have always the consensus flag=true? A user can't modify the consenus flag?

exactly, we already refuse the patient creation without that flag=true in core and so in api

src/main/java/org/isf/patient/model/PatientProfilePhoto.java
```diff
@@ -53,17 +55,17 @@ public class PatientProfilePhoto implements Serializable {
 
 	@Column(name="PAT_PHOTO")
 	@Lob
-	private byte[] photo;
+	private Byte[] photo;
```
dbmalkovsky, line 58: I don't understand the desire to user objects when the corresponding primitives exist.   What am I missing?  Why is this done?

goto-eof: I needed it for the converter, for automatization. The ModelMapper library accepts generic objects.

goto-eof: Please take a look to the PatientMapping class

mwithi: Primitive vs Object is a topic: it affects performances (I don't know how much though) but also it depends if we need to set a field as `null` rather its default value. I think that there is not a one-for-all rule, but in case we have some fields that might be "not set", we should use Object types, otherwise primitives are enough.

In this case, the ModelMapper implementation requires it, otherwise a manual boxing/unboxing needs to be done anyway.

Am I wrong?

goto-eof: Ok, I'll restore the previous type

goto-eof: done

mwithi (approved):

src/main/java/org/isf/patient/model/PatientProfilePhoto.java
mwithi: with the new mapper something goes wrong on patient's picture read/write from "ui" (serialized picture)

mwithi: with latest changes now we have byte[] (primitive) and patient's picure read/write working from "ui".
thanks"!

mwithi (approved):
Latest changes work for me! Thanks a lot!

PR #910, opened 2023-03-19
OP-722 Validate that a visit does not overlap other visits

joel-costigliola:
See OP-722.

- avoid overlapping visits within the same ward
- avoid patients ubiquity, i.e. overlapping same patient visit in different wards
- also check the visit duration is > 0

### DIFF
--- a/src/main/java/org/isf/visits/manager/VisitManager.java
+++ b/src/main/java/org/isf/visits/manager/VisitManager.java
@@ -21,9 +21,15 @@
  */
 package org.isf.visits.manager;
 
+import static java.util.Comparator.comparing;
+import static java.util.stream.Collectors.toList;
+
 import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.isf.generaldata.MessageBundle;
 import org.isf.menu.manager.UserBrowsingManager;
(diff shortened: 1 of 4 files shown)


joel-costigliola:
Hi team,
First PR :tada:  
There is a bit of formatting differences in Visit.java, AFAIK it is the result of Intellij Idea formatting the file withstyle preferences from the project, let me know if that's ok or not.

src/test/java/org/isf/visits/test/Tests.java
```diff
@@ -349,7 +605,6 @@ public void testVisitEquals() throws Exception {
 		int id = setupTestVisit(false);
 		Visit visit = visitManager.findVisit(id);
 
-		assertThat(visit.equals(visit)).isTrue();
```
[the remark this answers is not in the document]
joel-costigliola: This was reported by Intellij Idea as being always true and thus pointless, if we want to keep this for code coverage, better for go  with `assertThat(visit).isEqualTo(visit);`

src/main/java/org/isf/visits/manager/VisitManager.java
```diff
@@ -299,4 +358,9 @@ private String prepareSmsFromVisit(Visit visit) {
 	public Visit findVisit(int id) throws OHServiceException {
 		return ioOperations.findVisit(id);
 	}
+
+	// helper method, TODO could be moved to org.isf.utils.exception.model if the dependency to org.isf.generaldata is ok
+	private static OHExceptionMessage newOHExceptionMessage(String message) {
```
mwithi, line 363: Yes, I would move it in `org.isf.utils.exception.model` or I would just add specific constructors for different `OHSeverityLevel`s, wdyt?

mwithi: (yes, dependency is ok)

src/test/java/org/isf/visits/test/Tests.java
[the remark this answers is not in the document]
mwithi: I would agree with @joel-costigliola suggestion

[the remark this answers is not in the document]
dbmalkovsky: LGTM I will add a task for myself to find other occurrences and "fix" them.   Thanks.

src/main/java/org/isf/visits/manager/VisitManager.java
joel-costigliola: A search for `new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title")` returns 216 results in the codebase and all of them use `OHSeverityLevel.ERROR`.

I believe a constructor  like `public OHExceptionMessage(String message)` that sets default title to `MessageBundle.getMessage("angal.common.error.title")` and severity to `OHSeverityLevel.ERROR` would be helpful since I think it's natural for an exception to be at the error level by default.

The use of the `WARNING` level when creating `OHExceptionMessage` always sets the title to `null`, ex: 
```java
new OHExceptionMessage(null, "The size should be between 0 and 250!", OHSeverityLevel.WARNING)
```
We could introduce a `public OHExceptionMessage(String message, OHSeverityLevel level)` if you find there is enough value but I would do that in a different PR, PR that could also replace the duplicated the code I mentionned

mwithi: It's ok for another PR, please can you create a Jira issue for that? Thanks!

mwithi:
I did a quick test and seems to work well! Ok for the PR with new bundles (only language_en.properties) in GUI, thanks!

src/main/java/org/isf/visits/manager/VisitManager.java
joel-costigliola: Sure, will do

joel-costigliola: Just to make sure we are on the same page, my understanding is that I will move `newOHExceptionMessage` to a constructor and create a JIRA issue to add `public OHExceptionMessage(String message, OHSeverityLevel level)` and refactor the code to use the `public OHExceptionMessage(String message)` introduced in this PR, correct?

mwithi: yep! for now you can create the issue well explained, then start to work on it after OP-722 is merged. Thanks!

joel-costigliola: JIRA issue created: https://openhospital.atlassian.net/browse/OP-1056

joel-costigliola:
I'll remove the TODO and rebase the PR on the latest develop

PR #927, opened 2023-03-31
Chore: remove redundant String operation; no need to use toString() on a String object

dbmalkovsky:


mwithi (approved):
LGTM. Thanks a lot!

PR #928, opened 2023-03-31
Chore: remove redundant type cast

dbmalkovsky:
`priceRepository.findByList_id(list.getId());` returns a `List<Prices>`; no need to cast it again


mwithi (approved):
LGTM. Thanks a lot!

PR #929, opened 2023-03-31
Chore: mismatched number of placeholders in logger call

dbmalkovsky:
Exception tracebacks in logger call do not use a placeholder


src/main/java/org/isf/sms/providers/gsm/GSMGatewayService.java
```diff
@@ -128,7 +128,7 @@ public boolean initialize() {
 			} catch (PortInUseException e) {
 				LOGGER.error("Port in use: {}", portId.getCurrentOwner());
 			} catch (Exception e) {
-				LOGGER.error("Failed to open port {} {}", portId.getName(), e);
```
mwithi, line 131: Are not two arguments? `portId.getName()` and `e`?

mwithi (approved):
LGTM. Thanks a lot!

PR #933, opened 2023-04-04
Chore: remove redundant field initialization code

dbmalkovsky:
Fields already default to `null`, `false`, `0', etc. no need to do it again; it is just a performance hit.


mwithi (approved):
Resolved conflicts by myself.
Thanks a lot!

PR #934, opened 2023-04-04
OP-1066 new OHExceptionMessage() constructor; packages that start with A

dbmalkovsky:
See OP-1066

Core packages that start with 'A`


mwithi (approved):
LGTM. Thanks a lot!

PR #935, opened 2023-04-05
Chore: add 's' to "If not, see <http://www.gnu.org/licenses/>" in header

dbmalkovsky:
Change: `If not, see <http://www.gnu.org/licenses/>`
To: `If not, see <https://www.gnu.org/licenses/>`

Also corrected the import ordering of one file


mwithi (approved):
LGTM. Thanks a lot!

PR #951, opened 2023-04-17
OP-897 use value of zero to turn off password retries checking

dbmalkovsky:
See OP-897

BTW, I reset the default value for STRONGLENTH to the value specified in the JIRA for OP-894.  I'll fix the doc with the changes for this issue are made.


mwithi (approved):
LGTM. Thanks a lot!

PR #958, opened 2023-04-24
OP-1073 Extend from Auditable class like other objects

dbmalkovsky:
See OP-1073

All tests ran cleanly.

### DIFF
--- a/src/main/java/org/isf/examination/model/PatientExamination.java
+++ b/src/main/java/org/isf/examination/model/PatientExamination.java
@@ -21,7 +21,6 @@
  */
 package org.isf.examination.model;
 
-import java.io.Serializable;
 import java.time.LocalDateTime;
 
 import javax.persistence.AttributeOverride;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/examination/model/PatientExamination.java
```diff
@@ -54,7 +54,7 @@
 @AttributeOverride(name = "lastModifiedBy", column = @Column(name = "PEX_LAST_MODIFIED_BY"))
 @AttributeOverride(name = "active", column = @Column(name = "PEX_ACTIVE"))
 @AttributeOverride(name = "lastModifiedDate", column = @Column(name = "PEX_LAST_MODIFIED_DATE"))
-public class PatientExamination implements Serializable, Comparable<PatientExamination> {
+public class PatientExamination extends Auditable<String> implements Comparable<PatientExamination> {
```
mwithi, line 57: Why `Serializable` has been removed too?

dbmalkovsky: Three reasons the class doesn't implement `Serializable`:
1. no other class that extends `Auditable` implements `Serializable`
2. the class contains the object `patient` which is a non-serializable field
3. the methods `readObject()` and/or `write Object()` are not defined

PR #968, opened 2023-05-02
OP-1067 OHExceptionMessage: Description should come from the argument and not be hardcoded

Verdoso:
It was probably a typo when the new constructor was created as the only place where it is being used is where ErrorDescription.PASSWORD_TOO_SHORT is passed as description.

I verfified that constructor is not used anywhere else in openhospital-core, openhospital-api or openhospital-gui, so no other code is affected.

### DIFF
--- a/src/main/java/org/isf/utils/exception/model/OHExceptionMessage.java
+++ b/src/main/java/org/isf/utils/exception/model/OHExceptionMessage.java
@@ -50,7 +50,7 @@ public class OHExceptionMessage implements Serializable{
 	public OHExceptionMessage(String title, ErrorDescription description, String message, OHSeverityLevel level) {
 		super();
 		this.title = title;
-		this.description = ErrorDescription.PASSWORD_TOO_SHORT;
+		this.description = description;
 		this.message = message;
 		this.level = level;
 	}


dbmalkovsky (commented):
I think this constructor should be removed as the `description` field is never referenced.   FWIW, I deleted the setter and getter description and rebuilt all three repos without error.

mwithi:
> I think this constructor should be removed as the `description` field is never referenced. FWIW, I deleted the setter and getter description and rebuilt all three repos without error.

It seems `getDescription()` is referenced in `org.isf.shared.exceptions.OHAPIError.OHAPIError(HttpStatus, OHServiceException)`, isn't it?

dbmalkovsky:
Yes but once again the field, the setters, or getters aren't used.

mwithi:
> Yes but once again the field, the setters, or getters aren't used. The code just copies what is there and then doesn't do anything with it.

Yes it's true... and eventually the `enum ErrorDescription` is useless too...
Anyway, according with OH2-182 we need to better understand what the front-end really needs... Let's clarify that first.

Verdoso:
Even if the getter is not used explicitely, it might be used by Jackson when serializing, meaning that the description is to be used in the ui project, through JSON, outside of what Java can detect.
And then, it might simply be that OH2-182 is not finished yet, but it will use it later on.

That's why I kept the change to the minimum that we know: that hardcoding the description is useless and should be changed.

IMHO, we should fix this thing that we know, and then if OH2-182 does not need the description in the end, that task should be the one in charge of removing the field.

mwithi:
> Even if the getter is not used explicitely, it might be used by Jackson when serializing, meaning that the description is to be used in the ui project, through JSON, outside of what Java can detect. And then, it might simply be that [OH2-182](https://openhospital.atlassian.net/browse/OH2-182) is not finished yet, but it will use it later on.
> 
> That's why I kept the change to the minimum that we know: that hardcoding the description is useless and should be changed.
> 
> IMHO, we should fix this thing that we know, and then if [OH2-182](https://openhospital.atlassian.net/browse/OH2-182) does not need the description in the end, that task should be the one in charge of removing the field.

I totally agree, let's fix one thing per time (ans so let's merge this). Thanks!

PR #975, opened 2023-05-08
OP-1082: Remove deprecated methods org.isf.exa.manager.ExamBrowsingManager

dbmalkovsky:
See OP-1082

Rebuilt the Core jar and rebuild GUI and API repos cleanly.


mwithi (approved):
LGTM. Thanks a lot!

PR #976, opened 2023-05-08
OP-1083: remove deprecated method from org.isf.patient.manager.PatientBrowserManager

dbmalkovsky:
See OP-1083

Rebuilt Core and then successfully rebuilt GUI.    There are corresponding changes required in API repo that in addition fix some "FIXME" comments


mwithi (approved):
LGTM. Thanks a lot!

PR #977, opened 2023-05-09
OP-1085: remove deprecated method from org.isf.visits.service.VisitsI…

dbmalkovsky:
See OP-1085

Rebuilt Core and ran all tests. Then successfully rebuilt GUI and ran tests (after making class name changes that are pending).    Then rebuilt API repo and ran all tests (again after making class name changes that are pending).


mwithi (approved):
LGTM. Thanks a lot!

PR #982, opened 2023-05-11
OP-1087 Remove deprecated methods from org.isf.accounting.manager.BillBrowserManager

dbmalkovsky:
Removed methods, rebuilt the jar, ran the unit tests.

Used new jar to rebuild GUI and API without errors. Ran all the API tests successfully.

Note:
Had to change the body to allow for the parameter zero to get all payments as that is what is supported at the IO level of the code:
`public List<BillPayments> getPayments(int billID) throws OHServiceException`


mwithi (approved):
LGTM. Thanks a lot!

PR #983, opened 2023-05-11
OP-1090 Remove deprecated methods from org.isf.medicalstockward.manager.MovWardBrowserManager

dbmalkovsky:
See OP-1090

Removed methods, rebuilt the jar, ran the unit tests.

Used new jar and rebuilt GUI and there were no issues.

Used new jar and rebuilt API and there were errors.   I commented out three methods that were causing problems.   I checked the UI code base and it appears that these "/medicalstockward" endpoints are not being used but only defined.   I left them in the API code for easier updating of the API and UI code bases but those more familiar with the UI.

See https://github.com/informatici/openhospital-api/pull/271


mwithi (approved):
LGTM. Thanks a lot!

PR #985, opened 2023-05-11
OP-1092 lock correct table

dbmalkovsky:
See OP-1092


mwithi (changes requested):
Please could you fix also `PTC_CONSENSUNS` → `PTC_CONSENSUS` ?
After I will also test the whole thing...

dbmalkovsky:
I was just going to open another issue for that because in my testing I found the same name misspelling.

mwithi (approved):

PR #995, opened 2023-05-20
OP-1035 - remove administrative flag

goto-eof:


sql/step_04_all_following_steps.sql
```diff
@@ -87,3 +87,4 @@ source step_90_login_retries.sql;
 source step_91_alter_laboratory_with_status.sql;
 source step_92_add_dicom_data_table.sql;
 source step_93_patient_consensus_table.sql;
+source step_94_remove_administrative_column.sql;
```
mwithi, line 90: There's no need of a new step since step_93 has never been released so far. We can modify step_93 and devs will align their DB.

sql/step_94_remove_administrative_column.sql
mwithi, line 1: There's no need of a new step since step_93 has never been released so far. We can modify step_93 and devs will align their DB.

mwithi (changes requested):
Please change step_93 instead of adding a step_94, thanks!

mwithi (approved):

PR #998, opened 2023-05-24
OP-1095, OP-1098: Added constraints to FKs against the oh_user table

Verdoso:
I started the PR so people can review the changes first.

The only notable thing is that I took the occasion to rename the oh_session_audit -> oh_user foreign key to align the name with the convention used in the other tables.


sql/step_94_missing_user_related_fk.sql
```diff
@@ -0,0 +1,50 @@
+ALTER TABLE `oh_admission`
```
mwithi, line 1: in scripts we use to CAPITALIZE all name tables and to remove single quotes

Verdoso: Fixed with commit [a4d091e](https://github.com/informatici/openhospital-core/pull/998/commits/a4d091e7703a11350f77a67450fcab1b5e17ab82)

Verdoso:
Added FKs to all XXX_created_by columns in commit  https://github.com/informatici/openhospital-core/pull/998/commits/b484ca684d52086ac5c09ca31e98a2021fcbcb47

sql/step_04_all_following_steps.sql
```diff
@@ -87,3 +87,4 @@ source step_90_login_retries.sql;
 source step_91_alter_laboratory_with_status.sql;
 source step_92_add_dicom_data_table.sql;
 source step_93_patient_consensus_table.sql;
+source step_94_missing_user_related_fk.sql;
```
dbmalkovsky, line 90: FWIW, both this PR and https://github.com/informatici/openhospital-core/pull/1002 use `step_94`.   One of them is going to have to change 😏

sql/step_94_missing_user_related_fk.sql
```diff
@@ -0,0 +1,550 @@
...
+
+ALTER TABLE OH_SESSION_AUDIT 
+ADD INDEX FK_SESSION_AUDIT_CREATED_BY_IDX (SEA_CREATED_BY ASC) VISIBLE;
+
+ALTER TABLE OH_SESSION_AUDIT 
+ADD CONSTRAINT FK_SESSION_AUDIT_CREATED_BY
+  FOREIGN KEY (SEA_CREATED_BY)
+  REFERENCES OH_USER (US_ID_A)
+  ON DELETE NO ACTION
+  ON UPDATE NO ACTION;
```
mwithi, line 470: This should be removed because override the previous one

mwithi: sorry, I missread the code... this one can stay but should be CASCADE as well

Verdoso: You're right. It would not make sense to leave this one as no_action if we have cascade in the other.

Verdoso: Solved by https://github.com/informatici/openhospital-core/pull/998/commits/78e084d4ec8e39ff6dbc85f99b83a49f96cc7cae

sql/step_04_all_following_steps.sql
Verdoso: Renamed file to step_95 with https://github.com/informatici/openhospital-core/pull/998/commits/455ca67ce1a5e874188203ebe2ae395ac0db9914

sql/step_95_missing_user_related_fk.sql
```diff
@@ -0,0 +1,550 @@
+ALTER TABLE OH_ADMISSION
```
dbmalkovsky, line 1: @mwithi  Shouldn't these changes be made to the load_demo_data.sql file also?

Verdoso: Two places where tables are defined? O_o
A recipe for trouble :D

dbmalkovsky: The load_demo_data.sql file is typically only a database dump of the database once the demo data is loaded.     That dump includes recreating the tables after dropping them.    Only for very "small" changes is the file modified directly.

mwithi: Yes, we should update the load_demo_data.sql with a new dump.
Anyway, I found some problems in the step_95 that we might adjust before, see my comment here https://openhospital.atlassian.net/browse/OP-1103?focusedCommentId=14998

PR #1002, opened 2023-05-27
OP-898: Password idle implementation

dbmalkovsky:
See  OP-898

Set the default idle time to 365 days


sql/load_demo_data.sql
```diff
@@ -6251,6 +6251,7 @@ CREATE TABLE `oh_user` (
   `US_ID_A` varchar(50) NOT NULL DEFAULT '',
   `US_UG_ID_A` varchar(50) NOT NULL DEFAULT '',
   `US_PASSWD` varchar(60) NOT NULL DEFAULT '',
+  `US_PASSWD` varchar(60) NOT NULL DEFAULT '',
```
mwithi, line 6254: is this adding real and so a mistake or it's just Github showing it?

image

sql/load_demo_data.sql
```diff
@@ -6272,8 +6274,8 @@ CREATE TABLE `oh_user` (
 
 LOCK TABLES `oh_user` WRITE;
 /*!40000 ALTER TABLE `oh_user` DISABLE KEYS */;
-INSERT INTO `oh_user` VALUES ('admin','admin','$2a$10$FI/PMO0oSHHosF2PX8l3QuB0DJepVfnynbLZ9Zm2711bF2ch8db2S','administrator',NULL,NULL,NULL,NULL,1,0,0,NULL);
-INSERT INTO `oh_user` VALUES ('guest','guest','$2a$10$b0WlANdaNV7Ukn/klFGt3.euZ7PaHuJI6TtBSM2vdxkavvkUDbpo2','guest',NULL,NULL,NULL,NULL,1,0,0,NULL);
+INSERT INTO `oh_user` VALUES ('admin','admin','$2a$10$FI/PMO0oSHHosF2PX8l3QuB0DJepVfnynbLZ9Zm2711bF2ch8db2S','administrator',NULL,NULL,NULL,NULL,1,0,0,NULL,NULL,NULL);
+INSERT INTO `oh_user` VALUES ('guest','guest','$2a$10$b0WlANdaNV7Ukn/klFGt3.euZ7PaHuJI6TtBSM2vdxkavvkUDbpo2','guest',NULL,NULL,NULL,NULL,1,0,0,NULL,NULL,NULL);
```
mwithi, line 6278: According to the new step_94, only one field is added to the OH_USER table, why are there two new NULL values?

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -129,38 +129,75 @@ public boolean deleteUser(User user) throws OHServiceException {
 		return ioOperations.deleteUser(user);
 	}
 
-
+	/**
+	 * Increase the number of failed logon attemptes for {@link User}.
```
mwithi, line 133: maybe this?

```suggestion
	 * Increase the number of failed login attemptes for {@link User}.
```

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -129,38 +129,75 @@ public boolean deleteUser(User user) throws OHServiceException {
...
+	 *
+	 * @param user the {@link User}
+	 */
 	public void increaseFailedAttempts(User user) {
 		int newFailAttempts = user.getFailedAttempts() + 1;
 		ioOperations.updateFailedAttempts(user.getUserName(), newFailAttempts);
 	}
 
+	/**
+	 * Reset the number of failed logon attemptes to zero for {@link User}.
```
mwithi, line 143: also here?

```suggestion
	 * Reset the number of failed login attemptes to zero for {@link User}.
```

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -129,38 +129,75 @@ public boolean deleteUser(User user) throws OHServiceException {
...
 	public void unlockUser(User user) throws OHServiceException {
 		user.setAccountLocked(false);
 		user.setLockedTime(null);
 		user.setFailedAttempts(0);
-		ioOperations.updateFailedAttempts(user.getUserName(), 0);
-		ioOperations.updateUserLocked(user.getUserName(), false, null);
+		String userName = user.getUserName();
+		ioOperations.updateFailedAttempts(userName, 0);
+		ioOperations.updateUserLocked(userName, false, null);
+		ioOperations.setLastLogin(userName, null);
```
mwithi, line 183: wondering if we can just persist "user" instead to have 3 different calls to service layer?

mwithi (dismissed):
Just a few comments, otherwise is ok for me, thanks!

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
dbmalkovsky: I thought about that but I was trying to make the minimum number of changes

mwithi: ok, maybe we can put a TODO comment and open a new issue (task type) for that, but I didn't make a complete analysis to suggest the content

dbmalkovsky: With the number of columns currently in the table the UPDATE command will have a long list of "sets".   Are we really sure we want that given how infrequently most these commands are called?  I why would we want to update all the columns just to set the last login date/time given that particular call *will* be called often.

mwithi: well, this happens for all entities, it is JPA that carries on the fact to update one or more columns, from service point of view it will be only a "entity.save()" method and the JPA repository does the rest.
It is allowed to do differently and it is not wrong though, but it looks to me an anti-pattern for very small entities as pointed out here https://www.baeldung.com/spring-data-partial-update#custom-query.
That's why I'm suggest to just leave a comment to explain WHY it is done differently from other entities.

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -129,38 +129,79 @@ public boolean deleteUser(User user) throws OHServiceException {
 		return ioOperations.deleteUser(user);
 	}
 
-
+	// TODO:  revisit the individual methods for failed attempts, locking, last login time, etc.
+	// The original idea is that last login in time gets called frequently, and number of failed attempts less oftent, and locking/unloking users
```
mwithi, line 133:
```suggestion
	// The original idea is that last login in time gets called frequently, and number of failed attempts less often, and locking/unloking users 
```

PR #1010, opened 2023-06-03
OP-1103 Set MyISAM engine on related tables

npasquetto:
This pull request addresses an issue where the container fails to start due to a "Foreign key constraint is incorrectly formed" error at step 95. The issue was caused by mismatched storage engines between the tables involved in the foreign key relationship.

To resolve this issue, I have made the following changes:

- Updated the SQL table creation files to align the storage engine used for all tables involved in the foreign key relationship. Now, all tables (PATIENTEXAMINATION, SUPPLIER) are using the same storage engine (MyISAM).

These changes ensure that the foreign key constraints are correctly formed and can function properly between the tables.

Please review the changes and let me know if any further modifications are required.

**Testing**

I have tested these changes locally by rebuilding and running the container using Docker Compose. The container now starts successfully without any errors related to foreign key constraints.

**Related Issue**

[OP-1098](https://openhospital.atlassian.net/browse/OP-1098)

Thank you for reviewing this pull request. Your feedback is highly appreciated.

[OP-1098]: https://openhospital.atlassian.net/browse/OP-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sql/step_37_suppliers_table.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE SUPPLIER (
 	SUP_NOTE VARCHAR(200) NULL,
 	SUP_DELETED char(1) DEFAULT 'N',
 	PRIMARY KEY (SUP_ID)
-);
+) ENGINE=MyISAM;
```
dbmalkovsky, line 13: I am confused; if I look at load_demo_data.sql I see at line 6176:
`) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;`

I only looked there because I had assumed that the change suggested here should have been made there too.

sql/step_36_patientexamination_and_rollback_step23.sql
```diff
@@ -13,7 +13,7 @@ CREATE TABLE PATIENTEXAMINATION (
 	
 	INDEX ( PEX_PAT_ID  ) ,
 	PRIMARY KEY ( PEX_ID )
-);
+) ENGINE=MyISAM;
```
dbmalkovsky, line 16: I am confused if I look in `load_demo_data.sql` at line 5155 I see:
`) ENGINE=InnoDB AUTO_INCREMENT=50 DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;`

I only looked there because I had assumed that the change suggested here should have been made there too.

dbmalkovsky: So there must be some place else that the engine type is being set.

npasquetto: I tested by building the container with docker compose, which doesn't use `load_demo_data.sql` as it does not load the demo data but takes care of creating the database structure

mwithi:
For what I remember MyIsam engine DOES NOT support foreign keys, so setting MyIsam simply causes the FKs to be ignored, hence make the step_95 useless...

Needs more investigations

npasquetto:
> I’ve reproduced the error:
> 
> ```
> openhospital-core-database-1  | ERROR 1005 (HY000) at line 385: Can't create table `oh`.`oh_patientexamination` (errno: 150 "Foreign key constraint is incorrectly formed")
> ```
> 
> So apparently is starting at line 385 of the new `step_95_missing_user_related_fk.sql`, to understand why.
> 
> For what I remember MyIsam engine DOES NOT support foreign keys, so setting MyIsam simply causes the FKs to be ignored, hence make the step_95 useless...


I experienced that set the same engine also on patient examination CREATE TABLE statement resolve the problem, because the fk can be created on same engine based tables (same for supplier table). 

> 
> Moreover, it happens only in Docker while from mysql console it’s just working fine.
> 
> Needs more investigation

Could it be that your test environment has MyISAM as the _default_ engine?

mwithi:
I did a complete analysis under the Jira issue OP-1103 ([link](https://openhospital.atlassian.net/browse/OP-1103?focusedCommentId=14998))

mwithi:
> I experienced that set the same engine also on patient examination CREATE TABLE statement resolve the problem, because the fk can be created on same engine based tables (same for supplier table).

changing to MyISAM causes to have all FK ignored because they are not supported by the engine, that's why the errors disappear

> Could it be that your test environment has MyISAM as the default engine?

The opposite, I have default engine=InnoDB

image

That's why I saw every FK in `step_95` correctly created , also for `OH_PATIENTEXAMINATION` and `OH_SUPPLIER` tables whose creations have no engine specified in the previous scripts.

But according to my research, the problem seems more related to _foreign keys checks_, that for some reason I don't need to disable when creating the DB in my local MariaDB server while it is needed in the Docker env...

mwithi:
replaced by https://github.com/informatici/openhospital-core/pull/1018

PR #1016, opened 2023-06-13
OH2-202: Create backend for users settings

ArnaudFonzam:
See OH2-202.


src/main/java/org/isf/menu/manager/UserSettingManager.java
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	public List<UserSetting> getUserSetting(String userId) throws OHServiceException {
+		return userSettingIoOperationRepository.findAllByUSerID(userId);
+	}
+	
+	/**
+	 * Returns {@link UserSetting}s of the specified userId.
+	 *
+	 * @param userId - the user Id
+	 * @param dashboard - the name of the user setting
```
dbmalkovsky, line 75: why isn't the parameter `userSetting` ?

src/main/java/org/isf/menu/manager/UserSettingManager.java
```diff
@@ -0,0 +1,81 @@
...
+
+	/**
+	 * Inserts a new {@link UserSetting} into the DB.
+	 *
+	 * @param userSetting - the {@link UserSetting} to insert
+	 * @return <code>{@link UserSetting}</code> if the userSetting has been inserted, <code>null</code> otherwise.
+	 * @throws OHServiceException
+	 */
+	public UserSetting newUserSetting(UserSetting userSetting) throws OHServiceException {
+
```
dbmalkovsky, line 46: extra blank line?

src/main/java/org/isf/menu/manager/UserSettingManager.java
```diff
@@ -0,0 +1,81 @@
...
+	}
+	
+	/**
+	 * Updates a upade {@link UserSetting} into the DB.
+	 *
+	 * @param userSetting - the {@link UserSetting} to insert
+	 * @return <code>{@link UserSetting}</code> if the userSetting has been updated, <code>null</code> otherwise.
+	 * @throws OHServiceException
+	 */
+	public UserSetting updateUserSetting(UserSetting userSetting) throws OHServiceException {
```
dbmalkovsky, line 57: why in some places is this referred to as `userSetting` and in others as `configName`?   I would suggest picking one or the other.

ArnaudFonzam: userSetting is object of UserSetting and configName is the property of UserSetting

mwithi (approved):
LGTM. Thanks a lot!

src/main/java/org/isf/menu/service/UserSettingOperationRepository.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+@Repository
+public interface UserSettingOperationRepository extends JpaRepository<UserSetting, Integer> {
+
+	@Query(value = "select usersetting from UserSetting usersetting where usersetting.user=:userId")
+	List<UserSetting> findAllByUSerID(@Param("userId") String userId);
+	
+	@Query(value = "select usersetting from UserSetting usersetting where usersetting.user=:userId and configName=:configName")
+	UserSetting findUserSettingDashboardByUSerID(@Param("userId") String userId, @Param("dashboard") String configName);
```
mwithi, line 40:
```suggestion
	UserSetting findUserSettingByUSerID(@Param("userId") String userId, @Param("configName") String configName);
```

src/main/java/org/isf/menu/manager/UserSettingManager.java
```diff
@@ -0,0 +1,80 @@
...
+	 * Returns {@link UserSetting}s of the specified userId.
+	 *
+	 * @param userId - the user Id
+	 * @param configName - the name of the user setting
+	 *  @return <code>{@link UserSetting}</code> if the userSetting exist, <code>null</code> otherwise.
+	 */
+	public UserSetting getUserSettingDashboard(String userId, String configName) throws OHServiceException {
+		return userSettingIoOperationRepository.findUserSettingDashboardByUSerID(userId, configName);
+	}
+}
```
mwithi, line 80:
```suggestion
	 * @return <code>{@link UserSetting}</code> if the userSetting exist, <code>null</code> otherwise.
	 */
	public UserSetting getUserSetting(String userId, String configName) throws OHServiceException {
		return userSettingIoOperationRepository.findUserSettingByUSerID(userId, configName);
	}
}
```

mwithi (changes requested):
Add suggestions to fix tests.
Also, we need to update demo data (just install demo data, apply changes with then final script, then make a dump in the same way of current load_demo_data.sql)

sql/step_96_user_settings_table.sql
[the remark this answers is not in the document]
ArnaudFonzam: no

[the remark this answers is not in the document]
mwithi: so what happen the first time a user access on an empty DB?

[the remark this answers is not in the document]
mwithi: In other words, I think we should insert at least these two records:

```
1	admin	dashboard	<value>
2	guest	dashboard	<value>
```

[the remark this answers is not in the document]
ArnaudFonzam: I have discuss with silevester my colleague to know if it is the best way to keep initial value in the database and he said that on the front-end path the logic that he has implement is : If user access to the  without having  config he generate the initial valeur  then save it and if the value change he update the also the value. wdyt about this approch?

mwithi (approved):

PR #1030, opened 2023-07-05
OP-1119: fix mismatched column and data values for OH_LABORATORY

dbmalkovsky:
See OP-1119

Found by seeing warning messages when running the script.  Interrupted the script and did `SHOW WARNINGS` and found data truncation messages.


dbmalkovsky:
Closing this PR and will replace it with one that updates the INSERT statements instead of rearranging the columns.

PR #1031, opened 2023-07-05
OP-1111 rework load demo data

dbmalkovsky:
See OP-1111

Suggested new flow:

step_WWW_demo_database.sql: the master script that includes:

1. step_XXX_create_database.sql:  deletes the OH database if it exists and creates a new OH database
2. step_YYY_create_tables.sql: just creates the tables/columns/keys/etc generated by mysqldump
3. step_ZZZ_load_data.sql: as the name implies loads the data into the tables.

Thus creation and loading is separated and the individual scripts can be used else where as needed.


dbmalkovsky:
Closing this to take a different approach.

PR #1050, opened 2023-07-24
Upgrade all junit dependencies

dbmalkovsky:
Combine all recent dependabot updates to Junit into a single PR.

Also removed a duplicate definition of the surefire plugin.

Ran `mvn clean test` without error.


mwithi (approved):
LGTM. Thanks a lot!

PR #1052, opened 2023-07-25
Chore: remove final from private methods

dbmalkovsky:
There is no need to define a `private` method as `final` as it can not be overridden.


dbmalkovsky:
closing; duplicate PR

PR #1053, opened 2023-07-25
Chore: remove unneeded toString()

dbmalkovsky:
There is no need to use `toString()` as the concatenation implicitly does it.


mwithi (approved):
LGTM. Thanks a lot!

PR #1056, opened 2023-07-25
Chore: use single character in String.indexOf() method

dbmalkovsky:
gets a small performance boast


mwithi (approved):
LGTM. Thanks a lot!

PR #1058, opened 2023-08-01
Chore: remove `final` on static methods

dbmalkovsky:
Specifying `final` on a `static` methods is redundant an the method can't be overridden anyway

### DIFF
--- a/src/main/java/org/isf/generaldata/ConfigurationProperties.java
+++ b/src/main/java/org/isf/generaldata/ConfigurationProperties.java
@@ -66,7 +66,7 @@ public abstract class ConfigurationProperties {
 	 * @param fileProperties - the file name (to be available in the classpath)
 	 * @param logger - the {@link Logger} of the concrete class
 	 */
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/generaldata/ConfigurationProperties.java
```diff
@@ -66,7 +66,7 @@ protected ConfigurationProperties(String fileProperties, boolean exitOnFail) {
 	 * @param fileProperties - the file name (to be available in the classpath)
 	 * @param logger - the {@link Logger} of the concrete class
 	 */
-	public static final Properties loadPropertiesFile(String fileProperties, Logger logger) {
+	public static Properties loadPropertiesFile(String fileProperties, Logger logger) {
```
mwithi, line 69: this method is public, was final meant to be removed as well?

dbmalkovsky: @mwithi yes it it to be removed.  Updated the text of the PR to reflect the fact that it isn't limited to just `private` methods.

PR #1059, opened 2023-08-03
OH2-227: Persist the opd ward that comes from the front end

ArnaudFonzam:
See OH2-227.

### DIFF
--- a/src/main/java/org/isf/opd/manager/OpdBrowserManager.java
+++ b/src/main/java/org/isf/opd/manager/OpdBrowserManager.java
@@ -89,10 +89,6 @@ public class OpdBrowserManager {
 		// Check Ward
 		if (ward == null) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.pleaseselectaward.msg")));
-		} else {
-			if (!ward.isOpd()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -89,10 +89,6 @@ protected void validateOpd(Opd opd, boolean insert) throws OHDataValidationExcep
 		// Check Ward
 		if (ward == null) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.pleaseselectaward.msg")));
-		} else {
-			if (!ward.isOpd()) {
-				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifiedwardisnotenabledforopdservice.msg")));
-			}
 		}
```
mwithi, line 92: Why to remove this check? this is important in case the API are called directly

ArnaudFonzam: ah okay yes

ArnaudFonzam: we can close this PRs

PR #1061, opened 2023-08-03
Chore: remove unused class

dbmalkovsky:
Class is not used and to make matters even more strange this `abstract` class extends from a `concrete` class which shouldn't be done.


mwithi (approved):
LGTM. Thanks a lot!

PR #1064, opened 2023-08-16
Fix: restore class used in GUI

dbmalkovsky:
Class is strange as it is an abstract class that extends from a concrete class which shouldn't be done.   And since it really is a GUI class it is in the wrong project IMHO.


mwithi:
any chance to rework this class and directly in gui?

mwithi (approved):
LGTM. Thanks a lot!

PR #1072, opened 2023-08-25
Chore: remove unused fields, variables, methods

dbmalkovsky:
Rebuilt the jar with these changes and then rebuild and tested GUI and API.


mwithi (approved):
LGTM. Thanks a lot!

PR #1073, opened 2023-08-25
Chore: fix assertThat() without condition

dbmalkovsky:
The primary change was to add a condition to the `assertThat()`.   As it was the statement was incomplete and was really a no-op.

So instead of just `assertThat(someObject);` the statement becomes something like `assertThat(someObject).isNotNull();`.

There are a few other minor changes to improve the tests.

### DIFF
--- a/src/test/java/org/isf/ward/test/Tests.java
+++ b/src/test/java/org/isf/ward/test/Tests.java
@@ -146,7 +146,7 @@ public class Tests extends OHCoreTestCase {
 		List<Ward> wards = wardIoOperation.getIpdWards();
 		
 		// then:
-		assertThat(wards.size() == 1);
+		assertThat(wards).hasSize(1);
 		
 		// given:
 		foundWard.setBeds(0);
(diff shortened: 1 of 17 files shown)


src/test/java/org/isf/ward/test/Tests.java
```diff
@@ -169,7 +169,7 @@ public void testIoGetOpdWards() throws Exception {
 		List<Ward> wards = wardIoOperation.getOpdWards();
 		
 		// then:
-		assertThat(wards.size() == 0);
+		assertThat(wards).hasSize(1);
```
mwithi, line 172: hasSize(0)

dbmalkovsky: No; the previous test had no condition so it didn't realize that `0` was wrong.   Notice all the tests passed so `1` is the correct value.

mwithi (approved):
LGTM. Thanks a lot!

PR #1077, opened 2023-09-04
Chore: remove AttributeOverrides annotation

dbmalkovsky:
Before Java 8, a container annotation was required as wrapper to use multiple instances of the same annotation. As of Java 8, this is no longer necessary. Instead, these annotations should be used directly without a wrapper, resulting in cleaner and more readable code.

All tests ran cleanly.


mwithi (approved):
LGTM. Thanks a lot!

PR #1082, opened 2023-09-05
Chore: remove unused methods

dbmalkovsky:
This started off by trying to replace the deprecated method in `repository.getById(code);`.   The replacement method is `getReferenceById(ID)`.  When searching for the new method I found that there were no other occurrences.  That led me to see where the method (`public Malnutrition getMalnutrition(int code)`) was used and I found that it was only used once in the Manager class and that method was never called or tested.   So this PR just removes the unused methods.

FWIW, I rebuilt the CORE jar with this change and then built the GUI project cleanly.


dbmalkovsky:
Closing this PR; it breaks the API repo. :frowning:

PR #1083, opened 2023-09-05
Chore: remove deprecated method

dbmalkovsky:
The deprecated method `getById(code)` is now replaced by `getReferenceById(code)`.

Here is the Javadoc
<pre>
	/**
	 * Returns a reference to the entity with the given identifier. Depending on how the JPA persistence provider is
	 * implemented this is very likely to always return an instance and throw an
	 * {@link javax.persistence.EntityNotFoundException} on first access. Some of them will reject invalid identifiers
	 * immediately.
	 *
	 * @param id must not be {@literal null}.
	 * @return a reference to the entity with the given identifier.
	 * @see EntityManager#getReference(Class, Object) for details on when an exception is thrown.
	 * @deprecated use {@link JpaRepository#getReferenceById(ID)} instead.
	 * @since 2.5
	 */
	@Deprecated
	T getById(ID id);
</pre>


mwithi (approved):
LGTM. Thanks a lot!

PR #1084, opened 2023-09-05
Chore: removed unused `insert` parameter

dbmalkovsky:
Started removing the unused `insert` parameter which was left over from 2015 when it was used to determine whether to close the DB connection.

Also noticed that the routines always returned `true` so there was no need to `and` the `result` from the calls.

Built the new jar and recompiled and tested GUI and API without error.


mwithi (approved):
LGTM. Thanks a lot!

PR #1086, opened 2023-09-05
Chore: use multiple exception blocks instead of instanceof

dbmalkovsky:


src/main/java/org/isf/dicom/manager/SourceFiles.java
```diff
@@ -127,12 +127,10 @@ private void loadDicomDir(FileDicom fileDicom, File sourceFile, int patient) thr
 			if (!value.isDirectory()) {
 				try {
 					loadDicom(fileDicom, value, patient);
+				} catch(OHDicomException ohDicomException) {
```
mwithi, line 130:
```suggestion
				} catch (OHDicomException ohDicomException) {
```

mwithi (approved):
LGTM. Thanks a lot!

PR #1087, opened 2023-09-05
Chore: remove unused method

dbmalkovsky:
Originally was drawn to this method to replace the deprecated (since JDK9) items: `getModifiers()` and `InputEvent.SHIFT_MASK`.   When trying to test the changes I found that the method was never used.     Rebuilt the CORE jar without the method and then successfully built and tested the GUI and API repos.   So the final "change" is to remove the method.


mwithi (approved):
LGTM. Thanks a lot!

PR #1092, opened 2023-09-07
Chore: use consistent form for float and double literals

dbmalkovsky:
The code used a mixture of formats for numeric literals so changed them to the most common form which has a value after the period (`1.0` instead of `1.`) and lower case for the numeric type (`f` instead of `F` and `d` instead of `D`),


mwithi (approved):
LGTM. Thanks a lot!

PR #1093, opened 2023-09-07
Chore: class fields moved to being local variables

dbmalkovsky:
Improve class size and readability of code


mwithi (approved):
LGTM. Thanks a lot!

PR #1094, opened 2023-09-11
OP-1125: More org.isf.utils.table package to GUI

dbmalkovsky:
See OP-1125

Companion PR to https://github.com/informatici/openhospital-gui/pull/1821


mwithi (approved):
LGTM. Thanks a lot!

PR #1101, opened 2023-09-25
Chore: remove pre-sized array

dbmalkovsky:
There are two styles to convert a collection to an array:

- A pre-sized array, for example, c.toArray(new String[c.size()])
- An empty array, for example, c.toArray(new String[0])

In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.

However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.


mwithi (approved):
LGTM. Thanks a lot!

PR #1102, opened 2023-09-25
OP-1130 improve JPA methods

dbmalkovsky:
See OP-1130

Changes for the accounting package in Core; no corresponding changes required for GUI but there are 3 required in API (which I will create a PR for if these changes are accepted).

### DIFF
--- a/src/main/java/org/isf/accounting/manager/BillBrowserManager.java
+++ b/src/main/java/org/isf/accounting/manager/BillBrowserManager.java
@@ -140,34 +140,37 @@ public class BillBrowserManager {
 	
 	/**
 	 * Stores a new {@link Bill} along with all its {@link BillItems} and {@link BillPayments}
-	 * @param newBill - the bill to store.
+	 * @param bill - the bill to store.
 	 * @param billItems - the list of bill's items
 	 * @param billPayments - the list of bill's payments
+	 * @returns the persisted Bill object
 	 * @throws OHServiceException
 	 */
 	@Transactional(rollbackFor=OHServiceException.class)
 	@TranslateOHServiceException
-	public void newBill(
-			Bill newBill,
+	public Bill newBill(
+			Bill bill,
 			List<BillItems> billItems,
 			List<BillPayments> billPayments) throws OHServiceException {
-		validateBill(newBill, billItems, billPayments);
-		int billId = newBill(newBill);
+		validateBill(bill, billItems, billPayments);
+		Bill newBill = newBill(bill);
+		int billId = newBill.getId();
 		if (!billItems.isEmpty()) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/isf/accounting/manager/BillBrowserManager.java
```diff
@@ -143,26 +143,22 @@ public List<BillPayments> getPayments(int billID) throws OHServiceException {
...
 	 * @param billItems - the list of bill's items
 	 * @param billPayments - the list of bill's payments
-	 * @return <code>true</code> if the bill has been stored, <code>false</code> otherwise.
-	 * @throws OHServiceException 
+	 * @throws OHServiceException
 	 */
 	@Transactional(rollbackFor=OHServiceException.class)
 	@TranslateOHServiceException
-	public boolean newBill(
+	public void newBill(
```
mwithi, line 150: according to the JPA javadoc, shouldn't this return the _saved_ Entity?

dbmalkovsky: Yes it good/should.   I was trying to make the **minimum** change as previously it **always** returned `true`.  As there are 3 entities in this method (bill, billitem, and payments) it seemed like just returning `Bill` was too simple.   I can change it if you want.

mwithi: what do you mean with "too simple"?

dbmalkovsky: It seems that 3 entities being created and returning only one seems not quite right/incomplete/etc. But without making major changes (like creating a class to return all 3) returning a Bill is probably the best.  I'll change it tomorrow.

mwithi: I've noticed now that Bill POJO has not BillItems and BillPayments in it, but it should... In this case I agree would be a major change (probably worth it) and we can tackle it in a separated issue.

mwithi (approved):
LGTM. Thanks a lot!

src/main/java/org/isf/accounting/manager/BillBrowserManager.java
mwithi: The changes could work, should we open a new issue for the 3-in-1 class ? 
(I would modify the Bill one rather than creating a new class)

dbmalkovsky: Yes; please open an issue

mwithi: Done! OP-1395

PR #1104, opened 2023-09-28
OP-1130 improve JPA methods - admission package

dbmalkovsky:
See OP-1130

Changes for the admission package in Core; requires corresponding changes and the GUI and  API repo for which I will create the two PRs.


mwithi (approved):
LGTM. Thanks a lot!

PR #1105, opened 2023-09-29
OP-1130 improve JPA methods - admtype package

dbmalkovsky:
See OP-1130

Changes for the admtype package in Core; requires corresponding changes int the GUI and the API repo for which I will create the two PRs.


dbmalkovsky:
Closing this PR to try a different approach.

dbmalkovsky:
The original approach is the one of choice.

mwithi (approved):
LGTM. Thanks a lot!

PR #1107, opened 2023-09-29
OP-1130 improve JPA methods - anamnesis package

dbmalkovsky:
See OP-1130

Changes for anamnesis package

No changes needed in GUI or API

### DIFF
--- a/src/main/java/org/isf/anamnesis/manager/PatientHistoryManager.java
+++ b/src/main/java/org/isf/anamnesis/manager/PatientHistoryManager.java
@@ -33,13 +33,11 @@ public class PatientHistoryManager {
 	private PatientHistoryIoOperations ioOperations;
 
 	public PatientHistory getByPatientId(int patientId) {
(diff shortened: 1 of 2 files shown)


mwithi (commented):
Aim of the OP-1130 is only to change service layer or also to propagate changes to the above stack up to gui? In the latter case we should change also manager and so api and gui could be affected then

dbmalkovsky:
The API code doesn't use the `PatientHistoryManager` class.

The GUI code doesn't seem to have any use for the returned object for save or update methods.

There is nothing else to change at the current time.

mwithi:
> The API code doesn't use the `PatientHistoryManager` class.
> 
> The GUI code doesn't seem to have any use for the returned object for save or update methods.
> 
> There is nothing else to change at the current time.

Yes, I agree with current changes

mwithi (approved):
LGTM. Thanks a lot!

PR #1108, opened 2023-09-30
OP-1130 improve JPA methods - dicomtype package

dbmalkovsky:
See OP-1130

Changes to dicomtype package

There are corresponding changes in GUI and no changes in API


mwithi (approved):
LGTM. Thanks a lot!

PR #1111, opened 2023-10-03
OP-1130 JPA admtype changes (take 2)

dbmalkovsky:
See OP-1130

This is a second attempt at implementing the issue.

The most significant change is the addition of `JPAImmutableTriple` which is used to return values from CRUD operations.  The attempt is to remove checked exceptions that litter the code and catch them in the `manager` level where possible and return appropriate information to the caller.

The `JPAImmutableTriple` has 3 fields (thus its name):

1. result:  true or false based on the success of the call
2. object:  the object being processed; the new object after save, the updated object after update, and null when the object is deleted
3. errors: a list of formatted string objects that are the associated error message text(s); if the operation was a success there are not messages and the value is null.

If this seems reasonable and acceptable I have the corresponding changes ready for GUI and API.

### DIFF
--- a/src/main/java/org/isf/admtype/manager/AdmissionTypeBrowserManager.java
+++ b/src/main/java/org/isf/admtype/manager/AdmissionTypeBrowserManager.java
@@ -27,14 +27,16 @@ import java.util.List;
 import org.isf.admtype.model.AdmissionType;
 import org.isf.admtype.service.AdmissionTypeIoOperation;
 import org.isf.generaldata.MessageBundle;
-import org.isf.utils.exception.OHDataValidationException;
 import org.isf.utils.exception.OHServiceException;
-import org.isf.utils.exception.model.OHExceptionMessage;
+import org.isf.utils.tuple.JPAImmutableTriple;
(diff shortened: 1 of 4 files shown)


mwithi (commented):
Despite I understand it's just a different way of doing things, it would have a big impact all over the codebase, included the modules where OP-1130 is not needed.

Nevertheless, I don't get the real benefit to have the same `return` object for any CRUD method that eventually the consumer (gui or api) have to `cast` to the wanted object in order to use it, it sounds a bit overcomplicated. Moreover, where the actual `try/catch` was a must because imposed by the `throws` in the manager, now becomes "optional" to the dev, that has to manually check for `!getErrors().isEmpty` or just ignore them. Finally, the new `tuple` should be documented in order to have devs to use it in place of standard java patterns, so it would impacts the documentation too and the review process a lot.

In other words, are we sure about the cost/benefits ratio of such change? I confess I did not try the code so maybe I'm not really able to deduce it.

dbmalkovsky:
The approach will not be used; closing.

PR #1124, opened 2023-10-13
OP-1130 improve JPA methods - hospital package

dbmalkovsky:
See OP-1130

Changes for hospital package.

Note there are no real code changes (except for the tests when the optional return of an object is checked for) but rather javadoc updates.

In addition there are no changes required for this package in the GUI and/or API repos.


mwithi (approved):
LGTM. Thanks a lot!

PR #1136, opened 2023-10-23
OH2-241 Update Examination default values

ArnaudFonzam:
See OH2-241.

I have use the values specifies in examination.properties to update this file

- add enum Bowel, Ausculation and Diurese


src/main/java/org/isf/generaldata/ExaminationParameters.java
```diff
@@ -95,6 +95,9 @@ public final class ExaminationParameters extends ConfigurationProperties {
 	
 	public static int SAT_MIN;
 	private static final int DEFAULT_SAT_MIN = 50;
+	
+	public static int SAT_MAX;
+	private static final int DEFAULT_SAT_MAX = 100;
```
mwithi, line 100: This is not needed because saturation cannot be above 100%

mwithi (changes requested):
Revert change on SAT_MAX

src/main/java/org/isf/generaldata/ExaminationParameters.java
ArnaudFonzam: I use that value to check if saturation give by use is less than that Max value

mwithi: I understand, but it is not something that can be decided by the user, it is a percentage :)
so I think you can just hardcode it or at least don't allow the user to modify it like other parameters in line 156

ArnaudFonzam: okay can I fix value 99 to that param

mwithi: why 99? saturation can be 100 and it could means hyperventilation...

mwithi: you can just put hardcode 100 or if you want you can keep it and remove the line 156 further down

mwithi, line 100:
```suggestion
	public static int SAT_MAX = 100;
```

src/main/java/org/isf/generaldata/ExaminationParameters.java
```diff
@@ -150,6 +153,7 @@ private ExaminationParameters(String fileProperties) {
 		TEMP_STEP = myGetProperty("TEMP_STEP", DEFAULT_TEMP_STEP);
 
 		SAT_MIN = myGetProperty("SAT_MIN", DEFAULT_SAT_MIN);
+		SAT_MAX = myGetProperty("SAT_MAX", DEFAULT_SAT_MAX);
```
mwithi, line 156:
```suggestion
```

mwithi (approved):

mwithi (approved):
LGTM. Thanks a lot!

PR #1156, opened 2023-11-07
OP-1130 improve JPA methods - sessionaudit package

dbmalkovsky:
See OP-1130

Changes for `sessionaudit` package

There are no corresponding changes for GUI and API packages.


mwithi (approved):
LGTM. Thanks a lot!

PR #1160, opened 2023-11-11
OP-1130 improve JPA methods - dicom package

dbmalkovsky:
See OP-1130

Changes for `dicom` package

Fixed some spelling issues too

Note there are no changes in the `API` repo


src/main/java/org/isf/dicom/service/DicomIoOperations.java
```diff
@@ -48,14 +48,14 @@ public class DicomIoOperations
 	private DicomIoOperationRepository repository;
 	
 	/**
-	 * Load a list of id file for series
+	 * Load a list of file iss for a series.
```
mwithi, line 51: Like this?
```suggestion
	 * Load a list of file id for a series.
```

mwithi (approved):
LGTM. Thanks a lot!

src/main/java/org/isf/dicom/service/DicomIoOperations.java
dbmalkovsky: yes and adding the `s` to `serie`;   `series` is both singular and plural :smile:

mwithi (approved):
LGTM. Thanks a lot!

PR #1165, opened 2023-11-14
OP-1142 | Add German language files

mizzioisf:
Add CSV files for full German language support 
https://openhospital.atlassian.net/browse/OP-1142


mwithi (approved):
LGTM. Thanks a lot!

PR #1172, opened 2023-11-22
OP-1130 improve JPA methods - operation package

dbmalkovsky:
See OP-1130

Changes for the `operation` package.

Note: this requires an additional message in the bundle file in the GUI repo.


mwithi (approved):
LGTM. Thanks a lot!

PR #1176, opened 2023-11-25
OP-1130 improve JPA methods - permissions package

dbmalkovsky:
See OP-1130

Changes for the `permissions` package

Note: not changes required in GUI but many changes required here for adding the `throws` clause that was missing.


mwithi (approved):
LGTM. Thanks a lot!

PR #1189, opened 2023-12-08
OP-1130 improve JPA methods - medicalstock package

dbmalkovsky:
See OP-1130

Changes for `medicalstock` package.

Note there are not changes for the GUI repo but there are a fair number of changes in this PR.

### DIFF
--- a/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
+++ b/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
@@ -138,8 +138,8 @@ public class MovStockInsertingManager {
 			}
 
 			List<Integer> medicalIds = ioOperations.getMedicalsFromLot(lot.getCode());
-			if (movement.getMedical() != null && !(medicalIds.isEmpty() || (medicalIds.size() == 1 && medicalIds.get(0).intValue() == movement
-					.getMedical().getCode().intValue()))) {
+			if (movement.getMedical() != null && !(medicalIds.isEmpty() || medicalIds.size() == 1 && medicalIds.get(0).intValue() == movement
+					.getMedical().getCode().intValue())) {
 				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.thislotreferstoanothermedical.msg")));
 			}
 			if (GeneralData.LOTWITHCOST && chargingType) {
@@ -159,7 +159,7 @@ public class MovStockInsertingManager {
 	 * Verify if the referenceNumber is valid for CRUD and return a list of errors, if any
 	 *
 	 * @param referenceNumber - the lot to validate
-	 * @return list of {@link OHExceptionMessage}
+	 * @return list of {@link OHExceptionMessage}s
 	 * @throws OHServiceException
 	 */
 	protected List<OHExceptionMessage> checkReferenceNumber(String referenceNumber) throws OHServiceException {
@@ -182,7 +182,7 @@ public class MovStockInsertingManager {
 	 * Retrieves all the {@link Lot} associated to the specified {@link Medical}, expiring first on top
 	 *
 	 * @param medical the medical.
-	 * @return the retrieved lots.
+	 * @return the list of retrieved {@link Lot}s.
 	 * @throws OHServiceException
 	 */
 	public List<Lot> getLotByMedical(Medical medical) throws OHServiceException {
@@ -197,7 +197,7 @@ public class MovStockInsertingManager {
 	 *
 	 * @param medicalSelected the selected medical.
 	 * @param specifiedQuantity the quantity provided by the user.
-	 * @return <code>true</code> if is under the limit, false otherwise.
+	 * @return {@code true} if is under the limit, {@code false} otherwise.
 	 * @throws OHServiceException
 	 */
 	public boolean alertCriticalQuantity(Medical medicalSelected, int specifiedQuantity) throws OHServiceException {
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -250,36 +248,44 @@ public boolean newMultipleChargingMovements(List<Movement> movements, String ref
...
 		return ioOperations.prepareChargingMovement(movement);
 	}
 
-	public boolean storeLot(String lotCode, Lot lot, Medical med) throws OHServiceException {
-		return ioOperations.storeLot(lotCode, lot, med);
+	/**
+	 * Stores the specified {@link Lot}.
+	 * @param lotCode the {@link Lot} code.
+	 * @param lot the lot to store.
+	 * @param medical
```
mwithi, line 283:
```suggestion
	 * @param medical the medical to store.
```

mwithi: sorry my bad

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -251,41 +243,39 @@ public boolean lotExists(String lotCode) throws OHServiceException {
 	 * @param lotCode the {@link Lot} code.
 	 * @param lot the lot to store.
 	 * @param medical
-	 * @return <code>true</code> if the lot has been stored, <code>false</code> otherwise.
+	 * @return the stored {@link Lot} object.
 	 * @throws OHServiceException if an error occurred storing the lot.
 	 */
-	public boolean storeLot(String lotCode, Lot lot, Medical medical) throws OHServiceException {
```
mwithi, line 257: I cannot recall WHY we are passing the `lotCode` and `medical` when we can just pass `lot`, but maybe is an heritage from previous logic. I think we'll need to check further (I did, without finding anything obvious at first sight)

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -366,6 +359,9 @@ public List<Movement> getMovements(String wardId, LocalDateTime dateFrom, LocalD
 		for (int i = 0; i < pMovementCode.size(); i++) {
 			Integer code = pMovementCode.get(i);
 			Movement movement = movRepository.findById(code).orElse(null);
+			if (movement == null) {
+				throw new OHServiceException(new OHExceptionMessage("Movement '" + code + "' not found."));
+			}
```
mwithi, line 364: Why is this needed?

mwithi:
> Note there are not changes for the GUI repo

Maybe just removing the `WardPharmacyRectify.this` like things?

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
dbmalkovsky: Because `findBy()` returns `Optional`so that is why `orElse(null)` is added.   Thus if the `code` is not found the object is `null`.  That can happen if the movement got deleted since the `find` in line 357 was executed.   It is just protecting against a very rare NPE.

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
dbmalkovsky: Are you sure?  It seems like the way the code is now written you can start with a `Lot` object that is close to what you want and update the `code` and `lot` while keeping the other fields like `dueDate`, `cost`, etc. Without much analysis I am not ready to change to code.

dbmalkovsky:
> > Note there are not changes for the GUI repo
> 
> Maybe just removing the `WardPharmacyRectify.this` like things?

Done; see https://github.com/informatici/openhospital-gui/pull/1900

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
mwithi: no... the `code` is like a primary key for the `lot` and every `lot` can be associated to one `medical` only, forever.
I recall that before, the Lot object had no link to its Medical other than through a MedicalStockMovement, after we added Medical reference in the Lot object to simplify some joins. I cannot recall about the code instead...

Anyway, maybe we can just add a TODO comment like "to verify why lotCode and medical params are needed"

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
mwithi: I agree upon very rare... in theory the `findMovementWhereDatesAndId` could return a `List<Movement>` directly and the for cycle could be just remoed. I don't remember why it is returing a `List<Integer>` 🤔

Also here, we could put just a reminder and handle it in another issue.

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
dbmalkovsky: Done

mwithi: Thanks!

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
dbmalkovsky: Added TODO comment

mwithi: Thanks!

PR #1191, opened 2023-12-13
OH2-255 Revise and align permissions for CORE, API and UI

mwithi:
See OH2-255.

- [x] redefine permissions: 166 permissions with their new `ID`s (using `TRUNCATE`)
- [x] redefine group permissions: 214 grouppermissions, assigned to both 'admin' and 'guest' users, using new `ID`s
- [x] fix `oh_user_settings` table (missing some fields that were added only in demo data)
- [x] move `/discharges` to `/admissions/discharges` endpoint and remove related permissions (`admissions.*` ones will be used)
- [x] fix demo data, including 'doctor' and 'laboratorist' users with their userSettings


dbmalkovsky (dismissed):
LGTM :)

dbmalkovsky:
The format of the `load_demo_data.sql` file has significantly changed.  For example the `INSERT` statement first relists all the column name and then there is a single `VALUES` instead of the multiple `INSERT` lines in the previous version.    Is this intentional? 

I ran the scripts and things looked OK as far as I can tell.

dbmalkovsky (approved):
The format of the load_demo_data.sql file has significantly changed. For example the INSERT statement first relists all the column names and then there is a single VALUES cause instead of the multiple INSERT lines in the previous version. Is this intentional?

I ran the scripts and things looked OK as far as I can tell.

mwithi:
> Is this intentional?

yes, they script was used to be like that previously because despite is more verbose it is easier to check and fix (without replace with a full dump)

mwithi:
> > Is this intentional?
> 
> yes, they script was used to be like that previously because despite is more verbose it is easier to check and fix (without replace with a full dump)

sorry, I've oversight because in my file editor lines were "wrapped", let me fix, thanks!

npasquetto (approved):

PR #1192, opened 2023-12-13
OH2-258 | Change the value of enum LaboratoryStatus in lowercase

ArnaudFonzam:
See OH2-258.


src/main/java/org/isf/lab/model/LaboratoryStatus.java
```diff
@@ -22,16 +22,5 @@
...
-
-	private LaboratoryStatus(String s) {
-		status = s;
-	}
-
-	@Override
-	public String toString() {
-		return this.status;
-	}
+	draft, open, done, invalid, deleted;
```
dbmalkovsky, line 25: Why are the values lower case?   I seem to find them typically in upper case and either Oracle show them in upper case; see this [blog for example](https://blogs.oracle.com/javamagazine/post/how-to-make-the-most-of-java-enums)

ArnaudFonzam: the problem is about the value of LaboratoryStatus on the spec file the value is on lower case and it is that value the ui use to fixe the value of LaboratoryStatus but when we try now to save or update using that value we have 400 error. That is why I have change it to lower case

dbmalkovsky: I don't know what the "spec file" is so I can't fully answer this but it seem you are "fixing" a symptom instead of "fixing" the root cause of the issue.   It just "smells" bad to me.

ArnaudFonzam: the spec file is the oh.yml generete by swagger and share with the ui. you can see it on openapi folder of openhospital-api projet

ArnaudFonzam: This is the PR that automatically generates the oh.yml [file:](https://github.com/informatici/openhospital-api/pull/380)

mwithi: yes, it seems that the `springdoc-openapi-maven-plugin` generates the expected values for enum always lowercase, despite we defined UPPERCASE in CORE and API, so the UI is instructed to send lowercase...

Anyway I do recall that we used this flavor because the UI and API were using UPPERCASE before while in CORE we needed lowercase in the DB (@dbmalkovsky we have other cases like that). But at that time we were generating and updating the `oh.yaml` manually (openapi specfile, contract between API and UI).

So I think all lowercases are ok for me, because they would be the same across the whole application and the DB as well, without changing the way we use in the code (despite naming conventions) or we should a way to trick the maven plugin to generate a correct enum format (I'm going to search a solution, if exists) instead of:
```
          description: Laboratory status
          example: DRAFT
          enum:
            - draft
            - open
            - done
            - invalid
            - deleted
```

mwithi: Here is explained the problem, but I don't think the solution is less complex than just this PR content (which instead simplify all a bit with less code)

https://stackoverflow.com/questions/68747036/how-can-have-springdoc-openapi-use-the-jsonvalue-enum-format-without-changing-t

mwithi: Thanks to @dbmalkovsky provided [link](https://blogs.oracle.com/javamagazine/post/how-to-make-the-most-of-java-enums), I think I've found a **the problem**.

The problem for the spec file is the `@Override toString()` method, that is returning `value`s instead of `name`s. Providing instead a specific method to return an ENUM("`value`") solves the problem in the spec file.
```
public enum LaboratoryStatus {

	DRAFT("draft"), OPEN("open"), DONE("done"), INVALID("invalid"), DELETED("deleted");

	private String status;

	private LaboratoryStatus(String s) {
		status = s;
	}

	// do not override toString()
	public String toValue() { 
		return this.status;
	}
}
```
then the oh.yaml
```
...
        status:
          type: string
          description: Laboratory status
          example: DRAFT
          enum:
          - DRAFT
          - OPEN
          - DONE
          - INVALID
          - DELETED
...
```

The downside here is that we have to update all the occurrences like these ones to the new method.

in api:
```
	private static String draft = LaboratoryStatus.DRAFT.toString(); -> .toValue()
	private static String open = LaboratoryStatus.OPEN.toString(); -> .toValue()
	private static String deleted = LaboratoryStatus.DELETED.toString(); -> .toValue()
	private static String invalid = LaboratoryStatus.INVALID.toString(); -> .toValue()
```
in gui:
```
	lab.setStatus(LaboratoryStatus.DONE.toValue());
```
and in the ui to go back using UPPERCASE values as exposed in the specs.


**Another way** for api could be delegate to the `LaboratoryMapper.java` to do those translations (but in gui the only way would be the one above)

But, still, I found the solution provided in this PR is more straightforward and less error prone for what we want to achieve (deal with lowercases values in the DB).

Wdyt?

ArnaudFonzam: if we remove the .toString() this will mean changing the type of the status property in the Laboratory class in the core to LaboratoryStatus, because at the moment the property is of type String.
This is the point I want to make if we remove the .toString().

mwithi: > if we remove the .toString() this will mean changing the type of the status property in the Laboratory class in the core to LaboratoryStatus, because at the moment the property is of type String.

yes, you are right, my bad.

> if we adopt lower case enums I think we'll have to change this too labDTO.setStatus(LaboratoryStatus.valueOf(laboratory.getStatus().toUpperCase())); to this labDTO.setStatus(LaboratoryStatus.valueOf(laboratory.getStatus())) in laboratoryController.

right!

> In ui, gui, api and core the enums will be lowercase. And don't forget that the Diuris, Ausculation and Bowel enums value are already in lowercase.

yes, it's what I meant when I said "we have the same pattern for all enums that are persisted in the DB".

ArnaudFonzam: sorry I like to say But or so

mwithi: I fixed in your comment 😉

mwithi (approved):
LGTM. Thanks a lot!

PR #1200, opened 2023-12-19
OP-1130 improve JPA methods - operation package

dbmalkovsky:
See [OP-1130](https://openhospital.atlassian.net/browse/OP-1130)

Changes for the operation package that were missed in the original update (the `delete` method should return `void`).  
I found them when working on another JIRA issue.


mwithi (approved):
LGTM. Thanks a lot!

PR #1204, opened 2023-12-20
OH2-252 | Fix audit logging from API

ArnaudFonzam:
See OH2-252.

- [x] add UserSession class
- [x] add also all the class that are use in UserSession class like : DelayTimer
- [x] update getCurrentUser() method
- [x] add property updatable = false in all the model class to avoid updating createBy and createDate column


src/main/java/org/isf/sessionaudit/model/UserSession.java
```diff
@@ -0,0 +1,44 @@
+package org.isf.sessionaudit.model;
```
dbmalkovsky, line 1: Missing license header.

mwithi: True, in GUI it was there

src/main/java/org/isf/sessionaudit/model/UserSession.java
mwithi, line 1: We need a PR on GUI where we remove this class from there

mwithi: or we extend it adding the `restartSession()` and override `setTimer()` with `org.isf.utils.jobjects.DelayTimer` as parameter

src/main/java/org/isf/sessionaudit/model/UserSession.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+import org.isf.menu.model.User;
+import org.isf.utils.time.DelayTimer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UserSession {
+
+	private static final int LOGIN_FAILED = 2;
```
mwithi, line 13: Not used

mwithi (dismissed):
LGTM. Thanks a lot!

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -48,7 +48,9 @@ public class UserBrowsingManager {
 	private MenuIoOperations ioOperations;
 
 	public static String getCurrentUser() {
-		return MDC.get("OHUser");
+		if (UserSession.getUser() != null)
+			return UserSession.getUser().getUserName();
```
mwithi, line 52:
```suggestion
		if (UserSession.getUser() != null) {
			return UserSession.getUser().getUserName();
		}
```

src/main/java/org/isf/utils/db/DefaultAuditorAwareImpl.java
```diff
@@ -0,0 +1,13 @@
+package org.isf.utils.db;
```
dbmalkovsky, line 1: Missing the license header.

dbmalkovsky (dismissed):
LGTW, after adding the license header

src/main/java/org/isf/utils/db/DefaultAuditorAwareImpl.java
mwithi: Done, thanks!

PR #1209, opened 2023-12-25
Chore: remove unnecessary fully qualified name

dbmalkovsky:


src/main/java/org/isf/sms/service/SmsOperations.java
```diff
@@ -34,7 +34,7 @@
 
 /**
  * @author Mwithi
- * @see org.isf.sms.model.Sms
+ * @see Sms
```
mwithi, line 37: Should it be like this?

```suggestion
 * @see {@link Sms}
```

dbmalkovsky: Done

mwithi (approved):
LGTM. Thanks a lot!

PR #1218, opened 2024-01-01
Chore: add test if present with findById() returning Optional

dbmalkovsky:
Fix tests that were not checking the Optional result of findById()


mwithi (approved):
LGTM. Thanks a lot!

PR #1220, opened 2024-01-05
OP-1163 revert update to dcm4ch3 jars because of JDK17

dbmalkovsky:
See OP-1163

Guess this is another reason to move to JDK 17 :smirk:


mwithi:
> See [OP-1163](https://openhospital.atlassian.net/browse/OP-1163)
> 
> Guess this is another reason to move to JDK 17 😏

I guess too

mwithi (approved):
LGTM. Thanks a lot!

PR #1221, opened 2024-01-07
OP-1165-fix-circular-reference-in-jpa-config

mwithi:
See OP-1165.

Tests performed with https://github.com/informatici/openhospital-api/pull/422.


src/main/java/org/isf/utils/db/AuditorAwareConfig.java
```diff
@@ -0,0 +1,29 @@
+package org.isf.utils.db;
```
dbmalkovsky, line 1: Missing license header

dbmalkovsky (approved):
After license header is added it LGTM

src/main/java/org/isf/utils/db/AuditorAwareConfig.java
mwithi: done, thanks!

PR #1234, opened 2024-01-13
OP-1166 Fix lot validation

mwithi:
See OP-1166.

### DIFF
--- a/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
+++ b/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
@@ -85,7 +85,7 @@ public class MovStockInsertingManager {
 		if (movement.getType() == null) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.pleasechooseatype.msg")));
 		} else {
-			chargingType = movement.getType().getType().contains("+"); //else discharging
+			chargingType = movement.getType().getType().contains("+"); // else discharging
 
 			// Check supplier
 			if (chargingType) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -112,42 +112,41 @@ protected void validateMovement(Movement movement, boolean checkReference) throw
...
+			}
 
-				if (lot.getPreparationDate() != null && lot.getDueDate() != null && lot.getPreparationDate().isAfter(lot.getDueDate())) {
-					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.thepreparationdatecannotbyaftertheduedate.msg")));
-				}
+			if (lot.getPreparationDate() != null && lot.getDueDate() != null && lot.getPreparationDate().isAfter(lot.getDueDate())) {
+				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.thepreparationdatecannotbyaftertheduedate.msg")));
 			}
+		}
+		List<Integer> medicalIds = ioOperations.getMedicalsFromLot(lot.getCode());
```
dbmalkovsky, line 134: Shouldn't this be inside the `if (lot != null)` check?  If `lot` is `null` then `log.getCode()` will cause a NPE.

dbmalkovsky (commented):
Added a question after review of the code.   Other than that I believe it looks OK

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
mwithi: Good catch, all of them indeed... let me revise the logic. Thanks!

mwithi: @dbmalkovsky please check now. Added also https://github.com/informatici/openhospital-gui/pull/1936. Thanks!

dbmalkovsky (approved):
LGTM

PR #1235, opened 2024-01-14
OP-1167 Add error for unexpected condition on lot management

mwithi:
See OP-1167.

Paired with https://github.com/informatici/openhospital-gui/pull/1935.

### DIFF
--- a/src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
+++ b/src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
@@ -39,6 +39,8 @@ import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
 import org.isf.utils.time.TimeTools;
 import org.isf.ward.model.Ward;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -96,12 +100,17 @@ public List<Integer> getMedicalsFromLot(String lotCode) throws OHServiceExceptio
...
 		int qty = movement.getQuantity(); // movement initial quantity
-		List<Movement> dischargingMovements = new ArrayList<>();
+
+		if (lots.isEmpty()) {
+			LOGGER.warn("No lots with available quantity found for medical {}", medical.getDescription());
+			return dischargingMovements;
+		}
 		for (Lot lot : lots) {
-			Movement splitMovement = new Movement(movement.getMedical(), movement.getType(), movement.getWard(),
+			Movement splitMovement = new Movement(medical, movement.getType(), movement.getWard(),
```
dbmalkovsky, line 113: `medical` is extracted into a local variable so it doesn't need to be fetched over and over; then why aren't `type` and `ward` done the same way?

dbmalkovsky (approved):
LGTM; given my comment is ignored.

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
mwithi: `type` and `ward` are extracted only once. Maybe we can do it for `lot.getCode()` inside the `for` loop?

dbmalkovsky: It is inside of a loop so for every iteration they are fetched, aren't they?

mwithi: Yes, I mean extract the lot.getCode() at the begin of the loop

mwithi: done anyway

PR #1240, opened 2024-01-15
OP-1137 additional validation upon OPD save/update

dbmalkovsky:
See OP-1137

### DIFF
--- a/src/main/java/org/isf/opd/manager/OpdBrowserManager.java
+++ b/src/main/java/org/isf/opd/manager/OpdBrowserManager.java
@@ -112,13 +112,13 @@ public class OpdBrowserManager {
 		} else {
 			// Check double diseases
 			if (disease2 != null && disease.getCode().equals(disease2.getCode())) {
-				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifyingduplicatediseasesisnotallowed.msg")));
+				errors.add(new OHExceptionMessage(MessageBundle.formatMessage("angal.opd.specifyingduplicatediseasesisnotallowed.fmt.msg", "1", "2")));
 			}
 			if (disease3 != null && disease.getCode().equals(disease3.getCode())) {
-				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifyingduplicatediseasesisnotallowed.msg")));
+				errors.add(new OHExceptionMessage(MessageBundle.formatMessage("angal.opd.specifyingduplicatediseasesisnotallowed.fmt.msg", "1", "3")));
 			}
 			if (disease2 != null && disease3 != null && disease2.getCode().equals(disease3.getCode())) {
-				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifyingduplicatediseasesisnotallowed.msg")));
+				errors.add(new OHExceptionMessage(MessageBundle.formatMessage("angal.opd.specifyingduplicatediseasesisnotallowed.fmt.msg", "2", "3")));
 			}
 		}
 		try {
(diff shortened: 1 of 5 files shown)


src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -114,6 +121,29 @@ protected void validateOpd(Opd opd, boolean insert) throws OHDataValidationExcep
...
+				if (opdDisease == null) {
+					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifieddiseaseisnoenabledforopdservice.msg")));
+				}
+			}
+			if (disease3 != null) {
+				opdDisease = diseaseBrowserManager.getOPDDiseaseByCode(disease3.getCode());
+				if (opdDisease == null) {
+					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.opd.specifieddiseaseisnoenabledforopdservice.msg")));
+				}
+			}
```
mwithi, line 143: instead to have 3 queries on the DB (`getOPDDiseaseByCode`) wouldn't be better to get the list (`getDiseaseOpd`) and cycle over it?

by the moment the error message is the one prompted to the user, wouldn't be better to specify also which disease the user should change?

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
```diff
@@ -66,7 +73,7 @@ protected void setPatientConsistency(Opd opd) {
 	 * @param insert {@code true} or updated {@code false}
 	 * @throws OHDataValidationException
 	 */
-	protected void validateOpd(Opd opd, boolean insert) throws OHDataValidationException {
+	public void validateOpd(Opd opd, boolean insert) throws OHDataValidationException {
```
mwithi, line 76: so far, "validate" methods are defined `protected` and never tested, should we start from this PR?

dbmalkovsky: The 'validate' methods are tested; they just require a lot more setup and infrastructure.   I was trying to show explicitly the testing of the new validation requirements.

BTW, the new unit testing setup that I proposed would "fix" this issue by not add `.test` to the package names.  That would allow tests to access `protected` methods directly.

In the short run the simple PR would be to remove `.test` from all the test package names.

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
dbmalkovsky: There are 90 OPD related diseases.  To fetch and iterate over the list seemed to me to be more expensive then a simple DB query that either returns `null` or a single record.

Besides at least in the sample data the vast majority of OPD records only have a single disease listed; so it is a single query that returns a lot less number of records then getting them all.

As for the error message I was following the example of duplicate diseases that didn't specify which ones were the duplicates.

mwithi: > There are 90 OPD related diseases. To fetch and iterate over the list seemed to me to be more expensive then a simple DB query that either returns null or a single record.
>
> Besides at least in the sample data the vast majority of OPD records only have a single disease listed; so it is a single query that returns a lot less number of records then getting them all.

Ok, I agree

> As for the error message I was following the example of duplicate diseases that didn't specify which ones were the duplicates.

Ok, but while the "duplicate" warning is somehow self-explaining and doesn't need to be verbose (I can "see" what I'm sending to the BE), for the diseases' internal properties it is better to explain IMO, because they depends by the specific installation settings provided by the admin and the client could not necessarily acknowledged about it. For this reason I would even roll back the `angal.opd.specifyingduplicatediseasesisnotallowed.fmt.msg`.

src/main/java/org/isf/opd/manager/OpdBrowserManager.java
mwithi: Yes, what I meant is that validate methods are not tested directly so far.

I'm working on the Admission validation tests as well (OP-1136), and I think we will end up with a unique approach, but in a different issue as you was suggesting in OP-911, leveraging for sure the `protected` modifier.

mwithi (approved):
LGTM. Thanks a lot!

PR #1241, opened 2024-01-16
OP-1168 Add LogViewer

mwithi:
See OP-1168.


src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -0,0 +1,64 @@
...
+			return absolutePath;
+		} else {
+			return "No appender found with the name 'RollingFile' in the root logger.";
+		}
+	}
+
+	/**
+	 * Opens the application log file ({@code openhospital.log}) folder with the System default file explorer.
+	 *
+	 * @param path the path
```
dbmalkovsky, line 56: This line should be removed; there are no parameters to this method

src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * Gets the application log file ({@code openhospital.log}) absolute path defined in the {@code log4j.properties}
+	 * 
+	 * @return
+	 */
+	public static String getLogFileAbsolutePath() {
+		Logger rootLogger = Logger.getRootLogger();
+		Appender appender = rootLogger.getAppender("RollingFile");
+
+		if (appender != null && appender instanceof DailyRollingFileAppender) {
```
dbmalkovsky, line 43:
```suggestion
		if (appender instanceof DailyRollingFileAppender) {
```
the `instanceof` handles `null` arguments

src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import org.apache.log4j.Appender;
+import org.apache.log4j.DailyRollingFileAppender;
+import org.apache.log4j.Logger;
+
+public class LogUtil {
+
+	/**
```
dbmalkovsky, line 34: This method only has static methods (it is a util class after all) so there should be a `private` constructor to stop someone trying to instantiate this class.

src/main/java/org/isf/utils/log/LogUtil.java
mwithi: Thanks

src/main/java/org/isf/utils/log/LogUtil.java
mwithi: Thanks!

src/main/java/org/isf/utils/log/LogUtil.java
mwithi: nice tip! thanks!

dbmalkovsky (dismissed):
LGTM; one little "improvement" 😏

src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+
+import org.apache.log4j.Appender;
+import org.apache.log4j.DailyRollingFileAppender;
+import org.apache.log4j.Logger;
+
+public class LogUtil {
+
+	private LogUtil() {
+		super();
```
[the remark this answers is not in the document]
mwithi: well, it always extends the Object class... 😛 I liked more than an empty constructor, but removed anyway.
Thanks!

mizzioisf (approved):
LGTM, I think we can merge. I will test "in production" before 1.14.0 release.

dbmalkovsky (approved):
LGTM

PR #1242, opened 2024-01-16
OP-1111 Remove tables drop-create from demo data dump

mwithi:
See OP-1111.

- used `--complete-insert=TRUE --extended-insert=FALSE --no-create-info=TRUE --skip-triggers`
- fixed two missing `TRUNCATE` statements in `delete_all_data.sql`


sql/delete_all_data.sql
```diff
@@ -58,6 +58,8 @@ TRUNCATE TABLE OH_SUPPLIER;
 TRUNCATE TABLE OH_THERAPIES;
 TRUNCATE TABLE OH_VERSION;
 TRUNCATE TABLE OH_VISITS;
+TRUNCATE TABLE OH_TELEMETRY;
+TRUNCATE TABLE OH_USER_SETTINGS;
```
dbmalkovsky, line 62: aren't there 2 tables missing:   OH_DICOM_DATA and OH_HELP?

mwithi: OH_DICOM_DATA yes.
OH_HELP has never been used so far, but we can add.
Thanks!

PR #1243, opened 2024-01-17
OP-1020 | DELETE last movement in Pharmaceutical Stock

ArnaudFonzam:
See OP-1020.


src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -290,5 +290,36 @@ public List<MedicalWard> getMedicalsWardTotalQuantity(char wardId) throws OHServ
 	public long countAllActiveMovementsWard() {
 		return this.movementRepository.countAllActiveMovementsWard();
 	}
+	
+	/**
+	 * Get the {@link MedicalWard} associated to specified criteria.
+	 * @param wardCode the ward code
+	 * @param medical the medical code
+	 * @param lotCode the lot code
```
dbmalkovsky, line 298:
```suggestion
	 * @param lot the lot code
```
The method parameter is `lot` and not `lotCode`

src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1194,6 +1196,13 @@ public void testMovementGetOrigin() throws Exception {
 		assertThat(movement).isNotNull();
 		assertThat(movement.getSupplier()).isEqualTo(movement.getOrigin());
 	}
+	
+	@Test
+	public void testDeleteLastMovement() throws Exception {
+		int code = setupTestMovement(false);
+		Optional<Movement> movement = movementIoOperationRepository.findById(code); 
+		assertAll(()-> movBrowserManager.deleteLastMovement(movement.get()));
```
dbmalkovsky, line 1204: Doesn't `deleteLastMovement()` return `void`?  So what is being asserted here?  Wouldn't be better to do the delete and then try to fetch the same record which should then fail which can then be tested?   That is the pattern used in other Core tests FWIW.

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -224,7 +224,40 @@ public List<MedicalWardForPrint> convertWardDrugs(Ward wardSelected, List<Medica
...
+	 * @throws OHServiceException.
+	 */
+	public void deleteMedicalWard(MedicalWard medWard) throws OHServiceException {
+		ioOperations.deleteMedicalWard(medWard);
+	}
+	
+	/**
+	 * Updates the  MedicalWard.
+	 *
+	 * @param MedicalWard - the MedicalWard to update.
```
dbmalkovsky, line 254:
```suggestion
	 * @param medWard - the MedicalWard to update. 
```
Use the correct parameter name.   And why isn't there a `{@link }` to the object?

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -224,7 +224,40 @@ public List<MedicalWardForPrint> convertWardDrugs(Ward wardSelected, List<Medica
...
 		return drugPrint;
 	}
-
+	
+	/**
+	 * Get the  MedicalWard for the specified criteria.
+	 *
+	 * @param wardCode - the ward id  
+	 * @param medical - the mecical id 
+	 * @param lot - the lot id
```
dbmalkovsky, line 233:
```suggestion
	 * @param lotCode - the lot id
```
Use the correct parameter name.

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +152,60 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+	 * @return the retrieved movement.
+	 * @throws OHServiceException
+	 */
+	public Movement getLastMovement() throws OHServiceException {
+		return ioOperations.getLastMovement();
+	}
+
+	/**
+	 * Deletes the last Movement.
+	 *
```
dbmalkovsky, line 168: Missing `@parm lastMovement`

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -25,21 +25,51 @@
...
+	private MovWardBrowserManager movWardBrowserManager;
+
+	public MovBrowserManager(MedicalStockIoOperations ioOperations, LotIoOperationRepository lotRepository, MedicalsIoOperations medicalsIoOperation,
+					MedicalDsrStockMovementTypeBrowserManager medicalDsrStockMovTypeManager, MovWardBrowserManager movWardBrowserManager) {
+		this.ioOperations = ioOperations;
+		this.lotRepository = lotRepository;
+		this.medicalsIoOperation = medicalsIoOperation;
+		this.medicalDsrStockMovTypeManager = medicalDsrStockMovTypeManager;
+		this.movWardBrowserManager = movWardBrowserManager;
+
```
dbmalkovsky, line 70: blank line not needed

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +151,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+	 * @throws OHServiceException
+	 */
+	public Movement getLastMovement() throws OHServiceException {
+		return ioOperations.getLastMovement();
+	}
+
+	/**
+	 * Deletes the last Movement.
+	 *
+	 *@param lastMovement - the last movement to delete
```
dbmalkovsky, line 168:
```suggestion
	 * @param lastMovement - the last movement to delete
```

sql/step_100_add_delete_btn_in_medicalstock.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO `oh_menuitem` VALUES ('btnpharmstockcmovdelete','angal.common.delete.btn','angal.common.delete.btn','x','X','medicalstock','none','N',3);
```
dbmalkovsky, line 1: Please see the naming convention used in another PR:   https://github.com/informatici/openhospital-core/pull/1241/files#diff-30fbda93a84498a0e72d8670e216445f20349198a65e4cac507c97e2336cf4c2  and https://github.com/informatici/openhospital-core/pull/1241/files#diff-478eae7600e2ef8b726e86922e614a877a4f9d8a4be8e81d39a8990d3959953b

dbmalkovsky: Don't you also need to add something to `load_demo_data.sql`

ArnaudFonzam: okay I added also now

src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1194,6 +1196,15 @@ public void testMovementGetOrigin() throws Exception {
 		assertThat(movement).isNotNull();
 		assertThat(movement.getSupplier()).isEqualTo(movement.getOrigin());
 	}
+	
+	@Test
+	public void testDeleteLastMovement() throws Exception {
+		int code = setupTestMovement(false);
+		Optional<Movement> movement = movementIoOperationRepository.findById(code); 
+		movBrowserManager.deleteLastMovement(movement.get());
```
dbmalkovsky, line 1204: The previous statement returns an `Optional` which means it might be there or not there; yet you don't test for it and assume it is OK.   Please be more careful in your coding;

Yes I would agree that because it was just set up it should still be there but there is no guarantee that something unexpected happened.

sql/load_demo_data.sql
```diff
@@ -2074,6 +2074,7 @@ INSERT INTO `oh_groupmenu` (`GM_ID`, `GM_UG_ID_A`, `GM_MNI_ID_A`, `GM_ACTIVE`, `
 INSERT INTO `oh_groupmenu` (`GM_ID`, `GM_UG_ID_A`, `GM_MNI_ID_A`, `GM_ACTIVE`, `GM_CREATED_BY`, `GM_CREATED_DATE`, `GM_LAST_MODIFIED_BY`, `GM_LAST_MODIFIED_DATE`) VALUES (333,'doctor','usersusers',0,NULL,NULL,NULL,NULL);
 INSERT INTO `oh_groupmenu` (`GM_ID`, `GM_UG_ID_A`, `GM_MNI_ID_A`, `GM_ACTIVE`, `GM_CREATED_BY`, `GM_CREATED_DATE`, `GM_LAST_MODIFIED_BY`, `GM_LAST_MODIFIED_DATE`) VALUES (334,'doctor','groups',0,NULL,NULL,NULL,NULL);
 INSERT INTO `oh_groupmenu` (`GM_ID`, `GM_UG_ID_A`, `GM_MNI_ID_A`, `GM_ACTIVE`, `GM_CREATED_BY`, `GM_CREATED_DATE`, `GM_LAST_MODIFIED_BY`, `GM_LAST_MODIFIED_DATE`) VALUES (335,'admin','telemetry',1,NULL,NULL,NULL,NULL);
+INSERT INTO `oh_groupmenu` (`GM_ID`, `GM_UG_ID_A`, `GM_MNI_ID_A`, `GM_ACTIVE`, `GM_CREATED_BY`, `GM_CREATED_DATE`, `GM_LAST_MODIFIED_BY`, `GM_LAST_MODIFIED_DATE`) VALUES (336, 'admin','btnpharmstockcmovdelete',0,NULL,NULL,NULL,NULL);
```
mwithi, line 2077: When we add values to tables in demo data we should also increase the AUTOINCREMENT value (see line 1742)
The best way, though, is to just apply new script to a demo DB, insert new example data for the new features (optional) and then have a mysqldump with the same options (`--complete-insert=TRUE --extended-insert=FALSE --no-create-info=TRUE --skip-triggers` see OP-1111)

sql/step_04_all_following_steps.sql
```diff
@@ -92,4 +92,5 @@ source step_95_missing_user_related_fk.sql;
 source step_96_user_settings_table.sql;
 source step_97_fix_default_values.sql;
 source step_98_reset_permissions.sql;
-source step_99_telemetry.sql;
\ No newline at end of file
+source step_99_telemetry.sql;
+source step_100_add_delete_btn_in_medicalstock.sql;
```
mwithi, line 96: Please call it like that, we will rename it just before the merge. Thanks!

```suggestion
source step_aXXX_add_delete_btn_in_medicalstock.sql;
```

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +151,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+		Medical medical = lastMovement.getMedical();
+		int medicalCode = medical.getCode();
+		int quantity = lastMovement.getQuantity();
+		if (movType.getType().equals("+")) {
+			medical.setInqty(medical.getInqty() - quantity);
+			medicalsIoOperation.updateMedical(medical);
+			List<Integer> lastMovWithSameLot = ioOperations.getMedicalsFromLot(lotCode);
+			if (lastMovWithSameLot.size() == 1) {
+				lotRepository.deleteById(lotCode);
+			}
```
mwithi, line 185: If you want to be sure that the `Movement` you are deleting was creating a new lot from scratch, so it is safe to delete it, you may want to check if the same lot is used in other movements, so I think you can use the `findByLot` method for that.

Moreover the `getMedicalsFromLot` retrieves a list of `Medicals` from the lotCode, not `Movement`s. Since Lot<->Medical, its size will be always == 1. Indeed, I think that that method is an old code that was trying to retrieve a Medical from lots from movements: since `step_66_lot_information_at_the_wards_level.sql` we have Medical information in the Lot table itself with `LT_MDSR_ID`, so that method can be totally reworked and simplified.

Also, another method that can be safely deleted is `findAllByMedicalCodeOrderByLot_` because is identical to `findAllByMedicalCode` and never used.

If you want to adress also the last two problems above you are welcome, otherwise we can do it in another issue.

Thanks!

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +151,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+	 */
+	@Transactional
+	public void deleteLastMovement(Movement lastMovement) throws OHServiceException {
+
+		MovementType movType = medicalDsrStockMovTypeManager.getMovementType(lastMovement.getType().getCode());
+		String lotCode = lastMovement.getLot().getCode();
+		Medical medical = lastMovement.getMedical();
+		int medicalCode = medical.getCode();
+		int quantity = lastMovement.getQuantity();
+		if (movType.getType().equals("+")) {
```
mwithi, line 179: We had problems with some charset with "+", please use `contains`

```suggestion
		if (movType.getType().`equals`("+")) {
```

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +151,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+			Ward ward = lastMovement.getWard();
+			String wardCode = ward.getCode();
+			MedicalWard medWard = movWardBrowserManager.getMedicalWardByWardAndMedical(wardCode, medicalCode, lotCode);
+			medWard.setIn_quantity(medWard.getIn_quantity() - quantity);
+			if (medWard.getIn_quantity() == medWard.getOut_quantity()) {
+				movWardBrowserManager.deleteMedicalWard(medWard);
+			} else {
+				movWardBrowserManager.updateMedicalWard(medWard);
+			}
+			medical.setInqty(medical.getInqty() - quantity);
```
mwithi, line 197: Since it is a discharge movement, shouln't you deduct from outgoing quantity?

```suggestion
			medical.setOutqty(medical.getOutqty() - quantity);
```

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +151,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+		} else {
+			Ward ward = lastMovement.getWard();
+			String wardCode = ward.getCode();
+			MedicalWard medWard = movWardBrowserManager.getMedicalWardByWardAndMedical(wardCode, medicalCode, lotCode);
+			medWard.setIn_quantity(medWard.getIn_quantity() - quantity);
+			if (medWard.getIn_quantity() == medWard.getOut_quantity()) {
+				movWardBrowserManager.deleteMedicalWard(medWard);
+			} else {
+				movWardBrowserManager.updateMedicalWard(medWard);
+			}
```
mwithi, line 196: I think this is the tricky part: after a discharge `Movement`, the `MedicalWard` could have already been used in the ward itself or to patients, and in such case we cannot delete it anymore. So I think the we need to consider that, like if there are not `MovementWard` for that `MedicalWard` after the `Movement` date, or at least I can't think of anything else at the moment.

After that, if you want to understand if to update the MedicalWard or just delete it (because it was not there before) I think you have to check both `quantity == medWard.getIn_quantity() && medWard.getOut_quantity() == 0`... if so, it means that it has never been moved out from the ward and that the only IN movement was the one we are deleting now.

Comparing `medWard.getIn_quantity() == medWard.getOut_quantity()` doen't make sense to me, but maybe I'm missing some reasoning here.

src/main/java/org/isf/medicalstock/service/MovementIoOperationRepository.java
```diff
@@ -59,9 +59,13 @@ public interface MovementIoOperationRepository extends JpaRepository<Movement, I
 	@Query(value = "select max(mov.date) from Movement mov")
 	LocalDateTime findMaxDate();
 
-	@Query(value = "select mov.refNo from Movement mov where mov.refNo like :refNo")
-	List<String> findAllWhereRefNo(@Param("refNo") String refNo);
+	@Query(value = "select mov from Movement mov where mov.refNo = :refNo")
+	List<Movement> findAllWhereRefNo(@Param("refNo") String refNo);
+	
+	@Query(value = "SELECT * FROM OH_MEDICALDSRSTOCKMOV ORDER BY MMV_ID DESC limit 1", nativeQuery = true)
```
mwithi, line 65: Why not like this? 

```suggestion
	@Query(value = "select m from Movement order by m.code DESC limit 1")
```

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -224,7 +224,40 @@ public List<MedicalWardForPrint> convertWardDrugs(Ward wardSelected, List<Medica
...
+	 * @param lotCode - the lot id
+	 * @return the retrieved medical.
+	 * @throws OHServiceException if an error occurs retrieving the medical.
+	 */
+	public MedicalWard getMedicalWardByWardAndMedical(String wardCode, int medical, String lotCode) throws OHServiceException {
+		return ioOperations.getMedicalWardByWardAndMedical(wardCode, medical, lotCode);
+	}
+	
+	/**
+	 * Deletes the  MedicalWard.
```
mwithi, line 242: also here

```suggestion
	 * Deletes the specified {@link MedicalWard}.
```

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -224,7 +224,40 @@ public List<MedicalWardForPrint> convertWardDrugs(Ward wardSelected, List<Medica
...
+	 *
+	 * @param medWard - the MedicalWard to delete. 
+	 * @throws OHServiceException.
+	 */
+	public void deleteMedicalWard(MedicalWard medWard) throws OHServiceException {
+		ioOperations.deleteMedicalWard(medWard);
+	}
+	
+	/**
+	 * Updates the  MedicalWard.
```
mwithi, line 252:
```suggestion
	 * Updates the specified {@link MedicalWard}.
```

src/main/java/org/isf/medicalstock/service/MovementIoOperationRepository.java
```diff
@@ -59,9 +59,13 @@ public interface MovementIoOperationRepository extends JpaRepository<Movement, I
 	@Query(value = "select max(mov.date) from Movement mov")
 	LocalDateTime findMaxDate();
 
-	@Query(value = "select mov.refNo from Movement mov where mov.refNo like :refNo")
-	List<String> findAllWhereRefNo(@Param("refNo") String refNo);
+	@Query(value = "select mov from Movement mov where mov.refNo = :refNo")
+	List<Movement> findAllWhereRefNo(@Param("refNo") String refNo);
```
mwithi, line 63: What this change is for?

mwithi (changes requested):
Added some requests, after that I will review Junit tests

sql/load_demo_data.sql
mwithi: Sorry, you don't need to update the autoincrement because with OP-1111 we removed also the table creation statements.

sql/step_04_all_following_steps.sql
```diff
@@ -92,4 +92,5 @@ source step_95_missing_user_related_fk.sql;
 source step_96_user_settings_table.sql;
 source step_97_fix_default_values.sql;
 source step_98_reset_permissions.sql;
-source step_99_telemetry.sql;
\ No newline at end of file
+source step_99_telemetry.sql;
+source step_aXXX_add_delete_btn_in_medicalstock.sql;
```
mwithi, line 96: remember to rename also the file script

src/main/java/org/isf/medicalstock/service/MovementIoOperationRepository.java
mwithi: Why you markred as resolved without applying suggestions or without saying why you won't apply it?

src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1194,6 +1195,16 @@ public void testMovementGetOrigin() throws Exception {
...
+	
+	@Test
+	public void testDeleteLastMovement() throws Exception {
+		int code = setupTestMovement(false);
+		Optional<Movement> movement = movementIoOperationRepository.findById(code); 
+		assertThat(movement).isPresent();
+		movBrowserManager.deleteLastMovement(movement.get());
+		Optional<Movement> movement2 = movementIoOperationRepository.findById(code); 
+		assertThat(movement2).isNotPresent();
+	}
```
mwithi, line 1207: I think we should test also the cases when the deletion is denied according with `deleteLastMovement()` checks

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
```diff
@@ -122,4 +152,53 @@ private void check(LocalDateTime from, LocalDateTime to, String errMsgKey) throw
...
+			String wardCode = ward.getCode();
+			MedicalWard medWard = movWardBrowserManager.getMedicalWardByWardAndMedical(wardCode, medicalCode, lot.getCode());
+			medWard.setIn_quantity(medWard.getIn_quantity() - quantity);
+			if (medWard.getIn_quantity() == 0 && medWard.getOut_quantity() == 0) {
+				movWardBrowserManager.deleteMedicalWard(medWard);
+			} else {
+				movWardBrowserManager.updateMedicalWard(medWard);
+			}
+			medical.setOutqty(medical.getOutqty() - quantity);
+			medicalsIoOperation.updateMedical(medical);
```
mwithi, line 199: Here we have to care also about `oh_medicaldsrstockmovward` that we totally forget. 

**To test the new feature please follow these steps:**

1. start from a fresh demo DB
2. add a number of charges and discharges in "Pharmaceuticals Stock"
3. delete all of them in reverse order (you should be obliged to do so)
4. export a new dump and compare it with `load_demo_data.sql` script

At step 4, only `*_LOCK`, `*_LAST_MODIFIED_DATE`, `*_LAST_MODIFIED_BY` and `oh_session_audit` table should differ; quantities and number of `LOT`s, `MEDICAL`s and `MEDICALWARD`s as well as `MOVEMENT`s and `MOVEMENTWARD`s should be unchanged.

**To test _denied_ deletions:**

1. start from a fresh demo DB
2. add a number of charges and discharges in "Pharmaceutical Stock"
3. add a number of discharges (either to Patients, Internal or Wards) in "Pharmaceuticals Stock Ward"
4. deletion should be **denied** and we will need also OP-1169 to restore the initial dump

src/main/java/org/isf/medicalstock/service/MovementIoOperationRepository.java
ArnaudFonzam: when I apply the changes the tests fails. i think it is about the limit clause.

src/test/java/org/isf/medicalstock/test/Tests.java
ArnaudFonzam: okay

src/main/java/org/isf/medicalstock/service/MovementIoOperationRepository.java
mwithi: the limit clause is in both statements, so what do you mean?

mwithi: ok, looks like "limit" is a vendor dependent function so it has not been implemented in JPA/Hiberanate, I didn't expected that.
There are workarounds for that, but let's leave it as is, thanks.

ArnaudFonzam: okay

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
ArnaudFonzam: hello @mwithi  I followed all the steps above and everything works fine.

ArnaudFonzam: hello @mwithi it is okay now please can you test it again if it is okay.

src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1194,6 +1197,23 @@ public void testMovementGetOrigin() throws Exception {
...
+	
+	@Test
+    public void testDeleteLastMovementDenied() throws Exception {
+		int code = setupTestMovement(false);
+		Optional<Movement> movement = movementIoOperationRepository.findById(code - 1); 
+        assertThrows(NoSuchElementException.class, () -> movBrowserManager.deleteLastMovement(movement.get()));
```
mwithi, line 1215: why you are testing the deletion of a non existing movement (code - 1) ?
You should try the deletion of an existing movmement that for some reason cannot be deleted anymore
Please check also the indentation

src/main/java/org/isf/medicalstock/manager/MovBrowserManager.java
mwithi: where is the part caring about `oh_medicaldsrstockmovward` as I was suggesting?

ArnaudFonzam: Hello 
[@Alessandro Domanico](https://openhospitalworkspace.slack.com/team/US2C2NPA9)
 after investigation of you suggestion I have made a remark that in the Income tab In pharmaceutical stock ward,  the values that we have  is that tab came from Movement not from MovementWard. When we do now the new movement in pharmaceutical stock ward we save now that in Movement Ward table. for exemple when we discharge something to central shop to a ward the MovementWard is not create so we can not delete the movement in MovementWard when the last movement is discharge. see the following picture. 
movWard

mwithi: you're right, the Income Tab shows the movement from main pharmacy to the specific ward, while the `oh_medicaldsrstockmovward` table is used only for the ward outgoing movements

we have to block the deletion of a discharge movement from main pharmacy to a ward if the same medical has been furtherly discharged to a patient/ward/internal

src/test/java/org/isf/medicalstock/test/Tests.java
```diff
@@ -1194,6 +1197,27 @@ public void testMovementGetOrigin() throws Exception {
...
+	@Test
+    public void testDeleteLastMovementDenied() throws Exception {
+		int code = setupTestMovement(false);
+		Optional<Movement> movement = movementIoOperationRepository.findById(code); 
+		assertThat(movement).isPresent();
+		Movement mov = movement.get();
+		MovementType movType = new MovementType("discharge","Discharge","-");
+		mov.setType(movType);
+		assertThrows(UndeclaredThrowableException.class, () -> movBrowserManager.deleteLastMovement(mov));
+    }
```
mwithi, line 1220: I don't get this test, we try to delete a movement after changing is MovementType? What this mean?
Please fix also indentation.

ArnaudFonzam: we denied to delete movement if the last movement is discharge type and the medical has been used in the Ward by patient, for internal usage or tranfert to other ward. so I am try to fix the param and test

mwithi: ok, let me know when I can check again, thanks!

ArnaudFonzam: @mwithi  I updated the method you can check if it is okay for you.

mwithi: I've widely tested and it works fine for me, please rename aXXX to the next availbel number and (in step_04 too) and we can merge.
Thanks a lot!

mwithi (approved):

PR #1244, opened 2024-01-17
OP-1166 Fix lot validation when automaticlot in false and automaticlot out true

mwithi:
See OP-1166.

The idea is to make everything more explicit for future reference in the code (and also avoid errors)

### DIFF
--- a/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
+++ b/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
@@ -37,7 +37,6 @@ import org.isf.utils.db.TranslateOHServiceException;
 import org.isf.utils.exception.OHDataValidationException;
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
-import org.isf.utils.exception.model.OHSeverityLevel;
 import org.isf.utils.time.TimeTools;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -114,58 +115,113 @@ protected void validateMovement(Movement movement, boolean checkReference) throw
...
 		}
 
 		if (!errors.isEmpty()) {
 			throw new OHDataValidationException(errors);
 		}
 	}
 
+	private void validateLot(List<OHExceptionMessage> errors, Lot lot, boolean checkCode) {
+		if (checkCode) {
+			if (lot.getCode().length() >= 50) {
```
dbmalkovsky, line 201: These two conditions can be combined into a single `if`

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -114,58 +115,113 @@ protected void validateMovement(Movement movement, boolean checkReference) throw
...
+				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.common.error.title"),
```
dbmalkovsky, line 202: why are you using the long version of this method?  The single parameter version sets the title and serverity.

It could be:

`errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.thelotidistoolongmax50chars.msg")));`

See line 156 for example.

dbmalkovsky:
It would be good to get some unit tests for these changes.  I am assuming the current tests don't cover all the paths.

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
mwithi: yes, thanks!

mwithi:
> It would be good to get some unit tests for these changes. I am assuming the current tests don't cover all the paths.

I've ran the tests, in theory the validation method is unchanged in the output (is producing the same result)

mwithi:
@dbmalkovsky or you mean with that that current tests could be better?

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
mwithi: yes, sorry, I was copying from an old code where `validateLot` method was still separated. Fixed. Thanks!

dbmalkovsky (approved):
LGTM

PR #1246, opened 2024-01-22
Chore: better explain docker usage

mwithi:


dbmalkovsky (approved):
LGTM

emecas (commented):
Yes, the readability has improved slightly.

However, I found the section on "How to run Open Hospital" to be confusing. it corresponds to the following block after the changes introduced here, which contains links to the GUI and UI. If you click the link to the GUI, it doesn't mention Docker at all. I assume the same is true for the UI (I didn't try it). On the other hand, if you go by the API (there is no link available), it covers much more about Docker steps.

mwithi:
> Yes, the readability has improved slightly.
> 
> However, I found the section on "How to run Open Hospital" to be confusing. it corresponds to the following block after the changes introduced here, which contains links to the GUI and UI. If you click the link to the GUI, it doesn't mention Docker at all. I assume the same is true for the UI (I didn't try it). On the other hand, if you go by the API (there is no link available), it covers much more about Docker steps.

Yes, I get your point of view. I think that if we want to emphasize the use of Docker we should have a common chapter in every README in order to make them consistent.

I will open an issue for that.

PR #1250, opened 2024-01-26
Chore: update maven.yml for JDK 17

dbmalkovsky:
Need to explicitly require the JDK and it doesn't hurt to specify the distribution.


mwithi (approved):
LGTM. Thanks a lot!

PR #1256, opened 2024-02-02
OP-1141 Default passwords suggestion more simple html plain text

mwithi:
See OP-1141.

### DIFF
--- a/src/test/java/org/isf/utils/file/TestFileTools.java
+++ b/src/test/java/org/isf/utils/file/TestFileTools.java
@@ -22,7 +22,12 @@
 package org.isf.utils.file;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
+import java.io.File;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dbmalkovsky (approved):
LGTM but it would be nice if there was a test or two for this new method

mwithi:
> LGTM but it would be nice if there was a test or two for this new method

Added tests, please check! Thanks!

mwithi:
Strange test are not working, in my local do... Going to investigate

src/test/java/org/isf/utils/file/TestFileTools.java
```diff
@@ -531,4 +536,39 @@ public void testHumanReadableByteCount() throws Exception {
...
+	public void testReadFileToStringLineByLineWithHtml() {
+		File file = getFile("testFile.txt");
+		String result = FileTools.readFileToStringLineByLine(file.getAbsolutePath(), true);
+
+		assertNotNull(result);
+		assertTrue(result.startsWith("<html>"));
+		assertTrue(result.endsWith("</html>"));
+		assertTrue(result.contains("<br>"));
+		assertTrue(result.contains("This could be one line text"));
+		assertTrue(result.contains("or text with"));
```
dbmalkovsky, line 550: Shouldn't this block really be:


assertThat(result).isNotNull();
assertThat(result).startsWith("\<html\>");
assertThat(result).endsWith("\</html\>");
assertThat(result).contains("\<br\>");
assertThat(result).contains("This could be one line text");
assertThat(result).contains("or text with");

Same thing for the next test too.

dbmalkovsky (commented):
Left a comment about updating the test to `assertThat()`

src/test/java/org/isf/utils/file/TestFileTools.java
mwithi: you're right

dbmalkovsky (approved):
LGTM

PR #1257, opened 2024-02-04
OP-1177 Add script to fix class name in menuitem

mwithi:
See OP-1177.


mwithi:
> Doesn't the demo data script need to be updated too?

Yes, thank you!

dbmalkovsky (dismissed):
LGTM

PR #1258, opened 2024-02-07
OP-1050 - Demo data update

mizzioisf:
See https://openhospital.atlassian.net/browse/OP-1050
Updated demo data 
- Added patients, examinations, visits, etc. 
- Added pricelist
- Minor fixes


sql/load_demo_data.sql
```diff
@@ -1,8 +1,8 @@
--- MariaDB dump 10.19  Distrib 10.6.13-MariaDB, for Win64 (AMD64)
+-- MariaDB dump 10.19  Distrib 10.6.16-MariaDB, for linux-systemd (x86_64)
 --
--- Host: localhost    Database: oh
+-- Host: localhost    Database: ohdemo
```
dbmalkovsky, line 3: Is this change really intended?

mwithi: actually since now we have only data (since OP-1111) we could remove this comment because pointless
nevertheless, I think it would be added automatically by mysqldump
it's just a comment anyway

mwithi:
rough error here: while `CODE` is progressive and grows indefinitely, `OPD NO.` resets every year, so 2021 is a bit out of place here, although still possible someone have inserted data in that way.

Also: note Patient 544 looks like has a double OPD the same date and time.

image

mwithi:
The OP-1150 is asking to provide more than one pricelist, while I can see only Basic here.
(hint: create a copy from Manage Lists)

Also: I would give a price for "Amount per day"

image

sql/load_demo_data.sql
mizzioisf: Yes, let's keep it. It's just a comment but it is better that we enforce the idea of separating real data from demo data.

mizzioisf:
> rough error here: while `CODE` is progressive and grows indefinitely, `OPD NO.` resets every year, so 2021 is a bit out of place here, although still possible someone have inserted data in that way.
> 
> Also: note Patient 544 looks like has a double OPD the same date and time.
> 
> image

This is fixed now. Still working on the data.

mwithi:
> This is fixed now. Still working on the data.

Could you resolve also conflicts?

mizzioisf:
Everything should be ok now, hopefully we can close this :-)

mwithi:
Superseded by https://github.com/informatici/openhospital-core/pull/1266

PR #1267, opened 2024-02-12
OP-1179 Change DailyRollingFileAppender to RollingFileAppender

mwithi:
See OP-1179.

### DIFF
--- a/src/main/java/org/isf/utils/log/LogUtil.java
+++ b/src/main/java/org/isf/utils/log/LogUtil.java
@@ -26,8 +26,8 @@ import java.io.File;
 import java.io.IOException;
 
 import org.apache.log4j.Appender;
-import org.apache.log4j.DailyRollingFileAppender;
 import org.apache.log4j.Logger;
+import org.apache.log4j.RollingFileAppender;
 
 public class LogUtil {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -43,8 +43,8 @@ public static String getLogFileAbsolutePath() {
 		Logger rootLogger = Logger.getRootLogger();
 		Appender appender = rootLogger.getAppender("RollingFile");
 
-		if (appender instanceof DailyRollingFileAppender) {
-			DailyRollingFileAppender rollingFileAppender = (DailyRollingFileAppender) appender;
+		if (appender instanceof RollingFileAppender) {
+			RollingFileAppender rollingFileAppender = (RollingFileAppender) appender;
```
dbmalkovsky, line 47:
```suggestion
		if (appender instanceof RollingFileAppender rollingFileAppender) {	
```
Lines 46 and 47 can be replaced with this single line.   This is called a "Pattern variable" that was introduced in Java 16.

src/main/java/org/isf/utils/log/LogUtil.java
```diff
@@ -43,8 +43,8 @@ public static String getLogFileAbsolutePath() {
...
 			String relativePath = rollingFileAppender.getFile();
 			String absolutePath = new File(relativePath).getAbsolutePath();
 			return absolutePath;
```
dbmalkovsky, line 50: I don't see the use in assigning the `new File()...` to a variable and then returning the variable.   Why not just return `new File()...`?

dbmalkovsky (commented):
Left some comments.....

src/main/java/org/isf/utils/log/LogUtil.java
mwithi: Just the result of many rework... done, thanks!

dbmalkovsky (commented):
LGTM; thanks

PR #1269, opened 2024-02-13
OP-312 Fix for daemon not yet started

mwithi:
See OP-312.

Fix NPE when TelemetryDaemon is not yet started.


dbmalkovsky (approved):
LGTM

PR #1279, opened 2024-02-27
OP-1190 | add migration file to add inventory ward submenu in pharmacy

JantBogard:
See OP-1190.


sql/step_a10X_add_inventory_ward_sub_menu_in_pharmacy.sql
mwithi, line 1: better to rename it to `step_a1XX_...`

PR #1293, opened 2024-03-15
OP-1205 Update README.md - Docker usage documentation in core

emecas:
See OP-1205

Improve building section to run tests against MariaDB Docker instance

https://openhospital.atlassian.net/browse/OP-1205

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,29 @@ To run the tests simply issue:
 
     mvn test
 
-Tests are ran against an in-memory database (H2). To run tests against a MariaDB instance, you can change [`src/test/resources/resources/database.properties`][database.prop] and run the Docker container in the root folder with:
+Note: tests are ran against an in-memory database (H2). 
+
+To run tests against a MariaDB instance:
+
+you can change [`src/test/resources/resources/database.properties`][database.prop] 
+
+From:
+
+    jdbc.class=org.h2.Driver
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,29 @@ To run the tests simply issue:
 
     mvn test
 
-Tests are ran against an in-memory database (H2). To run tests against a MariaDB instance, you can change [`src/test/resources/resources/database.properties`][database.prop] and run the Docker container in the root folder with:
+Note: tests are ran against an in-memory database (H2).
```
dbmalkovsky, line 21:
```suggestion
Note: tests are run against an in-memory database (H2). 
```

mwithi: I think it was ok "ran", isn't it?

emecas: > I think it was ok "ran", isn't it?

The correct one is:  Note: tests are run against an in-memory database (H2).

Here's why:

Run vs. Ran: "Run" is the present participle of the verb "to run," which is the correct tense for describing an ongoing action like testing. "Ran" is the past participle, used for completed actions.

"run" is the grammatically correct choice in this context.

@dbmalkovsky what do you think?

dbmalkovsky: I think "run" is the correct choice; besides it just sound better to a native English speaker 😏

mwithi: I agree now, so it was wrong before 😆

PR #1299, opened 2024-03-25
OP-1184 | Fixes

ArnaudFonzam:
See OP-1184

PR #1300, opened 2024-03-25
OP-1199 OP-1200 | Fix after (#1299)

ArnaudFonzam:
See OP-1199 OP-1200.


dbmalkovsky:
In my testing of the new SpringBoot I have found that the `MedicalInventory` and `MedicalInventoryRow` objects are missing the public methods to  `setLock()` and `getLock()`.

In addition there is no reason to define the `lock` variable as an object; just use the primitive `int`.

dbmalkovsky:
In `MedicalInventoryIoOperation.java`; lines 49-53, the parameter name is `medicalinventory` but in the javadoc it is `medicalInventory` (camel cased).   The javadoc version is more appropriate IMHO.

dbmalkovsky:
In `MedicalInventoryManager.java` there is a similar problem to `MedicalInventoryIoOperation.java`; for the same range of lines. `medicalInventory` is the preferred version.

dbmalkovsky:
In `MedicalInventoryRowIoOperation.java` lines 44-48 have the same case error. `medicalInventoryRow` should be used.

dbmalkovsky:
In `MedicalInventoryRowManager.java` lines 41-45 have the same variable case error. `medicalInventoryRow` should be used.

src/main/java/org/isf/medicalinventory/model/MedicalInventoryRow.java
```diff
@@ -156,6 +156,15 @@ public void setNewLot(boolean isNewLot) {
...
+		return lock;
+	}
+
+	
+	public void setLock(int lock) {
+		this.lock = lock;
+	}
 
 	public String getSearchString() {
 		StringBuffer sbNameCode = new StringBuffer();
```
dbmalkovsky, line 170: Why in this method is `getMedical() != null` tested multiple times?  Why not test it once and if null return an empty string.

dbmalkovsky (commented):
Added comments

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -418,4 +425,81 @@ private List<Movement> prepareDishargingMovement(Movement movement, boolean chec
...
+	
+	/**
+	 * Prepare the insert of the specified {@link Movement}
+	 *
+	 * @param movement - the movement to store.
+	 * @return {@code true} if is already used,{@code false} otherwise.
+	 * @throws OHServiceException.
+	 */
+	@Transactional(rollbackFor = OHServiceException.class)
+	public boolean prepareDishargingOrChargeMovementInventory(Movement movement) throws OHServiceException {
```
mwithi, line 437: "prepare" movement should be something to prepare an object to be persisted afterward, not to check and return a boolean, and the javadoc seems not reflecting the behaviour either (`true` if is already used?)

please what is the logic behind and what are you trying to achieve?

src/main/java/org/isf/medicalinventory/service/MedicalInventoryIoOperationRepository.java
```diff
@@ -49,12 +49,12 @@ public interface MedicalInventoryIoOperationRepository extends JpaRepository<Med
...
+	List<MedicalInventory> findInventoryByParamsWithoutStatus(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, @Param("type") String type);
 	
 	@Query(value = "select medinv from MedicalInventory medinv where medinv.inventoryDate >= :dateFrom and medinv.inventoryDate < :dateTo and medinv.status = :status and medinv.inventoryType = :type")
 	Page<MedicalInventory> findInventoryByParamsPageable(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, @Param("status") String status, @Param("type") String type, Pageable pageable);
 	
 	@Query(value = "select medinv from MedicalInventory medinv where medinv.inventoryDate >= :dateFrom and medinv.inventoryDate < :dateTo and medinv.inventoryType = :type")
-	Page<MedicalInventory> findInventoryByParamsPageablewithoutSatus(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, @Param("type") String type, Pageable pageable);
+	Page<MedicalInventory> findInventoryByParamsWithoutStatusPageable(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, @Param("type") String type, Pageable pageable);
```
mwithi, line 58: Anyway, why do we need to specify `WithoutSomething`? Would be better to just use a positive form like `BetweenDate` instead of a generic `ByParams`?

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
ArnaudFonzam: The logic  behind this method is:

- to store the new movement
- to update in_quantity of Medical or MedicalWard if the movement is charge or update out_quantityof Medical or MedicalWard if the movement is discharge.

src/main/java/org/isf/medicalinventory/service/MedicalInventoryIoOperationRepository.java
ArnaudFonzam: okay let me update it

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -418,4 +425,81 @@ private List<Movement> prepareDishargingMovement(Movement movement, boolean chec
...
+			if (stockQuantityUpdated) {
+				return true;
+			}
+		}
+		return false;
+	}
+	
+	private boolean updateStockQuantity(Movement movement) throws OHServiceException {
+		String movType = movement.getType().getType();
+		if (movType.equals("+")) {
```
mwithi, line 454: we found that is safer to use `.contains` method instead of `.equals`

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
mwithi: > to store the new movement

that's why I'm saying, why to call it "prepare"? Moreover a "store" operation, like save or persist or update, should return the stored object as from JPA specs, not a boolean. 

Anyway why it is called prepare**DishargingOrCharge**MovementInventory while inside we only do `ioOperations.prepareDischargingMovement(movement)`?

And the `updateStockQuantity` and `updateMedicalWardQuantity` look copied and paste from the `MedicalStockIoOperations`, why this code duplication?

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -418,4 +418,17 @@ private List<Movement> prepareDishargingMovement(Movement movement, boolean chec
...
+	/**
+	 * Save the specified {@link Movement}
+	 *
+	 * @param movement - the movement to store.
+	 * @return the stored {@link Movement} object.
+	 * @throws OHServiceException.
+	 */
+	@Transactional(rollbackFor = OHServiceException.class)
+	public Movement storeMovement(Movement movement) throws OHServiceException {
+		return ioOperations.newMovement(movement);
```
mwithi, line 431: bypassing `validateMovement`?

mwithi: Also, maybe newMovement in org/isf/medicalstock/service/MedicalStockIoOperations.java should be protected (and not public) to avoid this direct call from other packages

ArnaudFonzam: I have remove the method

src/main/java/org/isf/medicalinventory/service/MedicalInventoryRowIoOperation.java
```diff
@@ -45,8 +45,8 @@ public class MedicalInventoryRowIoOperation {
 	 * @return the newly persisted {@link MedicalInventoryRow} object.
 	 * @throws OHServiceException
 	 */
-	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalinventoryRow) throws OHServiceException {
-		return repository.save(medicalinventoryRow);
+	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalInventoryRow ) throws OHServiceException {
+		return repository.save(medicalInventoryRow );
```
mwithi, line 49:
```suggestion
		return repository.save(medicalInventoryRow);
```

src/main/java/org/isf/medicalinventory/model/MedicalInventoryRow.java
```diff
@@ -156,11 +156,23 @@ public void setNewLot(boolean isNewLot) {
 	public void setRealQty(double realQty) {
 		this.realQty = realQty;
 	}
+	
+	public int getLock() {
+		return lock;
+	}
+
+	
+	public void setLock(int lock) {
```
mwithi, line 165:
```suggestion
	
	public void setLock(int lock) {
```

src/main/java/org/isf/medicalinventory/model/MedicalInventory.java
```diff
@@ -148,4 +148,14 @@ public String getWard() {
...
+
+	
+	public int getLock() {
+		return lock;
+	}
+
+	
+	public void setLock(int lock) {
+		this.lock = lock;
+	}
```
mwithi, line 160:
```suggestion
	
	public int getLock() {
		return lock;
	}
	
	public void setLock(int lock) {
		this.lock = lock;
	}
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryRowManager.java
```diff
@@ -42,8 +42,8 @@ public class MedicalInventoryRowManager {
 	 * @return the newly persisted {@link MedicalInventoryRow} object.
 	 * @throws OHServiceException
 	 */
-	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalinventoryRow) throws OHServiceException {
-		return iOoperation.newMedicalInventoryRow(medicalinventoryRow);
+	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalInventoryRow ) throws OHServiceException {
+		return iOoperation.newMedicalInventoryRow(medicalInventoryRow );
```
mwithi, line 46:
```suggestion
	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalInventoryRow) throws OHServiceException {
		return iOoperation.newMedicalInventoryRow(medicalInventoryRow);
```

src/main/java/org/isf/medicalinventory/service/MedicalInventoryRowIoOperation.java
```diff
@@ -45,8 +45,8 @@ public class MedicalInventoryRowIoOperation {
 	 * @return the newly persisted {@link MedicalInventoryRow} object.
 	 * @throws OHServiceException
 	 */
-	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalinventoryRow) throws OHServiceException {
-		return repository.save(medicalinventoryRow);
+	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalInventoryRow ) throws OHServiceException {
```
mwithi, line 48:
```suggestion
	public MedicalInventoryRow newMedicalInventoryRow(MedicalInventoryRow medicalInventoryRow) throws OHServiceException {
```

PR #1303, opened 2024-03-28
OP-1148: upgrade to Spring Boot 3.2.4

dbmalkovsky:
See OP-1148

Still need to work on the GUI side


.mvn/wrapper/maven-wrapper.properties
mwithi, line 1: shouln't .properties be excluded from license plugin?

pom.xml
```diff
@@ -10,43 +10,161 @@
...
+
 	<build>
 		<plugins>
 			<plugin>
 				<artifactId>maven-compiler-plugin</artifactId>
-				<version>3.12.1</version>
-				<configuration>
-					<release>${java.version}</release>
-				</configuration>
+				<version>3.11.0</version>
```
mwithi, line 159: was this supposed to be downgraded?

pom.xml
```diff
@@ -10,43 +10,161 @@
...
-			  <configuration>
-			    <redirectTestOutputToFile>true</redirectTestOutputToFile>
-			    <reportsDirectory>${project.build.directory}/test-reports</reportsDirectory>
-			  </configuration>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>3.2.2</version>
+				<configuration>
+					<redirectTestOutputToFile>true</redirectTestOutputToFile>
+					<reportsDirectory>${project.build.directory}/test-reports</reportsDirectory>
+				</configuration>
```
mwithi, line 167: was this supposed to be downgraded?

pom.xml
```diff
@@ -160,7 +277,9 @@
 						<project.inceptionYear>2006</project.inceptionYear>
 						<project.name>Open Hospital</project.name>
 						<project.url>www.open-hospital.org</project.url>
-						<project.description>Open Hospital is a free and open source software for healthcare data management.</project.description>
+						<project.description>Open Hospital is a free and open source software for healthcare data
+							management.
+						</project.description>
```
mwithi, line 282: maybe unwanted newline

src/main/java/org/isf/medicalinventory/model/MedicalInventory.java
```diff
@@ -148,4 +148,12 @@ public String getWard() {
...
 		this.ward = ward;
 	}
+
+	public int getLock() {
+		return lock;
+	}
+
+	public void setLock(int lock) {
+		this.lock = lock;
+	}
```
mwithi, line 158: This should be already landed on core, right?

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
```diff
@@ -46,7 +46,7 @@ public interface PatientIoOperationRepository extends JpaRepository<Patient, Int
 	@Query("select p from Patient p where p.code = :id and (p.deleted = :deletedStatus or p.deleted is null)")
 	List<Patient> findAllWhereIdAndDeleted(@Param("id") Integer id, @Param("deletedStatus") char deletedStatus);
 
-	@Modifying
+	@Modifying(clearAutomatically = true)
```
mwithi, line 49: Why this was necessary here?

src/main/java/org/isf/menu/service/UserMenuItemIoOperationRepository.java
```diff
@@ -36,7 +36,7 @@ public interface UserMenuItemIoOperationRepository extends JpaRepository<UserMen
 			"menuItem.mySubmenu, menuItem.myClass, menuItem.isASubMenu, menuItem.position, groupMenu.active " +
 			"from UserMenuItem menuItem, GroupMenu groupMenu, UserGroup  userGroup, User user " +
 			"where (user.userName=:userId) " +
-			"and (user.userGroupName=userGroup.code) " +
+			"and (user.userGroupName.code=userGroup.code) " +
```
mwithi, line 39: wondering how it was working before :/

src/main/java/org/isf/telemetry/envdatacollector/collectors/SoftwareDataCollector.java
```diff
@@ -78,8 +78,8 @@ public Map<String, String> retrieveData() throws OHException {
 			result.put(CollectorsConstants.OS_BITNESS, String.valueOf(os.getBitness()));
 			result.put(CollectorsConstants.OS_CODENAME, os.getVersionInfo().getCodeName());
 
-			SessionImplementor sessionImp = (SessionImplementor) em.getDelegate();
-			DatabaseMetaData dbmd = sessionImp.connection().getMetaData();
+			Connection connection = em.unwrap(Connection.class);
+			DatabaseMetaData dbmd = connection.getMetaData();
```
mwithi, line 82: Why was this change needed?

src/main/java/org/isf/utils/db/DbJpaUtil.java
mwithi, line 1: This class is still needed because of openhospital-gui/src/main/java/org/isf/lab/gui/LabBrowser.java

I think we will need an issue to remove it and to find a better way

src/main/resources/database.properties.sample
mwithi, line 1: this was needed as sample in case one wants to configure an application using OH-core.jar as lib

src/test/resources/version.properties
```diff
@@ -1,4 +1,4 @@
 # This file is not meant to be modified by the user
 VER_MAJOR=1
-VER_MINOR=13
+VER_MINOR=14
 VER_RELEASE=0
```
mwithi, line 4:
```suggestion
VER_RELEASE=1
```

src/main/java/org/isf/medicalinventory/model/MedicalInventory.java
dbmalkovsky: No; this is part of https://github.com/informatici/openhospital-core/pull/1300

src/main/java/org/isf/menu/service/UserMenuItemIoOperationRepository.java
dbmalkovsky: Same question I had; there are others like this too (including some of the tests worked when they shouldn't have)

src/main/java/org/isf/sms/manager/SmsManager.java
```diff
@@ -102,6 +102,7 @@ else if (textLenght > MAX_LENGHT && split) {
 			for (String part : parts) {
 				Sms sms = new Sms();
 				sms.setSmsNumber(number);
+				sms.setSmsDate(smsToSend.getSmsDate());
```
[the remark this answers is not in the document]
dbmalkovsky: The date has to be set because the object marks the field as NON-NULL in the column definition.   This is one of the cases that I mentioned that the newer spring boot and its dependencies are more strict.

src/main/java/org/isf/telemetry/envdatacollector/collectors/SoftwareDataCollector.java
dbmalkovsky: Because `getDelete()` no longer exists.

src/test/resources/version.properties
dbmalkovsky: Done

.mvn/wrapper/maven-wrapper.properties
dbmalkovsky: It is; see line 267 of pom.xml.  The license in this file is the Apache license which comes with the file.  See for example,
https://github.com/jhipster/jhipster-sample-app/blob/main/.mvn/wrapper/maven-wrapper.properties

pom.xml
dbmalkovsky: No; fixed

pom.xml
dbmalkovsky: yes; fixed

src/main/java/org/isf/utils/db/DbJpaUtil.java
dbmalkovsky: I don't see that class used any place in the GUI repo.

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
dbmalkovsky: Because without that 4 tests fail.

In TestMergePatient.java:

- testMgrMergePatientPatient1MissingInformation()
- testMgrMergePatient()
- testMergePatientHistory()

And in patient/Tests.java:

- testMergePatientHistory()

src/main/resources/database.properties.sample
dbmalkovsky: OK; it is put back

src/main/java/org/isf/utils/db/DbJpaUtil.java
mwithi: The chain is DbJpaUtil <- DbSingleJpaConn <- PrintLabels <- LabBrowser

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
mwithi: I see... actually the field has a default value `SMS_DATE TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP`, it would be nice to ask Spring authors how to leverage SQL defaults otherwise are they useless?

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
mwithi: I imagined that, but why they fail? maybe is there an underlying design problem?

.mvn/wrapper/maven-wrapper.properties
mwithi: oh, alright, sorry

src/main/java/org/isf/menu/service/UserMenuItemIoOperationRepository.java
dbmalkovsky: No; those lines are correct as they are joining on `string`s.   Line 39 was trying to join on the variable `userGroupName which is an object  of type `UserGroup` with a `string` code value in the `userGroup` object.   Not sure how it was ever working but then the newer Hibernate and Spring JPA seem to be more strict.

src/main/resources/application.yml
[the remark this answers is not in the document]
dbmalkovsky: Application context is not part of Spring Boot 3.  Rather application.properties or application.yml are used and the rest of the information is found by reflected.

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
dbmalkovsky: Maybe; it was low on my list of things to investigate deeply.  I just wanted to get the tests to work :smirk:

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
dbmalkovsky: It might not be Spring; it may be Hibernate.  I am just not sure.

src/main/java/org/isf/utils/db/DbJpaUtil.java
dbmalkovsky: The class still exists; I just removed methods that were not being used any longer.

dbmalkovsky:
Note:  the changing of the `GenerationType` was required as the code was no choosing SEQUENCE instead of IDENTITY.  See https://blog.stackademic.com/avoid-using-generationtype-auto-strategy-for-id-generation-in-spring-boot-a67c6239eb7d for discussion of why using AUTO is not recommended.

dbmalkovsky:
Note: in `UserMenuItem` the `@Convert` annotation was added because this is the only boolean column in the database that is encoded with `Y` and `N` instead of the standard `0` and `1` values.   See https://www.baeldung.com/java-hibernate-6-boolean-converters for a discussion about this change.

pom.xml
```diff
@@ -10,43 +10,196 @@
...
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>3.2.4</version>
+		<relativePath/> <!-- lookup parent from repository -->
+	</parent>
+
 	<groupId>org.isf</groupId>
 	<artifactId>OH-core</artifactId>
+	<packaging>jar</packaging>
 	<version>1.14.0-SNAPSHOT</version>
```
mwithi, line 23: strange, this is supposed to be 1.14.1-SNAPSHOT now

src/main/resources/application.yml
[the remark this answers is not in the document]
mwithi: I see

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
mwithi: good point. what about using annotations like `@Column(nullable = false, columnDefinition = "TIMESTAMP DEFAULT CURRENT_TIMESTAMP")` ?

pom.xml
dbmalkovsky: done

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
mwithi: I'm trying to understand why, despite all those tests are using the same `updateDeleted` method, they are not isolated each other and sharing _apparently_ the same persistence context while, for instance, `testMgrMergePatientMergeNotes` seems working regardless the new annotation. Are the failing methods calling each other in a way I'm not able to see?

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
dbmalkovsky: I have found another solution.   The  "problem" is that we have an existing database; those annotations are used when generating or regenerating tables.   The suggested change isn't pretty but it is the best one I could find on the web that works with our setup.

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
dbmalkovsky: See https://www.baeldung.com/spring-data-jpa-modifying-annotation and https://stackoverflow.com/questions/43665090/why-do-we-have-to-use-modifying-annotation-for-queries-in-data-jpa for the reasons.   I think we should change it to `@Modifying(clearAutomatically=true, flushAutomatically=true)`

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
mwithi: I remember this kind of solution, it's ok for me, tested?

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
mwithi: it looks like we are facing this "problem" now because of the introduction of `org.springframework.boot.test.context.SpringBootTest` in the new `OHCoreTestCase.java`, which anyway might be a symptom that tests are not correctly written causing uncommited transactions or uncompleted rollbacks (I guess the latter). If the uncommitted transactions and uncompleted rollbacks are exactly what we are testing, then the "problem" is expected and using clearAutomatically and flushAutomatically sounds reasonable, otherwise we should understande the root cause of the fails.

src/main/java/org/isf/sms/manager/SmsManager.java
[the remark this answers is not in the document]
dbmalkovsky: yes; the tests all pass

src/main/java/org/isf/patient/service/PatientIoOperationRepository.java
dbmalkovsky: IHMO, all the tests need to be reviewed.   Some of them are confusing and convoluted at the very least.   I think we can do that when I rename the test files (discussed previously) and create a new "draft" PR for  a newer and more complete testing structure.

mwithi (approved):
LGTM. Thanks a lot!

PR #1313, opened 2024-04-18
OP-1210 Use Constructor injection for packages starting with A through D.

dbmalkovsky:
See OP-1210

All unit tests pass.


mwithi (approved):
LGTM. Thanks a lot!

PR #1318, opened 2024-04-20
OP-1213 Use Constructor injection for Core packages starting with O and P

dbmalkovsky:
See OP-1209

### DIFF
--- a/src/main/java/org/isf/patvac/service/VaccinePatientMergedEventListener.java
+++ b/src/main/java/org/isf/patvac/service/VaccinePatientMergedEventListener.java
@@ -25,15 +25,18 @@ import java.util.List;
 
 import org.isf.patient.model.PatientMergedEvent;
 import org.isf.patvac.model.PatientVaccine;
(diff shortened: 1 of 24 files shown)


src/main/java/org/isf/patvac/service/VaccinePatientMergedEventListener.java
```diff
@@ -25,16 +25,19 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.event.EventListener;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
 @Component
 public class VaccinePatientMergedEventListener {
-	@Autowired
+
 	PatVacIoOperations patVacIoOperations;
```
mwithi, line 35: should this be private like others?

dbmalkovsky: yes; very good catch;.  Change made.

mwithi (approved):
LGTM. Thanks a lot!

PR #1325, opened 2024-05-02
OP-1185 | Update Inventory

ArnaudFonzam:
See OP-1185.


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -62,7 +62,7 @@ public MovStockInsertingManager(MedicalStockIoOperations medicalStockIoOperation
 	 * @param checkReference - if {@code true} it will use {@link #checkReferenceNumber(String) checkReferenceNumber}
 	 * @throws OHServiceException
 	 */
-	protected void validateMovement(Movement movement, boolean checkReference) throws OHServiceException {
+	protected void validateMovement(Movement movement, boolean checkReference, boolean checkWard) throws OHServiceException {
```
mwithi, line 65: this change will affect also API... why do you need to exclude the check on the Ward field?

mwithi:
Non needed anymore

PR #1327, opened 2024-05-09
OP-1239 re-enable logging to a physical file

dbmalkovsky:
See OP-1239


mwithi (approved):
Tested and it works fine, thanks!

PR #1328, opened 2024-05-09
OP-1219 Add Ward comparator to sort in combobox

mwithi:
See OP-1219.

### DIFF
--- a/src/main/java/org/isf/ward/model/Ward.java
+++ b/src/main/java/org/isf/ward/model/Ward.java
@@ -21,6 +21,8 @@
  */
 package org.isf.ward.model;
 
+import java.util.Comparator;
+
 import javax.persistence.AttributeOverride;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -35,7 +37,7 @@ import org.isf.utils.db.Auditable;
 import org.springframework.data.jpa.domain.support.AuditingEntityListener;
 
 @Entity
-@Table(name="OH_WARD")
+@Table(name = "OH_WARD")
 @EntityListeners(AuditingEntityListener.class)
 @AttributeOverride(name = "createdBy", column = @Column(name = "WRD_CREATED_BY", updatable = false))
 @AttributeOverride(name = "createdDate", column = @Column(name = "WRD_CREATED_DATE", updatable = false))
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/ward/model/Ward.java
```diff
@@ -314,7 +318,26 @@ public int hashCode() {
...
-	
 
+	public static class WardCodeComparator implements Comparator<Ward> {
+
+		@Override
+		public int compare(Ward object1, Ward object2) {
+			if (object2.getCode() == null) {
+				return -1;
+			}
+			return object1.getCode().compareTo(object2.getCode());
```
dbmalkovsky, line 329:
```suggestion
			return object2.getCode().compareTo(object1.getCode());
```
Because what if object1.getCode() returns null?

src/main/java/org/isf/ward/model/Ward.java
```diff
@@ -314,7 +318,26 @@ public int hashCode() {
...
+	}
+
+	public static class WardDescriptionComparator implements Comparator<Ward> {
+
+		@Override
+		public int compare(Ward object1, Ward object2) {
+			if (object2.getDescription() == null) {
+				return -1;
+			}
+			return object1.getDescription().compareTo(object2.getDescription());
```
dbmalkovsky, line 340:
```suggestion
			return object2.getDescription().compareTo(object1.getDescription());
```
Same comment as above

dbmalkovsky (approved):
LGTM; given my suggestions

src/main/java/org/isf/ward/model/Ward.java
mwithi: Actually the way I use it in the GUI (`wardsList.sort(new Ward.WardDescriptionComparator())`) doesn't require the null check, but if we want to check, then we should check both objects for null, wdyt?

The order in the `compareTo`, instead, will affect the result, that will be in alphabetical order or reversed, so I would not change that.

mwithi: Or we can just use `StringUtils.compare(object1.getCode(), object2.getCode())` (from Apache Commons) that is null safe? (cfr https://www.baeldung.com/java-compare-strings#comparison-apache)

mwithi: That anyway internally does:

```
    public static int compare(final String str1, final String str2, final boolean nullIsLess) {
        if (str1 == str2) { // NOSONARLINT this intentionally uses == to allow for both null
            return 0;
        }
        if (str1 == null) {
            return nullIsLess ? -1 : 1;
        }
        if (str2 == null) {
            return nullIsLess ? 1 : - 1;
        }
        return str1.compareTo(str2);
    }
```

mwithi: Check the new version please

dbmalkovsky (approved):
LGTM

PR #1336, opened 2024-05-23
OP-1218 Enable products medicals soft deletion

mwithi:
See OP-1218.


src/main/java/org/isf/medicals/model/Medical.java
```diff
@@ -21,6 +21,10 @@
  */
 package org.isf.medicals.model;
 
+import org.isf.medtype.model.MedicalType;
+import org.isf.utils.db.Auditable;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
```
dbmalkovsky: The import order defined in both Eclipse and IntelliJ has `jakarta` coming before `org`

dbmalkovsky (commented):
Left a comment about import order; see the .ide-settings for Eclipse and IntelliJ that both have them different than what is in Medical.java

src/main/java/org/isf/medicals/model/Medical.java
mwithi: fixed, thanks!

dbmalkovsky (approved):
LGTM

PR #1338, opened 2024-05-31
OP-1217 Add Medical Stock Feature

mwithi:
See OP-1217.


src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,136 @@
...
+
+@Entity
+@Table(name = "OH_MEDICALDSRSTOCK")
+@EntityListeners(AuditingEntityListener.class)
+@AttributeOverride(name = "createdBy", column = @Column(name = "MS_CREATED_BY", updatable = false))
+@AttributeOverride(name = "createdDate", column = @Column(name = "MS_CREATED_DATE", updatable = false))
+@AttributeOverride(name = "lastModifiedBy", column = @Column(name = "MS_LAST_MODIFIED_BY"))
+@AttributeOverride(name = "lastModifiedDate", column = @Column(name = "MS_LAST_MODIFIED_DATE"))
+@AttributeOverride(name = "active", column = @Column(name = "MS_ACTIVE"))
+public class MedicalStock {
```
dbmalkovsky:
```suggestion
public class MedicalStock extends Auditable<String> {
```
Look at `Lot` and `Movement` in the same package

src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,136 @@
...
+
+	public MedicalStock() {
+		super();
```
dbmalkovsky: Not needed

src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	public MedicalStock(@NotNull Medical medical, @NotNull LocalDate balanceDate, @NotNull int balance, LocalDate nextMovDate, Integer days) {
+		super();
```
dbmalkovsky: Not needed

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -359,6 +370,57 @@ protected Medical updateMedicalOutcomingQuantity(int medicalCode, double increme
 		return medicalRepository.save(medical);
 	}
 
+	/**
+	 * Updates the medical stock balance for the specified medical at the specified date.
+	 * 
+	 * If the date is present in the table, the balance is updated If the date is not present, a new balance is inserted for the date and the previous one is
```
dbmalkovsky:
```suggestion
	 * If the date is present in the table, the balance is updated. If the date is not present, a new balance is inserted for the date and the previous one is
```

src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,135 @@
...
+
+	@NotNull
+	@Column(name = "MS_BALANCE")
+	private int balance;
+
+	@Column(name = "MS_DATE_NEXT_MOV")
+	private LocalDate nextMovDate;
+
+	@Column(name = "MS_DAYS")
+	private Integer days;
```
dbmalkovsky, line 85: Why isn't this just `int` like line 79 above?

src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,135 @@
...
+	public LocalDate getNextMovDate() {
+		return nextMovDate;
+	}
+
+	public void setNextMovDate(LocalDate nextMovDate) {
+		this.nextMovDate = nextMovDate;
+	}
+
+	public Integer getDays() {
+		return days;
```
dbmalkovsky, line 128: if the type changes then this needs changing too.

src/main/java/org/isf/medicalstock/model/MedicalStock.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	public void setDays(Integer days) {
```
dbmalkovsky, line 131: if the type changes then this needs changing too.

src/test/java/org/isf/medicalstock/Tests.java
```diff
@@ -1372,6 +1392,101 @@ void testDeleteLastMovementWithPriorMovements(boolean in, boolean out, boolean t
 		assertThat(followingMovement2).isNotPresent();
 	}
 
+	@Test
```
dbmalkovsky: Is there a reason these tests are not set up to be `@ParmeterizedTest`s?  Even if there should be no interaction running them that way would prove it, wouldn't it?

dbmalkovsky (commented):
I left a few comments after a quick review.

src/main/java/org/isf/medicalstock/model/MedicalStock.java
mwithi: because it must be nullable

src/test/java/org/isf/medicalstock/Tests.java
mwithi: yes, I was in doubt about this indeed... let me add

dbmalkovsky (dismissed):
LGTM

My preference is when a `then block` ends with a `return` there is no need for an `else block`.  It reduces indentation.

See for example the `updateMedicalStockTable()` method in `MedicalStockIoOperations`

mwithi:
> LGTM
> 
> My preference is when a `then block` ends with a `return` there is no need for an `else block`. It reduces indentation.
> 
> See for example the `updateMedicalStockTable()` method in `MedicalStockIoOperations`

you mean this way? I think in this case it would reduce the readability making more difficult to understand what the logic is for

```
	private MedicalStock updateMedicalStockTable(Medical medical, LocalDate date, int incrementQuantity) throws OHServiceException {

		List<MedicalStock> medicalStockList = medicalStockRepository.findByMedicalCodeOrderByBalanceDateDesc(medical.getCode());
		MedicalStock medicalStock;

		if (medicalStockList.isEmpty() && incrementQuantity < 0) {
			throw new OHServiceException(
							new OHExceptionMessage("Medical '" + medical.getDescription() + "' (" + medical.getCode() + ") not found (not possible)."));
		}
		if (medicalStockList.isEmpty()) {
			// first insert
			medicalStock = new MedicalStock();
			medicalStock.setMedical(medical);
			medicalStock.setBalanceDate(date);
			medicalStock.setBalance(incrementQuantity); // balance = first increment
			return medicalStockRepository.save(medicalStock);

		}
		medicalStock = medicalStockList.get(0);
		if (TimeTools.isSameDay(date, medicalStock.getBalanceDate())) {
			// update
			int balance = medicalStock.getBalance();
			medicalStock.setBalance(balance + incrementQuantity);
			medicalStock.setBalanceDate(date);
			return medicalStockRepository.save(medicalStock);

		}
		// update previous
		medicalStock.setNextMovDate(date);
		medicalStock.setDays(TimeTools.getDaysBetweenDates(medicalStock.getBalanceDate(), date, true));
		medicalStockRepository.save(medicalStock);

		// insert new
		int newBalance = medicalStock.getBalance() + incrementQuantity;
		MedicalStock newMedicalStock = new MedicalStock();
		newMedicalStock.setMedical(medical);
		newMedicalStock.setBalanceDate(date);
		newMedicalStock.setBalance(newBalance);
		return medicalStockRepository.save(newMedicalStock);
	}
```

dbmalkovsky:
Yes

dbmalkovsky (approved):
LGTM

PR #1342, opened 2024-06-04
OP-1278 Testing Coverage Improvements for Admission module

dbmalkovsky:
See OP-1278


src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -502,6 +504,7 @@ else if (dateOut.isBefore(ad.getAdmDate()) || dateOut.equals(ad.getAdmDate())) {
 				// date control
 				if (admission.getDeliveryDate() == null) {
 					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.admission.controln1datenodeliverydatefound.msg")));
+					throw new OHDataValidationException(errors);
```
mwithi: Why in this case we throw the error directly?

src/test/java/org/isf/admission/Tests.java
```diff
@@ -288,6 +291,24 @@ void testIoGetAdmittionsByDatePageable(boolean maternityRestartInJune) throws Ex
...
+		String str = "2000-01-01T10:11:30";
+		String str2 = "2023-05-05T10:11:30";
+		LocalDateTime dateFrom = LocalDateTime.parse(str);
+		LocalDateTime dateTo = LocalDateTime.parse(str2);
+		Admission foundAdmission = admissionIoOperation.getAdmission(id);
+		// when:
+		List<Admission> patients = admissionIoOperation.getAdmissionsByAdmDate(dateFrom, dateTo);
+
+		// then:
+		assertThat(patients.get(0).getId()).isEqualTo(foundAdmission.getId());
```
mwithi:
```suggestion
		List<Admission> admissions = admissionIoOperation.getAdmissionsByAdmDate(dateFrom, dateTo);

		// then:
		assertThat(admissions.get(0).getId()).isEqualTo(foundAdmission.getId());
```

src/test/java/org/isf/admission/Tests.java
```diff
@@ -572,6 +601,16 @@ void testIoDeletePatientPhoto(boolean maternityRestartInJune) throws Exception {
 		assertThat(updatedPatient.getPatientProfilePhoto().getPhoto()).isNull();
 	}
 
+	@ParameterizedTest(name = "Test with MATERNITYRESTARTINJUNE={0}")
+	@MethodSource("maternityRestartInJune")
+	@Transactional
+		// requires active session because of lazy loading of patient photo
```
mwithi:
```suggestion
	// requires active session because of lazy loading of patient photo
```

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: Because the persistence code in the following block of code requires delivery date to be non-null.

src/test/java/org/isf/admission/Tests.java
```diff
@@ -1078,6 +1141,193 @@ void testMgrDeletePatientPhoto(boolean maternityRestartInJune) throws Exception
...
+
+	@ParameterizedTest(name = "Test with MATERNITYRESTARTINJUNE={0}")
+	@MethodSource("maternityRestartInJune")
+	void testMgrValidateDeliveryDate(boolean maternityRestartInJune) throws Exception {
+		GeneralData.MATERNITYRESTARTINJUNE = maternityRestartInJune;
+		int id = setupTestAdmission(false, true);
+		Admission admission = admissionBrowserManager.getAdmission(id);
+		GeneralData.LANGUAGE = "en";
+
+		admission.setVisitDate(null);
```
mwithi: this should not be needed for this test

src/test/java/org/isf/admission/Tests.java
```diff
@@ -1244,6 +1493,102 @@ void testMgrValidateMaternity(boolean maternityRestartInJune) throws Exception {
...
+		int id = setupTestAdmission(false);
+		String str = "2000-01-01T10:11:30";
+		String str2 = "2023-05-05T10:11:30";
+		LocalDateTime dateFrom = LocalDateTime.parse(str);
+		LocalDateTime dateTo = LocalDateTime.parse(str2);
+		int page = 0;
+		int size = 10;
+		Admission foundAdmission = admissionIoOperation.getAdmission(id);
+		// when:
+		List<Admission> patients = admissionBrowserManager.getDischarges(dateFrom, dateTo, page, size);
```
mwithi: strange that `getDischarges` requires page and size like the pageable version... 🤔

src/test/java/org/isf/admission/Tests.java
```diff
@@ -1244,6 +1493,102 @@ void testMgrValidateMaternity(boolean maternityRestartInJune) throws Exception {
...
+		GeneralData.MATERNITYRESTARTINJUNE = maternityRestartInJune;
+		// given:
+		int id = setupTestAdmission(false);
+		String str = "2000-01-01T10:11:30";
+		String str2 = "2023-05-05T10:11:30";
+		LocalDateTime dateFrom = LocalDateTime.parse(str);
+		LocalDateTime dateTo = LocalDateTime.parse(str2);
+		Admission foundAdmission = admissionIoOperation.getAdmission(id);
+		// when:
+		List<Admission> patients = admissionBrowserManager.getAdmissionsByDate(dateFrom, dateTo);
```
mwithi: Maybe `getAdmissionsByDate` method name needs to be more expressive

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
mwithi: Sorry, I don't understand this.

Also in lines 314 and 320 should not be there the `throw new`, unless we want to stop immediately the validation at that point (makes sense).

Now that I think about it, it seems we miss a check for patient == null, is it?

src/test/java/org/isf/admission/Tests.java
dbmalkovsky: Yes it is required to force a different code path; I was trying to execute both branches of a test.

dbmalkovsky: So I believe this should stay.

src/test/java/org/isf/admission/Tests.java
mwithi: It seems also is not used in other components, maybe is an old version of the same method? In such case we can delete it and the related test as well

src/test/java/org/isf/admission/Tests.java
mwithi: ...like `getAdmissionsByAdmissionDate`

src/test/java/org/isf/admission/Tests.java
mwithi: Ah ok, but it doesn't change the test outcome, does it?

dbmalkovsky: no; it just increases line coverage 😄

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
mwithi: ok, got it, but so we could put it in a `else` block?

same for the following check, where `admission.getCtrlDate1() == null` could be replaced with `ctrl1Date`:

```
			LocalDateTime ctrl2Date = admission.getCtrlDate2();
			if (ctrl2Date != null) {
				if (admission.getCtrlDate1() == null) {
					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.admission.controldaten2controldaten1notfound.msg")));
				}
				// date control
				LocalDateTime limit;
				if (admission.getDisDate() == null) {
					limit = today;
				} else {
					limit = admission.getDisDate();
				}
				if (ctrl1Date != null && (ctrl2Date.isBefore(ctrl1Date) || ctrl2Date.isAfter(limit))) {
					errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.admission.pleaseinsertavalidcontroln2date.msg")));
				}
			}
```

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -308,6 +308,18 @@ public Patient deletePatientPhoto(int id) throws OHServiceException {
...
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.admission.admissionwardcannotbeempty.msg")));
+			throw new OHDataValidationException(errors);
+		}
+
+		LocalDateTime dateIn = admission.getAdmDate();
+		if (dateIn == null) {
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.admission.admissiondatecannotbeempty.msg")));
+			throw new OHDataValidationException(errors);
+		}
+
```
mwithi: Also in lines 314 and 320 should not be there the throw new, unless we want to stop immediately the validation at that point (makes sense).

Now that I think about it, it seems we miss a check for `patient == null` after line 326, is it?

src/test/java/org/isf/admission/Tests.java
mwithi: 😁

```suggestion
		admission.setVisitDate(null); // here just to increase code tests coverage
```

src/test/java/org/isf/admission/Tests.java
dbmalkovsky: `getDischarges()` is used in the API repo in `AdmissionController`.

See https://github.com/informatici/openhospital-api/blob/0dab980a096894145ebb682c3ea6e79cbca778bd/src/main/java/org/isf/admission/rest/AdmissionController.java#L270

So I've left it the way it was.

src/test/java/org/isf/admission/Tests.java
dbmalkovsky: Changing this ripples into the API repo code and requires a bunch of changes in CORE.  @mwithi  You still want to change it?

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: I have changes ready to add the `patient` check and rework the section slightly.   Waiting on responses to other comments before comitting.

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: I have this change ready once other issues are decided on.

src/test/java/org/isf/admission/Tests.java
mwithi: Yes, no problem, I think it's needed for clearer methods naming

src/test/java/org/isf/admission/Tests.java
mwithi: That endpoint is using the pageable version

https://github.com/informatici/openhospital-api/blob/0dab980a096894145ebb682c3ea6e79cbca778bd/src/main/java/org/isf/admission/rest/AdmissionController.java#L278

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: committed

src/test/java/org/isf/admission/Tests.java
dbmalkovsky: my bad; changes committed

src/test/java/org/isf/admission/Tests.java
dbmalkovsky: committed the CORE changes; will look at GUI and API next

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: code committed

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
```diff
@@ -166,22 +166,8 @@ public PagedResponse<Admission> getAdmissionsPageable(LocalDateTime dateFrom, Lo
...
-	 * @param dateTo
-	 * @param page
-	 * @param size
-	 * @return the list of completed Admissions (could be empty)
-	 * @throws OHServiceException
-	 */
-	public List<Admission> getDischarges(LocalDateTime dateFrom, LocalDateTime dateTo, int page, int size) throws OHServiceException {
-		return ioOperations.getAdmissionsByDischargeDate(dateFrom, dateTo, PageRequest.of(page, size));
+	public List<Admission> getAdmissionsByAdmissionDate(LocalDateTime dateFrom, LocalDateTime dateTo) throws OHServiceException {
+		return ioOperations.getAdmissionsByAdmissionDate(dateFrom, dateTo);
```
mwithi, line 155: Not sure here... is `getDischarges` disappeared now?

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
mwithi: My suggestion here is not to throw the exception on line 500 but rather using an `else` in line 501.
Same at line 517, where since the exception is not thrown the `ctrl1Date` could throw NPE in line 525

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: yes

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
dbmalkovsky: done

src/main/java/org/isf/admission/manager/AdmissionBrowserManager.java
mwithi: ok, right, we have the `getDischargesPageable` now, thanks!

mwithi:
Looking at the `AdmissionIoOperationRepository` only, it seems we have kind of duplications in methods, which differ in returning types only or having not used params:

```
@Query(value = "select a FROM Admission a WHERE a.admDate >= :dateFrom AND a.admDate <= :dateTo and a.deleted = 'N'")
List<Admission> findAllWhereAdmissionDate(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, Pageable pageable);

@Query(value = "select a FROM Admission a WHERE a.disDate >= :dateFrom AND a.disDate <= :dateTo and a.deleted = 'N'")
List<Admission> findAllWhereDischargeDate(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, Pageable pageable);

@Query(value = "select a FROM Admission a WHERE a.admDate >= :dateFrom AND a.admDate <= :dateTo and a.deleted = 'N'")
List<Admission> findAllWhereAdmissionDate(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo);

@Query(value = "select a FROM Admission a WHERE a.admDate >= :dateFrom AND a.admDate <= :dateTo and a.deleted = 'N'")
Page<Admission> findAllWhere_AdmissionDate_Paginated(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, Pageable pageable);

@Query(value = "select a FROM Admission a WHERE a.disDate >= :dateFrom AND a.disDate <= :dateTo and a.deleted = 'N'")
Page<Admission> findAllWhere_DischargeDate_Paginated(@Param("dateFrom") LocalDateTime dateFrom, @Param("dateTo") LocalDateTime dateTo, Pageable pageable);
```

By removing the `Pageable` parameter from the first two, the first and the third are equal and one can be removed.
After that, the effects on the `AdmissionIoOperations` are obvious, but I found also that:

- `getAdmissionsByAdmissionDatePages` is never used, only tested and maybe can be removed and so the `getAdmissions` in manager
- `getAdmissionsByDischargeDate` is never used, but it can stay for "completeness"

Wdyt?

dbmalkovsky:
I've made the changes you suggested; built the jar; rebuild and tested GUI and API.   Looks good.

mwithi (approved):
LGTM. Thanks a lot!

PR #1343, opened 2024-06-05
OP-107 FHIR service layer

Th0rst3nH:
OP-107 Created FHIR Message Layer


src/main/java/org/isf/fhir/FhirService.java
```diff
@@ -0,0 +1,42 @@
+package org.isf.fhir;
```
dbmalkovsky: Thanks for the PR; please add the standard license headers to each file.

mwithi:
Hi @Th0rst3nH thanks for the PR! Could you please give some documentation on how it is meant to be used in the rest of the application?

src/main/java/org/isf/fhir/FhirService.java
mwithi, line 1: Should this be under package `org.isf.fhir.service` ?

Th0rst3nH: Yes can do :)

Th0rst3nH:
Yes, this is basically  just a simple string json converter to the standard FHIR Message. What kind of documentation do you need? Please see the test how to use it.

src/test/java/org/isf/fhir/service/FhirPatientServiceTest.java
```diff
@@ -0,0 +1,35 @@
+package org.isf.fhir.service;
```
dbmalkovsky, line 1: Please add missing license header

src/test/java/org/isf/fhir/service/FhirPatientServiceTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
```
dbmalkovsky, line 3: Project standards are to enumerate all imports and not use package level wildcards; thanks

PR #1341, opened 2024-06-04
OP-1185 | Update MedicalInvenotyRow Add Column IS_NEW_LOT

ArnaudFonzam:
See OP-1185.

PR #1344, opened 2024-06-06
Fix: fix step_98 and add extra sql for fixing existing installations

mwithi:


dbmalkovsky (commented):
Doesn't line 1740 of load_demo_data.sql need to be changed also?

mwithi:
> Doesn't line 1740 of load_demo_data.sql need to be changed also?

yes but it is going to be done in #1341

dbmalkovsky (approved):
LGTM

PR #1349, opened 2024-06-15
OP-1239 found another instance that needed to be changed

dbmalkovsky:
See OP-1239

Missed the resources in the test subdirectory.


mwithi (approved):
LGTM. Thanks a lot!

PR #1351, opened 2024-06-17
Chore: Hibernate renamed org.hibernate.type to org.hibernate.orm.jdbc.bind

dbmalkovsky:
Found this when trying to output sql parameters into the log


mwithi (approved):
LGTM. Thanks a lot!

PR #1354, opened 2024-06-21
Chore: update to Spring Boot 3.2.7

dbmalkovsky:
Both GUI and API tested after the change without any obvious issues

This seemed like a safer change then the one suggested in https://github.com/informatici/openhospital-core/pull/1339


mwithi (approved):
LGTM. Thanks a lot!

PR #1368, opened 2024-07-12
Chore: remove unused files

dbmalkovsky:
When looking at coverage numbers I found these two files that appeared not to be used.

I removed them from CORE; rebuilt the jar, and then ran `mvn clean test` on both the GUI and API projects successfully.


mwithi (approved):
LGTM. Thanks a lot!

PR #1370, opened 2024-07-19
OP-1290: Testing Coverage Improvements for ServicePrinting module

dbmalkovsky:
See OP-1290


rpt_base/TestReport.jasper
mwithi, line 1: do you have also the corresponding `.jrxml` file?

dbmalkovsky: No; the test path doesn't need it.

mwithi: ok, but if one wants to inspect and/or modify it in the future has to start from scratch?

dbmalkovsky: Added the file.

mwithi (approved):
LGTM. Thanks a lot!

PR #1375, opened 2024-07-23
OP-1301-fix-units-packets-management-in-main-store

mwithi:
See OP-1301.

- [x] Add new exception for cases when ordered discharging quantity is greater than the one available in the main store for the specified medical
- [x] [Fix one test that by the name was supposed to fail indeed and the new introduced check brought it to surface


dbmalkovsky (dismissed):
LGTM (assuming the tests get fixed)

Just wondering if the Log statements should be left in for the system admin to help track down the problem.

mwithi:
> LGTM (assuming the tests get fixed)
> 
> Just wondering if the Log statements should be left in for the system admin to help track down the problem.

Yes I was wondering too, let me adjust

mwithi:
going to fix tests...

dbmalkovsky (dismissed):
LGTM

dbmalkovsky (approved):
LGTM

PR #1381, opened 2024-07-31
OP-1189 | validate inventory

ArnaudFonzam:
See OP-1189.


src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +234,47 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+	public boolean validateInventory(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList) throws OHServiceException {
+		String dischargeCode = inventory.getDischargeType();
+		String chargeCode = inventory.getChargeType();
+		Integer supplierId = inventory.getSupplier();
+		String wardCode = inventory.getDestination();
+		MovementType chargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(chargeCode);
+		MovementType dischargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(dischargeCode);
+		Supplier supplier = supplierManager.getByID(supplierId);
+		Ward ward = wardManager.findWard(wardCode);
+		LocalDateTime today = LocalDateTime.now();
```
dbmalkovsky: Why isn't this `TimeTools.getNow()` which handles seconds correctly?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +234,47 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+		for (Iterator<MedicalInventoryRow> iterator = inventoryRowSearchList.iterator(); iterator.hasNext();) {
+			String reference = "Ref-"+LocalDateTime.now();
```
dbmalkovsky, line 250: Is there any possibility of generating the same `reference` string if the code runs very, very fast?

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -495,6 +496,28 @@ void testMgrUpdateMedicalInventoryRow() throws Exception {
...
+		medicalsIoOperationRepository.saveAndFlush(medical);
+		Lot lot = testLot.setup(medical, false);
+		lotIoOperationRepository.saveAndFlush(lot);
+		MedicalInventoryRow medicalInventoryRow = testMedicalInventoryRow.setup(savedInventory, medical, lot, false);
+		int inventoryId = medicalInventoryRow.getId();
+		MedicalInventoryRow newMedicalInventoryRow = medicalInventoryRowManager.newMedicalInventoryRow(medicalInventoryRow);
+		assertThat(newMedicalInventoryRow).isNotNull();
+		List<MedicalInventoryRow> medicalInventoryRows = medicalInventoryRowManager.getMedicalInventoryRowByInventoryId(inventoryId);
+		assertThat(medicalInventoryRows).isNotEmpty();
+		assertThat(medicalInventoryRows.size()).isEqualTo(1);
```
dbmalkovsky:
```suggestion
		assertThat(medicalInventoryRows).hasSize(1);
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +234,47 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
 			throw new OHDataValidationException(errors);
 		}
 	}
+	
+	@Transactional(rollbackFor = OHServiceException.class)
+	public boolean validateInventory(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList) throws OHServiceException {
```
mwithi: this is a wrong pattern: validation method should not "doing anything" per se, but only returning a list of errors, that if it is empty means validation is ok. See other validation methods in other managers.

src/main/java/org/isf/medicalstock/service/MedicalStockIoOperations.java
```diff
@@ -296,7 +296,34 @@ public boolean lotExists(String lotCode) throws OHServiceException {
...
+			for (Object[] result : wardsTotalQuantities) {
+				lotCode = (String) result[0];
+				Double wardsTotalQuantity = (Double) result[1];
+
+				// Update the lot if found
+				lot.setWardsTotalQuantity(wardsTotalQuantity);
+			}
+			return lot;
+		}
+		return null;
```
mwithi: replicating the same code of `getLotsByMedical(Medical medical, boolean removeEmpty)` is a bad code design. Please be DRY.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +226,39 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+					String lotCodeOfMovement = mov.getLot().getCode();
+					Lot lot = movStockInsertingManager.getLot(lotCodeOfMovement);
+					double mainStoreQty = (double)lot.getMainStoreQuantity();
+					Integer medicalCode = medical.getCode();
+					Integer movMedicalCode = mov.getMedical().getCode();
+					if (movMedicalCode.equals(medicalCode) && lotCodeOfMovement.equals(lotCode)) {
+						if (mainStoreQty != theoQty) { 
+							medicalInventoryRow.setTheoreticQty(mainStoreQty);
+							medicalInventoryRow.setRealqty(mainStoreQty);
+							medicalInventoryRowManager.updateMedicalInventoryRow(medicalInventoryRow);
```
mwithi: Maybe I've not posted this comment: if a theoretic quantity is updated, it should also be notified to the user, in order to bring his/her attention on verifying the real quantity (that in turn, should NOT be updated by the validation like the other fields as it is a user input). So I remark again that the validate method should return a list of "errors" (i.e. messages) to be displayed to the user like other validation methods do.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +226,43 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+					if (lots != null) {
+						lot = lots.get(0);
+						double mainStoreQty = (double)lot.getMainStoreQuantity();
+						Integer medicalCode = medical.getCode();
+						Integer movMedicalCode = mov.getMedical().getCode();
+						if (movMedicalCode.equals(medicalCode) && lotCodeOfMovement.equals(lotCode)) {
+							if (mainStoreQty != theoQty) { 
+								medicalInventoryRow.setTheoreticQty(mainStoreQty);
+								medicalInventoryRow.setRealqty(mainStoreQty);
+								medicalInventoryRowManager.updateMedicalInventoryRow(medicalInventoryRow);
```
mwithi: Maybe I've not posted this comment: if a theoretic quantity is updated, it should also be notified to the user, in order to bring his/her attention on verifying the real quantity (that in turn, should NOT be updated by the validation like the other fields as it is a user input). So I remark again that the validate method should return a list of "errors" (i.e. messages) to be displayed to the user like other validation methods do.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +226,56 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
 			throw new OHDataValidationException(errors);
 		}
 	}
+	
+	@Transactional(rollbackFor = OHServiceException.class)
+	public void validateInventory(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList) throws OHDataValidationException, OHServiceException {
+		List<OHExceptionMessage> errors = new ArrayList<>();
```
dbmalkovsky: I don't believe the `errors` variable is referenced in this method.

src/main/java/org/isf/medicalinventory/service/MedicalInventoryRowIoOperation.java
```diff
@@ -93,4 +93,16 @@ public List<MedicalInventoryRow> getMedicalInventoryRowByInventoryId(int invento
...
+
+	/**
+	 * Return {@link MedicalInventoryRow} for passed param.
+	 *
+	 * @param medicalCode - the medical code.
+	 * @param lotCode - the lot code.
+	 * @return the {@link MedicalInventoryRow} object.
+	 * @throws OHServiceException
+	 */
+	public MedicalInventoryRow getMedicalInventoryRowByMedicalCodeAndLotCode(Integer medicalCode, String lotCode) throws OHServiceException {
```
dbmalkovsky: In the method definition (see line 39 below in the next file); the field is defined as a `int` yet the method definition here is an `Integer`.  This causes boxing/unboxing and is a performance hit. My preference is to use primitives instead of objects unless required.

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -24,6 +24,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
```
dbmalkovsky: where is `ArrayList` used in this file?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: Hi @ArnaudFonzam, why you marked this conversation as resolved without any change to the code?

src/main/java/org/isf/medicalinventory/service/MedicalInventoryRowIoOperation.java
mwithi: Hi @ArnaudFonzam why you marked as resolved this conversation without changes to the code or reason for keep the code uchanged?

src/test/java/org/isf/medicalsinventory/Tests.java
mwithi: Same here... if the import is not used please remove it, thanks!

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +226,55 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+						}
+					}
+				}
+			}
+		}
+		if (updated) {
+			List<OHExceptionMessage> errors = new ArrayList<>();
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.inventory.sometheoriticqtyhavebeenupdatepleasecontrole.msg")));
+			throw new OHDataValidationException(errors);
+		}
```
mwithi, line 278: Instead to have a single generic message, I would StringBuild a message including the generic bundle and appending the list of products/lots that have been actually changed.

ArnaudFonzam:
Hello I have close this PR because I used previously the Intesys repos And they need to close all repos conserning oh. I have create new PR by using my repos [OP-1189](https://github.com/informatici/openhospital-core/pull/1401)

[OP-1189]: https://openhospital.atlassian.net/browse/OP-1189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1382, opened 2024-07-31
OP-1292: Testing Coverage Improvements for SMS module 

dbmalkovsky:
See OP-1292

FWIW, I also made some code changes/improvements/spelling along with adding the tests.

### DIFF
--- a/src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
+++ b/src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
@@ -44,16 +44,16 @@ import feign.slf4j.Slf4jLogger;
 @PropertySource("classpath:sms.properties")
 public class SkebbyGatewayService implements SmsSenderInterface {
 
-	public static final String SERVICE_NAME = "skebby-gateway-service";
-	public static final String RESPONSE_SUCCESS = "OK";
-	public static final String KEY_PASSWORD = "skebby-gateway-service.password";
-	public static final String KEY_USERNAME = "skebby-gateway-service.username";
-	public static final String KEY_MESSAGE_TYPE = "skebby-gateway-service.message-type";
+	private static final String SERVICE_NAME = "skebby-gateway-service";
+	private static final String RESPONSE_SUCCESS = "OK";
+	protected static final String KEY_PASSWORD = "skebby-gateway-service.password";
+	protected static final String KEY_USERNAME = "skebby-gateway-service.username";
+	protected static final String KEY_MESSAGE_TYPE = "skebby-gateway-service.message-type";
 
-	public static final String KEY_USER_KEY = "skebby-gateway-service.userKey";
-	public static final String KEY_ACCESS_TOKEN = "skebby-gateway-service.accessToken";
+	protected static final String KEY_USER_KEY = "skebby-gateway-service.userKey";
+	protected static final String KEY_ACCESS_TOKEN = "skebby-gateway-service.accessToken";
 
-	public static final String KEY_SENDER = "skebby-gateway-service.sender";
+	private static final String KEY_SENDER = "skebby-gateway-service.sender";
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
```diff
@@ -44,16 +44,16 @@
...
-	private static final String SERVICE_NAME = "skebby-gateway-service";
-	private static final String RESPONSE_SUCCESS = "OK";
-	private static final String KEY_PASSWORD = "skebby-gateway-service.password";
-	private static final String KEY_USERNAME = "skebby-gateway-service.username";
-	private static final String KEY_MESSAGE_TYPE = "skebby-gateway-service.message-type";
+	public static final String SERVICE_NAME = "skebby-gateway-service";
+	public static final String RESPONSE_SUCCESS = "OK";
+	public static final String KEY_PASSWORD = "skebby-gateway-service.password";
+	public static final String KEY_USERNAME = "skebby-gateway-service.username";
+	public static final String KEY_MESSAGE_TYPE = "skebby-gateway-service.message-type";
```
mwithi: are we sure these fields should be public?

src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
```diff
@@ -44,16 +44,16 @@
...
 
-	private static final String KEY_USER_KEY = "skebby-gateway-service.userKey";
-	private static final String KEY_ACCESS_TOKEN = "skebby-gateway-service.accessToken";
+	public static final String KEY_USER_KEY = "skebby-gateway-service.userKey";
+	public static final String KEY_ACCESS_TOKEN = "skebby-gateway-service.accessToken";
```
mwithi: Also these?

src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
dbmalkovsky: I made them public so I could use them in the tests without duplicating the strings.

mwithi: yes I imagined that, but since are very sensitive fields shouldn't we use getters?

dbmalkovsky: If we used getters they would have to be `public` too.   So instead I made them protected and moved the test into the same package structure as the `SkebbyGatewayService.java` file.

src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
dbmalkovsky: See above

src/main/java/org/isf/sms/providers/skebby/SkebbyGatewayService.java
mwithi: sorry, my bad 😌 those are not the actual keys and tokens but just the strings to retrieve them from configuration files, so they were already safe, no matter those modificators... if you prefer, we can revert the last commit

dbmalkovsky: No; I think that the tests really should follow the main structure/layout so access to protected things are easier.   I would eventually like to make all the tests look more like this.

mwithi: shouldn't we do it in one shot later on instead to apply the new pattern only to this package?

dbmalkovsky: It is already done in other packages, for example, generaldata and utils.

mwithi (approved):
LGTM. Thanks a lot!

PR #1386, opened 2024-08-08
OH2-363 | Remove permissions CRUD Core methods

SilverD3:
Removal of unused methods related to [OH2-363](https://openhospital.atlassian.net/browse/OH2-363)

[OH2-363]: https://openhospital.atlassian.net/browse/OH2-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/permissions/manager/GroupPermissionManager.java
+++ b/src/main/java/org/isf/permissions/manager/GroupPermissionManager.java
@@ -33,7 +33,7 @@ import org.springframework.stereotype.Component;
 @Component
 public class GroupPermissionManager {
 
-	private GroupPermissionIoOperations operations;
+	private final GroupPermissionIoOperations operations;
 
 	public GroupPermissionManager(GroupPermissionIoOperations groupPermissionIoOperations) {
 		this.operations = groupPermissionIoOperations;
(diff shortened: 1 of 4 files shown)


src/main/java/org/isf/permissions/manager/GroupPermissionManager.java
```diff
@@ -33,7 +33,7 @@
 @Component
 public class GroupPermissionManager {
 
-	private GroupPermissionIoOperations operations;
+	private final GroupPermissionIoOperations operations;
```
dbmalkovsky, line 36: I do not see the value of adding `final`.  The variable is already `private` and adding `final` doesn't add anything to how the variable can be used.   In addition, if it is added, then there are hundreds of other occurrences like this in the code base that should be changed to match this pattern.   It isn't worth it IMHO.

SilverD3: Variables which values are set via DI (Dependency Injection) should be final as there are initialized once.

mwithi: Thank you @SilverD3, could you please open a "task" for this so we don't forget?

SilverD3: Sure. Issue OP-1325

mwithi (approved):
LGTM. Thanks a lot!

PR #1387, opened 2024-08-08
OP-1188  |  Delete Inventory

loique70:
See OP-1188.

Add deletion and test  function.

### DIFF
--- a/src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
+++ b/src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
@@ -214,10 +214,10 @@ public class MedicalInventoryManager {
 
 	/**
 	 * Marks an inventory as deleted by changing its status.
-	 * @param inventoryId - the ID of the inventory to delete.
+	 * @param medicalInventory - the medicalInventory of the inventory to delete.
 	 * @throws OHServiceException if an error occurs during the operation.
 	 */
-	public void deleteInventory(int inventoryId) throws OHServiceException {
-		ioOperations.deleteInventory(inventoryId);
+	public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceException {
+		ioOperations.deleteInventory(medicalInventory);
 	}
 }
--- a/src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
+++ b/src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
@@ -211,4 +211,13 @@ public class MedicalInventoryManager {
 			throw new OHDataValidationException(errors);
 		}
 	}
+
+	/**
+	 * Marks an inventory as deleted by changing its status.
+	 * @param inventoryId - the ID of the inventory to delete.
+	 * @throws OHServiceException if an error occurs during the operation.
+	 */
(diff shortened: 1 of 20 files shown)


src/main/java/org/isf/medicalinventory/service/MedicalInventoryIoOperation.java
```diff
@@ -198,4 +202,32 @@ public MedicalInventory getInventoryById(Integer inventoryId) throws OHServiceEx
...
 		return null;
 	}
+
+	/**
+	 * Marks an inventory as deleted by changing its status.
+	 * @param inventoryId - the ID of the inventory to delete.
+	 * @throws OHServiceException if an error occurs during the operation.
+	 */
+	public void deleteInventory(int inventoryId) throws OHServiceException {
+		List<OHExceptionMessage> errors = new ArrayList<>();
```
dbmalkovsky: Why use an List?   The way I read the code only one of 3 errors can be generated at any time.  Why not just throw each one as it happens?

src/main/java/org/isf/utils/exception/model/OHExceptionMessage.java
```diff
@@ -30,7 +30,7 @@
  *
  * @author akashytsa
  */
-public class OHExceptionMessage implements Serializable{
+public class OHExceptionMessage extends Throwable implements Serializable{
```
dbmalkovsky, line 33: This should not be needed.

loique70: This is comming from develop branch after the mergin.

src/main/java/org/isf/medicalinventory/service/MedicalInventoryIoOperation.java
loique70: You hav right, I have applyed the changes as you suggested.

src/main/java/org/isf/medicalinventory/service/MedicalInventoryIoOperation.java
```diff
@@ -22,12 +22,16 @@
 package org.isf.medicalinventory.service;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
```
dbmalkovsky, line 25: where is this import used in this file?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -211,4 +211,13 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
 		}
 	}
+
+	/**
+	 * Marks an inventory as deleted by changing its status.
+	 * @param medicalInventory - the medicalInventory of the inventory to delete.
+	 * @throws OHServiceException if an error occurs during the operation.
+	 */
+	public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceException {
+		ioOperations.deleteInventory(medicalInventory);
```
mwithi, line 221: I think we should also delete new lots created within the inventory, wdyt @ArnaudFonzam ?

ArnaudFonzam: absolutely @mwithi

mwithi:
Closing waiting a new PR

PR #1388, opened 2024-08-08
OP-1322: validate user name

dbmalkovsky:
See OP-1322

### DIFF
--- a/src/main/java/org/isf/menu/manager/UserBrowsingManager.java
+++ b/src/main/java/org/isf/menu/manager/UserBrowsingManager.java
@@ -43,6 +43,8 @@ import org.springframework.stereotype.Component;
 @Component
 public class UserBrowsingManager {
 
+	private static final String VALID_USERID_PATTERN = "^[A-Za-z0-9.]{1,255}$";
+
 	private MenuIoOperations ioOperations;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -97,9 +99,13 @@ public User getUserByName(String userName) throws OHServiceException {
 	 */
 	public User newUser(User user) throws OHServiceException {
 		String username = user.getUserName();
+		if (!username.matches(VALID_USERID_PATTERN)) {
+			throw new OHDataValidationException(
+							new OHExceptionMessage(MessageBundle.getMessage("angal.userbrowser.theusernamecontainsinvalidcharacters.msg")));
+		}
```
mwithi, line 105: why we don't adopt the "validation" pattern like in the other managers?

dbmalkovsky: Because this is the one and only place it is needed to be checked.  Why generate extra work.   See the next test too for another example of something done once.

mwithi: I see

mwithi (approved):
LGTM. Thanks a lot!

PR #1389, opened 2024-08-08
OP-1324: remove unused code

dbmalkovsky:
See OP-1324

Sorry the test class was added by accident; it is OK and would have been pushed soon anyway. :smirk:


dbmalkovsky:
Seems like there is an expectation that the export to CSV will be added; closing this PR.

PR #1390, opened 2024-08-20
OP-1187 | Add function for printing inventory repport

loique70:


src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -1097,4 +1099,37 @@ private static Date toDate(LocalDate localDate) {
 		return Optional.ofNullable(localDate).map(t -> Date.from(t.atStartOfDay().atZone(ZoneId.systemDefault()).toInstant())).orElse(null);
 	}
 
-}
+	public JasperReportResultDto getInventoryReportPdf(MedicalInventory inventory, String jasperFileName, int printQtyReal) throws OHServiceException {
+		try {
+			HashMap<String, Object> parameters = new HashMap<>();
```
dbmalkovsky:
```suggestion
			Map<String, Object> parameters = new HashMap<>();
```
The definition of the variable should use the interface not the concreate class.

src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -1097,4 +1099,37 @@ private static Date toDate(LocalDate localDate) {
...
+			parameters.put("City", hosp.getCity());
+			parameters.put("Email", hosp.getEmail());
+			parameters.put("Telephone", hosp.getTelephone());
+			parameters.put("inventoryId", String.valueOf(inventory.getId()));
+			parameters.put("inventoryReference", inventory.getInventoryReference());
+			parameters.put("inventoryStatus", status != null ? status : inventory.getStatus());
+			parameters.put("printQtyReal", printQtyReal);
+
+			LocalDateTime inventoryDateTime = inventory.getInventoryDate();
+			DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy");
```
dbmalkovsky: Why did you not use the static variable defined at line 89?

loique70: I have not make the remark for this variable declaration thanks.

src/main/java/org/isf/stat/manager/JasperReportsManager.java
loique70: Thank you.

src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -49,6 +49,8 @@
 import org.isf.generaldata.MessageBundle;
 import org.isf.hospital.manager.HospitalBrowsingManager;
 import org.isf.hospital.model.Hospital;
+import org.isf.medicalinventory.model.InventoryStatus;
```
mwithi, line 52: this import seems not used in the code

src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -1097,4 +1099,37 @@ private static Date toDate(LocalDate localDate) {
...
+	public JasperReportResultDto getInventoryReportPdf(MedicalInventory inventory, String jasperFileName, int printQtyReal) throws OHServiceException {
+		try {
+			Map<String, Object> parameters = new HashMap<>();
+			Hospital hosp = hospitalManager.getHospital();
+			String status = MessageBundle.getMessage(inventory.getStatus());
+			parameters.put("Hospital", hosp.getDescription());
+			parameters.put("Address", hosp.getAddress());
+			parameters.put("City", hosp.getCity());
+			parameters.put("Email", hosp.getEmail());
+			parameters.put("Telephone", hosp.getTelephone());
```
mwithi, line 1111: For this you can use `getHospitalParameters()` method

src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -1097,4 +1099,37 @@ private static Date toDate(LocalDate localDate) {
 		return Optional.ofNullable(localDate).map(t -> Date.from(t.atStartOfDay().atZone(ZoneId.systemDefault()).toInstant())).orElse(null);
 	}
 
-}
+	public JasperReportResultDto getInventoryReportPdf(MedicalInventory inventory, String jasperFileName, int printQtyReal) throws OHServiceException {
+		try {
+			Map<String, Object> parameters = new HashMap<>();
+			Hospital hosp = hospitalManager.getHospital();
+			String status = MessageBundle.getMessage(inventory.getStatus());
```
mwithi, line 1106: This cannot work, the `inventory.getStatus()` is not a bundle key but just a string from `enum InventoryStatus`.
If you want to have its translation I think we have to add some method in the `MedicalInventoryManager`.
I think that in the same way we could also add the `enum InventoryType`

```suggestion
			String status = inventory.getStatus();
```

mwithi (changes requested):
Please see suggestions

mwithi:
Closing for new PR on the way

PR #1391, opened 2024-08-22
OP-1309 | Create endpoints for CRUD on GroupPermissions

SilverD3:
See OP-1309


src/test/java/org/isf/menu/TestPermission.java
```diff
@@ -0,0 +1,30 @@
+package org.isf.menu;
```
dbmalkovsky, line 22: Missing license header.

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -172,9 +172,8 @@ public void resetFailedAttempts(User user) {
 	 * Lock the {@link User} from logging into the system.
 	 *
 	 * @param user the {@link User}
-	 * @throws OHServiceException
 	 */
-	public void lockUser(User user) throws OHServiceException {
+	public void lockUser(User user) {
```
dbmalkovsky: Why are you removing the `throws`?   Please read the code where the `ioOperations` class is annotated with `@TranslateOHServiceException` which uses `OHServiceExceptionTranslator` to wrap calls to the database using `Aspects` to catch and rethrow new exceptions.   Please review all the cases where you've removed the `throws` clause.  Thanks.

dbmalkovsky: Also check any new methods you added that don't currently include the `throws` clause.

SilverD3: My IDE (IntelliJ Idea) is removing them when I use the auto-clean feature, because it determines that this exception is not thrown by the method concerned. I will pay attention to this so that these Exceptions are no longer removed. 

I do not understand why we add OHServiceException on methods that do not explicitly throw this exception.
When I call a JpaRepository method like save(), all kinds of exceptions can be thrown like DataIntegrityViolationException, InvalidDataAccessResourceUsageException,..., it is not for that reason that I add all these exceptions in the signature of the method in question, but those exceptions are still raised when they're thrown by JpaRepository.

mwithi: @SilverD3 we add `throws OHServiceException` so callers know about it and they can catch and handle them, make sense to you?

SilverD3: Yeah! I see

SilverD3:
I think this PR is ready to be merged

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
mwithi: So are you going to restore them?

SilverD3: Already!

mwithi: what do you mean?

SilverD3: I was saying that I already restored them. I was asking the question just to understand.

mwithi: I'm asking because I see they are still removed on lines 188, 197 and 212 in the latest commit

SilverD3: Oh sorry. I missed those ones. Thanks! It's okay now

mwithi (approved):
LGTM. Thanks a lot!

PR #1394, opened 2024-08-29
OP-1295: Testing Coverage Improvements for UTILS module

dbmalkovsky:
See OP-1295

The overall coverage rate for the package is 81%.  Not every subpackage meets the 80% threshold and getting there is almost impossible.

BTW, I also fixed some bugs in the ExcelExporter class as the code has never been run/tested/used to my knowledge.

### DIFF
--- a/src/test/java/org/isf/utils/log/TestLogUtil.java
+++ b/src/test/java/org/isf/utils/log/TestLogUtil.java
@@ -1,0 +1,36 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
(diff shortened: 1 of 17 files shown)


mwithi (approved):
LGTM. Thanks a lot!

src/test/java/org/isf/utils/log/TestLogUtil.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.isf.utils.pagination;
```
mwithi, line 22: This package declaration looks wrong. Going to fix it with a new commit

PR #1402, opened 2024-09-09
OH2-322 | Auditing fields are not set/updated on Delivery Result Type

ArnaudFonzam:
See OH2-322.


src/main/java/org/isf/dlvrrestype/model/DeliveryResultType.java
```diff
@@ -21,16 +21,29 @@
  */
 package org.isf.dlvrrestype.model;
 
+import java.io.Serializable;
+
+import org.isf.utils.db.Auditable;
```
dbmalkovsky: The imports are out of order; `org` should be last.
See https://github.com/informatici/openhospital-core/blob/develop/.ide-settings/eclipse/OpenHospital.importorder

mwithi:
Hello @ArnaudFonzam, did you test also other entities? The Delievery Result Type stated in the issue's description was just an example

ArnaudFonzam:
> Hello @ArnaudFonzam, did you test also other entities? The Delievery Result Type stated in the issue's description was just an example

yes I have test all others And all others Object in administration Activity are working well. but we have another entity like DicomData and DicomType which don't have auditing fields I should add also for those twoo entity ?

mwithi:
> yes I have test all others And all others Object in administration Activity are working well. but we have another entity like DicomData and DicomType which don't have auditing fields I should add also for those twoo entity ?

The related tables (`oh_dicom_data` and `oh_dicomtype`) still miss those audit fields indeed... in order to add those we need a script for the DB... If you want to do it now for me it's ok, otherwise we can create a dedicated issue for this, how do you feel about this?

ArnaudFonzam:
> > yes I have test all others And all others Object in administration Activity are working well. but we have another entity like DicomData and DicomType which don't have auditing fields I should add also for those twoo entity ?
> 
> The related tables (`oh_dicom_data` and `oh_dicomtype`) still miss those audit fields indeed... in order to add those we need a script for the DB... If you want to do it now for me it's ok, otherwise we can create a dedicated issue for this, how do you feel about this?

For me It is better to fix its on this task.

sql/step_04_all_following_steps.sql
```diff
@@ -100,4 +100,5 @@ source step_a103_fix_MedicalDsrStockMovementTypeBrowser.sql;
 source step_a104_add_inventory_in_pharmacy.sql;
 source step_a105_medicaldsr_soft_deletion.sql;
 source step_a106_medicaldsrstock_control.sql;
-source step_a107_add_permission_admin_access.sql;
\ No newline at end of file
+source step_a107_add_permission_admin_access.sql;
+source step_a108_alter_table_dicom_type_and_dicom_data.sql;
```
mwithi: let's call it this way

```suggestion
source step_a108_audit_dicom_type_and_dicom_data.sql;
```

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -440,4 +440,16 @@ private List<Movement> prepareDishargingMovement(Movement movement, boolean chec
...
+	 * Retrieves all medicals referencing the specified code.
+	 * 
+	 * @param lotCode the lot code.
+	 * @return the ids of medicals referencing the specified lot.
+	 * @throws OHServiceException if an error occurs retrieving the referencing medicals.
+	 */
+	public List<Integer> getMedicalsFromLot(String lotCode) throws OHServiceException {
+		return ioOperations.getMedicalsFromLot(lotCode);
+	}
+
```
mwithi: is this meant to be in this PR?

src/main/java/org/isf/supplier/model/Supplier.java
```diff
@@ -39,6 +35,10 @@
 import jakarta.persistence.Transient;
 import jakarta.validation.constraints.NotNull;
 
+import org.apache.commons.lang3.StringUtils;
+import org.isf.utils.db.Auditable;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
+
```
mwithi, line 41: Also this is meant to be in this PR?

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
ArnaudFonzam: no it is and error I have remove it

src/main/java/org/isf/supplier/model/Supplier.java
ArnaudFonzam: yes it is import organisation that I have update

mwithi (approved):
LGTM. Thanks a lot!

PR #1403, opened 2024-09-09
OH2-378 | Set default value for supplierDeleted

ArnaudFonzam:
See OH2-378.


src/main/java/org/isf/supplier/model/Supplier.java
```diff
@@ -82,7 +82,7 @@ public class Supplier extends Auditable<String> implements Serializable {
 	private String supNote;
 
 	@Column(name = "SUP_DELETED", columnDefinition = "char(1) default 'N'")
-	private char supDeleted;
+	private char supDeleted = 'N';
```
mwithi, line 85: So it can be removed from the constructor?

ArnaudFonzam: yes

mwithi (approved):
LGTM. Thanks a lot!

PR #1404, opened 2024-09-09
OP-1148 Fix upgrade to spring boot 3 2

mwithi:
See OP-1148.

Also fixed Feign JUnit test (the TelemetryUrl may change over time)

### DIFF
--- a/src/main/java/org/isf/generaldata/DatabaseConfig.java
+++ b/src/main/java/org/isf/generaldata/DatabaseConfig.java
@@ -1,0 +1,10 @@
+package org.isf.generaldata;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@PropertySource("classpath:database.properties")
+public class DatabaseConfig {
+
+}
--- a/src/main/java/org/isf/generaldata/DatabaseConfig.java
+++ b/src/main/java/org/isf/generaldata/DatabaseConfig.java
@@ -1,3 +1,24 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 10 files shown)


src/main/java/org/isf/generaldata/DatabaseConfig.java
```diff
@@ -0,0 +1,10 @@
+package org.isf.generaldata;
```
dbmalkovsky, line 22: Missing license header

src/main/java/org/isf/generaldata/DatabaseConfig.java
```diff
@@ -0,0 +1,10 @@
...
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@PropertySource("classpath:database.properties")
```
mwithi, line 28: Tests can be fixed either adding the following property or adding `database.properties` to tests resources.
Of course the latter means that `database.properties` must be always present in the classpath, either from GUI or API.
Wdyt?

```suggestion
@PropertySource(value = "classpath:database.properties", ignoreResourceNotFound = true)
```

mwithi: Before, in the `applicationContext.xml` it was set to false:

```
	<bean
		class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
		<property name="location">
			<value>classpath:database.properties</value>
		</property>
		<property name="ignoreResourceNotFound" value="false" />
	</bean>
```

dbmalkovsky: I would lean towards the change in the `@PropertySource`

mwithi: I think that if we set not compulsory, it may lead to confusion and subtle behaviors... I've indeed discovered the problem because I was providing database.properties but was ignored...

mwithi: let's try with `database.properties` compulsory, so we can spot potential issues in other connected repositories

mizzioisf: I agree, in order to be consistent in every installation I would set the database.properties file as mandatory.

mwithi: Ok, then, I'm going to merge. Thanks!

PR #1406, opened 2024-09-13
OH2-253 | Reorganize user related endpoints

SteveGT96:
See [OH2-253](https://openhospital.atlassian.net/browse/OH2-253)

[OH2-253]: https://openhospital.atlassian.net/browse/OH2-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sql/step_96_user_settings_table.sql
mwithi, line 1: this file should remain unchanged, IDKW it is shown here 👀

sql/step_a109_audit_dicom_type_and_dicom_data.sql
mwithi, line 1: why you renamed the previous steps? just add the new one as `step_a109`

SteveGT96: My bad ! I've renamed the wrong one.

sql/step_96_user_settings_table.sql
SteveGT96: Maby just a code format change. I've restored changes and my editor code style is set to Open Hospital.

mwithi (dismissed):
LGTM. Thanks a lot!

sql/load_demo_data.sql
SilverD3, line 1: Add a migration script instead of modifying and existing one. Actually, to execute this db migration I have to clean my database and that's no good.
Don't forget that the app is already in production and users cannot clean their database just to apply migrations

SilverD3 (changes requested):

sql/load_demo_data.sql
mwithi: demo data is not meant for production, only changes to the main schema need migration steps.
of course demo data must be updated as well in order to be always coherent with the main schema.

SilverD3: Yeah! But for the instance, another migration step can be added just like step `step_a107_add_permission_admin_access.sql` to make the database migration easy, and add that script in the `create_all_demo.sql` only.
After positioning myself on this PR to test it, I found that the admin group no longer had permissions to manage user groups, and I couldn't find a migration script to run to correct my database without having to delete it. 
@SteveGT96 was explaining that it is the responsibility of the person deploying or testing the application to identify what has changed and to correct it, what can be tedious and impossible to do.

SteveGT96: Steps are meaned for main schema changes as @mwithi was saying. And demo data should always match the whole migration scripts. In production the solution to what I've done will just to ask to the admin to remove users.red permission to other groups.

mwithi: ok, hold on.

Migration scripts are meant for a clean DB, that has at least "admin" and "guest" users (the latter not very important though). If we are adding new permissions that before were not there (or we forgot to add in a previous step) we need a migration script, so that a fresh DB will have all permissions in place. 

Neverheless, we have also to balance this proliferiation of scripts: if a script has not been released yet (see scripts in last release 1.14.2), we can just modify the existing scripts instead of adding a new one. Of course this should be notified to the dev community so it is known that the new permissions will come along with a DB reenactment.

For demo data, in general, we should follow this procedure: 1. load_demo_data.sql 2. apply scripts for new features 3. add new demo data from the interface for the new features 4. dump a new load_demo_data.sql.

Neverheless, also for demo data, we can balance... if we just need to adjust one value in one table, we can modify the load_demo_data.sql script directly.

So questions: 

- is a fresh DB having all the expected admin permissions in place with the actual 108 steps or not?
- if not, can we modify a not yet released script to fix the problem (of course a concerning script) or not?
- if not, we need a migration script

SteveGT96: Where is the 'If yes' ?

SteveGT96: Anyway, `a fresh DB has all the expected admin permissions in place with the actual 108 steps`.

SilverD3: Here new permissions are been created, and I don't think permissions are demo data.

mwithi: the "if yes" option is in the question itself... Ok, so we don't need to add new steps.
therefore, I can assume that only the DB demo had missing values, hence its change, do you confirm?

mwithi: > Here new permissions are been created, and I don't think permissions are demo data.

@SilverD3 do you refer to `usergroups.create`, `usergroups.read`, `usergroups.update` and `usergroups.delete`?
@SteveGT96 these are not found in any of the 108 scripts indeed... so, as said before, please add them in the step 109, already part of this PR.

mwithi: (FWIW, I can't always be up to date on each line changed in the code, so I'm trying to give general procedures)

SteveGT96: OK !

SteveGT96: @mwithi , should I also asign them to the admin group in the same step ? or in the demo data ?

SilverD3: > @SilverD3 do you refer to usergroups.create, usergroups.read, usergroups.update and usergroups.delete?

Yes

mwithi: > @mwithi , should I also asign them to the admin group in the same step ? or in the demo data ?

yes, add them and assign to the admin user please. But why you removed from demo data? It is correct to add to main schema AND to the demo data.

SteveGT96: > > @mwithi , should I also asign them to the admin group in the same step ? or in the demo data ?
> 
> yes, add them and assign to the admin user please. But why you removed from demo data? It is correct to add to main schema AND to the demo data.

So I should kip them in both demo data and scripts, right ?

mwithi: > So I should kip them in both demo data and scripts, right ?

Yes.
If something is not clear we can discuss further on Slack.

SilverD3 (commented):
Looks good!

mwithi (approved):
LGTM. Thanks a lot!

PR #1410, opened 2024-09-22
OP-1335 change script to make it version independent.

dbmalkovsky:
See OP-1335


sql/step_89_alter_tables_innodb.sql
```diff
@@ -2,9 +2,18 @@ SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
```
mwithi: it is strange, because the check on FOREIGN_KEY is disabled at the beginning of the script exactly to avoid the DROP and ADD of the same (similarly in some other steps). Maybe MariaDB 10.11.8 changed the way to disable foreign key check? Or maybe these two tables have something different from others?

dbmalkovsky: My reading indicates that the checks became more strict (and maybe adding other global variables that controlled checking; can't remember) and that the way I changed the script is the more "acceptable" way.

mwithi: Yes what I'm saying is that if we should do the same for other scripts would result in a lot of changes (dropping and recreating foreign keys for each alter), while finding the right variable setup would be easier I guess.

Anyway, we can merge this as fix, and maybe create a new issue for the variable part, but I would add a reminder in a comment for this script that has been changed, wdyt?

dbmalkovsky: I updated the file with a comment about what is being done and I added an issue (that you can change as you wish) about reviewing other scripts.   As for finding the right variable setup, I tried various combinations without luck.  And even if one can be found there is no guarantee that it will work in the next version of the database (as was the case here; it worked in 10.6.x but failed in 10.8.x).

mwithi: It has been working for long time, but both methods have vantages and disvantage, like dropping and recreating a foreign key, especially on exisiting databases, means recreating also the indexes associated with them, reducing performance in the operation. On the other hand disabling the checks works better for small changes but maybe not always sufficient. So I'm not sure it should be done on all scripts like the OP-1338 suggests.
I will put as pending for further analysis.

mwithi (approved):
LGTM. Thanks a lot!

PR #1411, opened 2024-09-22
OP-1336: make timestamp columns backward compatible with newer versions of Mariadb

dbmalkovsky:
See OP-1336


mwithi (approved):
LGTM. Thanks a lot!

PR #1413, opened 2024-09-27
Chore: no LOGGER needed and wrong class specified in addition

dbmalkovsky:
Found because the wrong class was specified for the `.getLogger()` method.


mwithi (approved):
LGTM. Thanks a lot!

PR #1419, opened 2024-10-01
Chore: remove variable holding unused utility class

dbmalkovsky:
`TestPermission` is a utility class with only static methods and is never used in this test.


mwithi (approved):
LGTM. Thanks a lot!

PR #1420, opened 2024-10-01
Chore: previous test proves it is empty so remove next test

dbmalkovsky:
Line 57 shows `configData` is empty so testing a `get` on the object is silly.


mwithi (approved):
LGTM. Thanks a lot!

PR #1421, opened 2024-10-01
Chore: changes found from bad local variable naming conventions

dbmalkovsky:
Found local variables that started with a capital letter.  In one case it should be a `static final` and in the other the variable isn't needed.


mwithi (approved):
LGTM. Thanks a lot!

PR #1430, opened 2024-10-04
OP-1305 | Add confirm inventory method

ArnaudFonzam:
See OP-1305.


dbmalkovsky (dismissed):
There should be multiple tests to validate the new method.  I can't see approving this PR without it.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+		String chargeCode = inventory.getChargeType();
+		Integer supplierId = inventory.getSupplier();
+		String wardCode = inventory.getDestination();
+		MovementType chargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(chargeCode);
+		MovementType dischargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(dischargeCode);
+		Supplier supplier = supplierManager.getByID(supplierId);
+		Ward ward = wardManager.findWard(wardCode);
+		LocalDateTime today = LocalDateTime.now();
+		for (Iterator<MedicalInventoryRow> iterator = inventoryRowSearchList.iterator(); iterator.hasNext();) {
+			String reference = "Ref-" + LocalDateTime.now();
```
mwithi, line 411: Make sense to prefix with "Inv-"?
```suggestion
			String reference = "Inv-" + LocalDateTime.now();
```

mwithi: Shouldn't the reference be given by the inventory record itself?

```suggestion
			String reference = "Ref-" + LocalDateTime.now();
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			MedicalInventoryRow medicalInventoryRow = (MedicalInventoryRow) iterator.next();
+			double theoQty = medicalInventoryRow.getTheoreticQty();
+			double realQty = medicalInventoryRow.getRealQty();
+			Double ajustQty = theoQty - realQty;
+			Medical medical = medicalInventoryRow.getMedical();
+			String lotCode = medicalInventoryRow.getLot().getCode();
+			Lot currentLot = movStockInsertingManager.getLot(lotCode);
+			boolean isNew = medicalInventoryRow.isNewLot();
```
mwithi, line 419: This should not be necessary because new lots have been already created at this point so the only thing to do is to use them.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+				dischargeMovement.add(movement);
+				dischargeMovement = movStockInsertingManager.newMultipleDischargingMovements(dischargeMovement, reference);
+			}
+			// If new lot has been inserted, create charge movement with real qty
+			if (isNew) {
+				Movement movement = new Movement(medical, chargeType, null, currentLot, today, (int) realQty, supplier, reference);
+				List<Movement> chargeMovement = new ArrayList<>();
+				chargeMovement.add(movement);
+				chargeMovement = movStockInsertingManager.newMultipleChargingMovements(chargeMovement, reference);
+			}
```
mwithi: This should not be necessary because new lots have been already created at this point so the only thing to do is to use them, am I wrong?
```suggestion
			if (realQty > theoQty) { // charge movement when realQty > theoQty
				Movement movement = new Movement(medical, chargeType, null, currentLot, today, -(ajustQty.intValue()), supplier, reference);
				List<Movement> chargeMovement = new ArrayList<>();
				chargeMovement.add(movement);
				chargeMovement = movStockInsertingManager.newMultipleChargingMovements(chargeMovement, reference);
			} else { // discharge movement when realQty < theoQty
				Movement movement = new Movement(medical, dischargeType, ward, currentLot, today, ajustQty.intValue(), null, reference);
				List<Movement> dischargeMovement = new ArrayList<>();
				dischargeMovement.add(movement);
				dischargeMovement = movStockInsertingManager.newMultipleDischargingMovements(dischargeMovement, reference);
			}
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
ArnaudFonzam: here is the reference of movement not the reference of inventory

mwithi: I know, but why should they differ?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
ArnaudFonzam: yes the lot is create but it is not link to the movement.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
ArnaudFonzam: okay I will use the reference of inventory instead of creating new reference

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: the lot is linked to a movement within the movement itself, not in any other way...

ArnaudFonzam: please when you doing charge movement in main store with new lot you create first the new lot and then link the lot to the movement right ? but why here we just create lot without  link that lot to movement ?

mwithi: the charge movement will be done in the first `if` if you remove the condition `isNew`

ArnaudFonzam: okay

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
ArnaudFonzam: Look that method if we use the same reference for different movement : 

```
protected List<OHExceptionMessage> checkReferenceNumber(String referenceNumber) throws OHServiceException {
		if (referenceNumber == null || referenceNumber.isEmpty()) {
			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.multiplecharging.pleaseinsertareferencenumber.msg")));
		} else {
			if (refNoExists(referenceNumber)) {
				errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.medicalstock.multiplecharging.theinsertedreferencenumberalreadyexists.msg")));
			}
		}
		return errors;
	}
```

>

mwithi: we always do that, just use the methods:

- newMultipleChargingMovements
- newMultipleDischargingMovements

ArnaudFonzam: okay

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+		MovementType dischargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(dischargeCode);
+		Supplier supplier = supplierManager.getByID(supplierId);
+		Ward ward = wardManager.findWard(wardCode);
+		LocalDateTime now = TimeTools.getNow();
+		for (Iterator<MedicalInventoryRow> iterator = inventoryRowSearchList.iterator(); iterator.hasNext();) {
+			MedicalInventoryRow medicalInventoryRow = (MedicalInventoryRow) iterator.next();
+			StringBuilder referenceBuilder = new StringBuilder();
+			referenceBuilder.append(inventory.getInventoryReference())
+			                .append("-")
+			                .append(now);
```
mwithi: There's no need of adding timestamp here
```suggestion
			referenceBuilder.append(inventory.getInventoryReference());
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			double theoQty = medicalInventoryRow.getTheoreticQty();
+			double realQty = medicalInventoryRow.getRealQty();
+			Double ajustQty = theoQty - realQty;
```
mwithi: Difference is better to calculate as "now" - "before" so it automatically reflect the sign of the operation:
```suggestion
			Double ajustQty = realQty - theoQty;
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			Medical medical = medicalInventoryRow.getMedical();
+			String lotCode = medicalInventoryRow.getLot().getCode();
+			Lot currentLot = movStockInsertingManager.getLot(lotCode);
+			if (realQty > theoQty) { // charge movement when realQty > theoQty
+				referenceBuilder.append("-charge");
+				String reference = referenceBuilder.toString();
+				Movement movement = new Movement(medical, chargeType, null, currentLot, now, -(ajustQty.intValue()), supplier, reference);
```
mwithi:
```suggestion
				Movement movement = new Movement(medical, chargeType, null, currentLot, now, ajustQty.intValue(), supplier, reference);
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			referenceBuilder.append(inventory.getInventoryReference())
+			                .append("-")
+			                .append(now);
+			double theoQty = medicalInventoryRow.getTheoreticQty();
+			double realQty = medicalInventoryRow.getRealQty();
+			Double ajustQty = theoQty - realQty;
+			Medical medical = medicalInventoryRow.getMedical();
+			String lotCode = medicalInventoryRow.getLot().getCode();
+			Lot currentLot = movStockInsertingManager.getLot(lotCode);
+			if (realQty > theoQty) { // charge movement when realQty > theoQty
```
mwithi:
```suggestion
			if (ajustQty > 0) { // charge movement when realQty > theoQty
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+				referenceBuilder.append("-charge");
+				String reference = referenceBuilder.toString();
+				Movement movement = new Movement(medical, chargeType, null, currentLot, now, -(ajustQty.intValue()), supplier, reference);
+				List<Movement> chargeMovement = new ArrayList<>();
```
mwithi: You are resetting everytime the list of charging movements... this should be likely be defined before the iteration cycle

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+				chargeMovement.add(movement);
+				chargeMovement = movStockInsertingManager.newMultipleChargingMovements(chargeMovement, reference);
```
mwithi: in this way you are inserting only one movement... this should be likely called after the iteration cycle, and only if `chargeMovement` list is not empty
```suggestion
				chargeMovement = movStockInsertingManager.newMultipleChargingMovements(chargeMovement, reference);
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			} else if (realQty < theoQty) { // discharge movement when realQty < theoQty
+				referenceBuilder.append("-discharge");
+				String reference = referenceBuilder.toString();
+				Movement movement = new Movement(medical, dischargeType, ward, currentLot, now, ajustQty.intValue(), null, reference);
+				List<Movement> dischargeMovement = new ArrayList<>();
```
mwithi: Same here, you are resetting the list of dicharging movements at every cycle, move outside the iteration cycle

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+				dischargeMovement.add(movement);
+				dischargeMovement = movStockInsertingManager.newMultipleDischargingMovements(dischargeMovement, reference);
```
mwithi: you are inserting only one movement... this should be likely called after the iteration cycle, and only if dischargeMovement is not empty

```suggestion
				dischargeMovement = movStockInsertingManager.newMultipleDischargingMovements(dischargeMovement, reference);
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			                .append("-")
+			                .append(now);
+			double theoQty = medicalInventoryRow.getTheoreticQty();
+			double realQty = medicalInventoryRow.getRealQty();
+			Double ajustQty = theoQty - realQty;
+			Medical medical = medicalInventoryRow.getMedical();
+			String lotCode = medicalInventoryRow.getLot().getCode();
+			Lot currentLot = movStockInsertingManager.getLot(lotCode);
+			if (realQty > theoQty) { // charge movement when realQty > theoQty
+				referenceBuilder.append("-charge");
```
mwithi: this would append a "-charge" at each cycle.... I think we should just define a `chargeReference` and a `dischargeReference` outside the iteration cycle

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+				String reference = referenceBuilder.toString();
```
mwithi: `.toString()` will be called automatically when casting to String
```suggestion
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: Moreover, since we are adding two suffix, in the validation we have to check that the reference typed in by the user does not exceed 40 chars, as MMV_REFNO is defined as VARCHAR(50)

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: Moreover there is a problem with AUTOMATICLOT_IN=yes: before the inventory feature, the charging movements were done generating automatically a new lot code and asking the user only the expiring date. 
Now, instead, we want to insert a charge movement where the lot is specified (because already created/generated).
We didn't take this into account when creating new lots during the inventory... let's think how to solve this problem...

Option 1)
We don't create lots during the inventory and leave the system to do it during confirmation and according with the general settings, using the data in the inventory row. In this case we can also avoid to delete new lots when deleting an inventory or an inventoryrow with "N"

Option 2)
We change the way a new charge movement is created by taking into account that the lot might be provided even if the AUTOMATICLOT_IN=yes. 
> For instance the check in `MedicalStockIoOperations.newMovement(Movement)` at the moment is `if (isAutomaticLotMode() || "".equals(lotCode))`
> Maybe is enough to change it to `if (isAutomaticLotMode() && "".equals(lotCode))`, but it needs testing

In the case 2) we have to pay attention to not brake previous behaviour.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi, line 1: Summarizing all comments below, this should be the result:

```
	/**
	 * Confirm the Inventory rows of inventory.
	 *
	 * @param inventory - The {@link MedicalInventory}
	 * @param inventoryRowSearchList- The list of {@link MedicalInventory}
	 * @throws OHDataValidationException
	 */
	@Transactional(rollbackFor = OHServiceException.class)
	public List<Movement> confirmMedicalInventoryRow(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList) throws OHServiceException {
		// validate the inventory
		this.validateMedicalInventoryRow(inventory, inventoryRowSearchList);
		// get general info
		String referenceNumber = inventory.getInventoryReference();
		// TODO: make possibility to allow charges and discharges with same referenceNumber
		String chargeReferenceNumber = referenceNumber + "-charge";
		String dischargeReferenceNumber = referenceNumber + "-discharge";
		MovementType chargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(inventory.getChargeType());
		MovementType dischargeType = medicalDsrStockMovementTypeBrowserManager.getMovementType(inventory.getDischargeType());
		Supplier supplier = supplierManager.getByID(inventory.getSupplier());
		Ward ward = wardManager.findWard(inventory.getDestination());
		LocalDateTime now = TimeTools.getNow();
		// prepare movements
		List<Movement> chargeMovements = new ArrayList<>();
		List<Movement> dischargeMovements = new ArrayList<>();
		for (Iterator<MedicalInventoryRow> iterator = inventoryRowSearchList.iterator(); iterator.hasNext();) {
			MedicalInventoryRow medicalInventoryRow = (MedicalInventoryRow) iterator.next();

			double theoQty = medicalInventoryRow.getTheoreticQty();
			double realQty = medicalInventoryRow.getRealQty();
			Double ajustQty = realQty - theoQty;
			Medical medical = medicalInventoryRow.getMedical();
			Lot currentLot = medicalInventoryRow.getLot();
			if (ajustQty > 0) { // charge movement when realQty > theoQty
				Movement movement = new Movement(medical, chargeType, null, currentLot, now, ajustQty.intValue(), supplier, chargeReferenceNumber);
				chargeMovements.add(movement);
			} else if (ajustQty < 0) { // discharge movement when realQty < theoQty
				Movement movement = new Movement(medical, dischargeType, ward, currentLot, now, -(ajustQty.intValue()), null, dischargeReferenceNumber);
				dischargeMovements.add(movement);
			} // else ajustQty = 0, continue
		}
		// create movements
		List<Movement> insertedMovements = new ArrayList<>();
		if (!chargeMovements.isEmpty()) {
			insertedMovements.addAll(movStockInsertingManager.newMultipleChargingMovements(chargeMovements, chargeReferenceNumber));
		}
		if (!dischargeMovements.isEmpty()) {
			insertedMovements.addAll(movStockInsertingManager.newMultipleDischargingMovements(dischargeMovements, dischargeReferenceNumber));
		}
		return insertedMovements;
	}
```

Testing Option 2):

In `if (isAutomaticLotMode() || "".equals(lotCode))` the OR (`||`) here is needed in order to quickly fallback into generating a lotcode when AUTOMATICLOT_IN=false and the user leaves the code field empty.
So, changing to AND (`&&`) here would works for the inventory confirmation but brakes this behavior.
Still investigating...

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: pushed a change for Option 2)

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: pushed a change for Option 2)

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
...
+			StringBuilder referenceBuilder = new StringBuilder();
+			referenceBuilder.append(inventory.getInventoryReference())
+			                .append("-")
+			                .append(now);
+			double theoQty = medicalInventoryRow.getTheoreticQty();
+			double realQty = medicalInventoryRow.getRealQty();
+			Double ajustQty = realQty - theoQty;
+			Medical medical = medicalInventoryRow.getMedical();
+			String lotCode = medicalInventoryRow.getLot().getCode();
+			Lot currentLot = movStockInsertingManager.getLot(lotCode);
```
mwithi: why do we need to fetch the lot from the DB and not using the `medicalInventoryRow.getLot()` directly?

ArnaudFonzam: Because we don't have information about mainStoreQty.

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +575,63 @@ void testDeleteInventory() throws Exception {
 		assertThat(deletedInventory).isNotNull();
 		assertThat(deletedInventory.getStatus()).isEqualTo(InventoryStatus.canceled.toString());
 	}
+	
+	@Test
+	void testValidateMedicalInventory() throws Exception {
+		Ward ward = testWard.setup(false);
+		wardIoOperationRepository.saveAndFlush(ward);
+		MedicalInventory inventory = testMedicalInventory.setup(ward, false);
```
mwithi, line 583: you have to persist also chargeType and dischargeType as "inventory+" and "inventory-" are not found in the inmemory DB

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: where do you use mainStoreQty here? It should be already saved in the theoQty

ArnaudFonzam: yes I think we can remove it

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +575,63 @@ void testDeleteInventory() throws Exception {
...
+		inventory = medicalInventoryIoOperation.updateMedicalInventory(inventory);
+		assertThat(inventory).isNotNull();
+		assertThat(inventory.getStatus()).isEqualTo(status);		
+	}
+	
+	@Test
+	void testConfirmMedicalInventory() throws Exception {
+		Ward ward = testWard.setup(false);
+		wardIoOperationRepository.saveAndFlush(ward);
+		MedicalInventory inventory = testMedicalInventory.setup(ward, false);
```
mwithi, line 682: you have to persist also chargeType and dischargeType as "inventory+" and "inventory-" are not found in the inmemory DB

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +586,72 @@ void testDeleteInventory() throws Exception {
...
+		assertThat(inventory).isNotNull();
+		assertThat(inventory.getStatus()).isEqualTo(status);		
+	}
+	
+	@Test
+	void testConfirmMedicalInventory() throws Exception {
+		Ward ward = testWard.setup(false);
+		wardIoOperationRepository.saveAndFlush(ward);
+		MovementType chargeType = new MovementType("inventory+", "inventory +", "+", "non-operational");
+		MovementType dischargeType = new MovementType("inventory-", "inventory -", "-", "non-operational");
```
mwithi:
```suggestion
		MovementType chargeType = new MovementType("inventory+", "Inventory+", "+", "non-operational");
		MovementType dischargeType = new MovementType("inventory-", "Inventory-", "-", "non-operational");
```

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +586,72 @@ void testDeleteInventory() throws Exception {
...
+		Supplier supplier = new Supplier(3, "INVEnTORY", null, null, null, null, null, null);
```
mwithi:
```suggestion
		Supplier supplier = new Supplier(3, "INVENTORY", null, null, null, null, null, null);
```

src/test/java/org/isf/medicalsinventory/Tests.java
mwithi: it seems that the supplier also is missing when it comes to create the charge movement

ArnaudFonzam: I added also destination but I have the same error

mwithi: Fixed with 59237d65c8af98832088ea2c552157d5199cc5c2

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -330,7 +330,7 @@ public List<Movement> newMultipleChargingMovements(List<Movement> movements, Str
 				errors.add(new OHExceptionMessage(
 								mov.getMedical() != null ? mov.getMedical().getDescription()
 												: MessageBundle.getMessage("angal.medicalstock.nodescription.txt")));
-				throw new OHDataValidationException(errors);
+				throw new OHDataValidationException(errors.get(0));
```
mwithi: Revert change that was breaking many tests
```suggestion
				throw new OHDataValidationException(errors);
```

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +587,72 @@ void testDeleteInventory() throws Exception {
...
+		List<Movement> insertMovements = medicalInventoryManager.confirmMedicalInventoryRow(savedInventory, medicalInventoryRows);
+		assertThat(insertMovements).isNotEmpty();
+		int inventoryId = savedInventory.getId();
+		inventory = medicalInventoryIoOperation.getInventoryById(inventoryId);
+		assertThat(inventory).isNotNull();
+		String status = InventoryStatus.done.toString();
+		inventory.setStatus(status);
+		inventory = medicalInventoryIoOperation.updateMedicalInventory(inventory);
+		assertThat(inventory).isNotNull();
+		assertThat(inventory.getStatus()).isEqualTo(status);
```
mwithi: I think status should be automatically set in the confirm and then only checked during the test

src/test/java/org/isf/medicalsinventory/Tests.java
mwithi, line 1: I think some tests should be executed with parameters like other test in medicalstock packages
```
@ParameterizedTest(name = "Test with AUTOMATICLOT_IN={0}, AUTOMATICLOT_OUT={1}, AUTOMATICLOTWARD_TOWARD={2}")
@MethodSource("automaticlot")
```

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +587,72 @@ void testDeleteInventory() throws Exception {
...
+		medicalDsrStockMovementTypeIoOperationRepository.saveAndFlush(chargeType);
+		supplierIoOperationRepository.saveAndFlush(supplier);
+		wardIoOperationRepository.saveAndFlush(destination);
+		MedicalInventory inventory = testMedicalInventory.setup(ward, false);
+		MedicalInventory savedInventory = medicalInventoryIoOperation.newMedicalInventory(inventory);
+		MedicalType medicalType = testMedicalType.setup(false);
+		Medical medical = testMedical.setup(medicalType, false);
+		Lot lot = testLot.setup(medical, false);
+		MedicalInventoryRow medicalInventoryRow = testMedicalInventoryRow.setup(savedInventory, medical, lot, false);
+		medicalInventoryRow.setRealqty(60);
```
mwithi: I think it worth to test at least 3 inventoryRow: one leading to a charge, another leading to a discharge and one leading to 0, in order to cover the whole method under test

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -572,4 +587,72 @@ void testDeleteInventory() throws Exception {
 		assertThat(deletedInventory).isNotNull();
 		assertThat(deletedInventory.getStatus()).isEqualTo(InventoryStatus.canceled.toString());
 	}
+
+	@Test
+	void testValidateMedicalInventory() throws Exception {
```
mwithi: If you check the coverage of this method, you will see that most of it is not covered:

image


Missing also a `testValidateMedicalInventoryRow`:

image

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,4 +386,55 @@ public void deleteInventory(MedicalInventory medicalInventory) throws OHServiceE
 		}
 		ioOperations.deleteInventory(medicalInventory);
```
mwithi, line 376: I think the other `deleteMedicalInventory` method in this manager can be deleted?

ArnaudFonzam: yes

ArnaudFofou:
to be close. the new work is on https://github.com/informatici/openhospital-core/pull/1436

mwithi:
Superseded by https://github.com/informatici/openhospital-core/pull/1436

PR #1431, opened 2024-10-08
OP-1311 | Create 'oh_settings' table and methods in CORE

SilverD3:
See OP-1311.


sql/step_a111_settings.sql
```diff
@@ -0,0 +1,95 @@
...
+(NULL, 'ALLOWPRINTOPENEDBILL', 'bool', NULL, 'FALSE', 'FALSE', 'Allow open bills printing', 'admin', NULL, NOW(), NULL, 0, 1, 1),
+(NULL, 'ALLOWMULTIPLEOPENEDBILL', 'bool', NULL, 'FALSE', 'TRUE', 'Allow multiple open bills for a patient', 'admin', NULL, NOW(), NULL, 0, 1, 1),
+(NULL, 'DEBUG', 'bool', NULL, 'FALSE', 'FALSE', 'Use debug mode', 'admin', NULL, NOW(), NULL, 0, 1, 1),
+(NULL, 'SESSIONTIMEOUT', 'number', NULL, '5', '5', 'Number in minute for session timeout', 'admin', NULL, NOW(), NULL, 0, 1, 1),
+
+-- Not sure if we need to store MODE in the database
+-- (NULL, 'MODE', 'text', NULL, ', ', ', 'admin', NULL, NOW(), NULL, 0, 1, 1),
+
+-- Not sure if we need to store DEMODATA in the database
+-- (NULL, 'DEMODATA', 'bool', NULL, 'FALSE', 'FALSE', NULL, 'admin', NULL, NOW(), NULL, 0, 1, 1),
```
SilverD3: @mwithi @dbmalkovsky please could you explain the use of these settings and tell me if we should store them in the database

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,95 @@
...
+  SETT_VALUE_OPTIONS VARCHAR(500) NULL DEFAULT NULL COMMENT "Comma-separated list of possible values, in case of type 'select'",
+  SETT_DEFAULT_VALUE VARCHAR(255) NOT NULL,
+  SETT_VALUE VARCHAR(255) NOT NULL,
+  SETT_DESCRIPTION VARCHAR(500) NULL,
+  SETT_CREATED_BY VARCHAR(50) NULL DEFAULT NULL,
+  SETT_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
+  SETT_CREATED_DATE DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
+  SETT_MODIFIED_DATE DATETIME NULL DEFAULT NULL,
+  SETT_DELETED TINYINT(1) NOT NULL DEFAULT 0,
+  SETT_EDITABLE TINYINT(1) NOT NULL DEFAULT 1,
```
mwithi: what this two are meant for? Why should we provide soft-deletion and in which case a setting should not be editable (i.e. not changed even by the admin user)?

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,95 @@
...
+  SETT_CODE VARCHAR(50) NOT NULL,
+  SETT_VALUE_TYPE ENUM('bool', 'number', 'text', 'select') NOT NULL DEFAULT 'text',
+  SETT_VALUE_OPTIONS VARCHAR(500) NULL DEFAULT NULL COMMENT "Comma-separated list of possible values, in case of type 'select'",
+  SETT_DEFAULT_VALUE VARCHAR(255) NOT NULL,
+  SETT_VALUE VARCHAR(255) NOT NULL,
+  SETT_DESCRIPTION VARCHAR(500) NULL,
+  SETT_CREATED_BY VARCHAR(50) NULL DEFAULT NULL,
+  SETT_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
+  SETT_CREATED_DATE DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
+  SETT_MODIFIED_DATE DATETIME NULL DEFAULT NULL,
```
mwithi: normally the fields and their order are the following:
```suggestion
  SETT_CREATED_BY VARCHAR(50) NULL DEFAULT NULL,
  SETT_CREATED_DATE DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP(),
  SETT_LAST_MODIFIED_BY VARCHAR(50) NULL DEFAULT NULL,
  SETT_LAST_MODIFIED_DATE DATETIME NULL DEFAULT NULL,
```

sql/step_a111_settings.sql
mwithi: they are used in the scripts and also `APISERVER` is still experimental, you can remove them

src/main/java/org/isf/settings/manager/SettingManager.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

src/main/java/org/isf/settings/model/Setting.java
```diff
@@ -0,0 +1,152 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

src/main/java/org/isf/settings/model/Setting.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @author Silevester D.
+ * @since v1.15
+ */
+@Entity
+@Table(name = "OH_SETTINGS")
+@AttributeOverride(name = "createdBy", column = @Column(name = "SETT_CREATED_BY", updatable = false))
+@AttributeOverride(name = "createdDate", column = @Column(name = "SETT_CREATED_DATE", updatable = false))
+@AttributeOverride(name = "lastModifiedBy", column = @Column(name = "SETT_MODIFIED_BY"))
+@AttributeOverride(name = "lastModifiedDate", column = @Column(name = "SETT_MODIFIED_DATE"))
```
mwithi:
```suggestion
@AttributeOverride(name = "lastModifiedBy", column = @Column(name = "SETT_LAST_MODIFIED_BY"))
@AttributeOverride(name = "lastModifiedDate", column = @Column(name = "SETT_LAST_MODIFIED_DATE"))
```

src/main/java/org/isf/settings/service/SettingIoOperationRepository.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

src/main/java/org/isf/settings/service/SettingIoOperations.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```

sql/step_a111_settings.sql
mwithi, line 1: Do you think we should add also a field for the category (e.g. General, Pharmacy, Laboratory, etc...) and a bool to specify if the application needs a restart in order to see the new value to take place (e.g. LANGUAGE)?

mwithi, line 1: Should this feature include also all other .properties in the future?

sql/step_a111_settings.sql
SilverD3: For `SETT_DELETED` I was just following convention, but I think we don't really need soft deletion as `settings` can't be deleted, at least, by the user.

For `SETT_EDITABLE`, I think we should not allow users to change some settings value like `PARAMSURL` and "report file names" (like `VISITSHEET`)

sql/step_a111_settings.sql
SilverD3: For the category, does it affect the way settings are handled? I mean, should "General settings" be handled differently from "Pharmacy settings". Anyway, if you like it, it's okay for me, we can add categories.

> a bool to specify if the application needs a restart in order to see the new value to take place (e.g. LANGUAGE)

Not really. As settings are read (in real time) from the database, we don't need to restart the app to apply changes if the user change any setting.

SilverD3: I was thinking I already added them all. If there are missing ones, please could you list them under this PR so that I can add them also?

mwithi: the category may help in arranging the settings in GUI and UI (e.g. different tabs), rather than hardcode the layout, but I don't know if there are other better solutions.

In the case of LANGUAGE at least the application has to reload the bundles files (unless we change the way they are loaded). I don't know other case but it might be there.

mwithi: I would add them few at time, at least as they are divided now in the code, hence the idea of the categories.
Anyway, they are the ones extending `ConfigurationProperties`

sql/step_a111_settings.sql
mwithi: all fields are there to be modifiable, otherwise they would be inside the code and not outside. The only setting that might benefit of this is SINGLEUSER by the moment is very sensitive, although I don't think one should write directly in the DB in order to change it.

Also, about the DEFAULT_VALUE: that is meant to take place when the parameter is not specified at all, but this means that the record is not in the DB, so also its default value... the default value will be still hardcoded in the code and this field will be there just as reminder to reset the field, right?

SilverD3: > all fields are there to be modifiable, otherwise they would be inside the code and not outside. The only setting that might benefit of this is SINGLEUSER by the moment is very sensitive, although I don't think one should write directly in the DB in order to change it.

Should I then remove or keep it?

> Also, about the DEFAULT_VALUE: that is meant to take place when the parameter is not specified at all, but this means that the record is not in the DB, so also its default value... the default value will be still hardcoded in the code and this field will be there just as reminder to reset the field, right?

No, once we have moved the settings to the database, we'll now only rely on the database to read settings. If a settings is not present in the database (no record for the corresponding setting), we just raise an exception and cancel the current operation. DEFAULT_VALUE let users see the "factory" value of the setting and is also use to reset a setting value to its "factory" value. 
If you see things differently, explain to me so that I can put it into practice.

sql/step_a111_settings.sql
SilverD3: Okay. Let's add these two columns: `category` and `needRestart`

sql/step_a111_settings.sql
mwithi: > Should I then remove or keep it?

Let's remove it for now as it sound useless

> If you see things differently, explain to me so that I can put it into practice.

We are just brainstorming here, by the moment the issue is not describing every single case.
What do you mean with "just raise an exception and cancel the current operation"? I imagine the settings are loaded at the application start, so "cancelling the current operation" means exit the application?
Anyway, let's keep DEFAULT_VALUE for the "factory reset".

sql/step_a111_settings.sql
SilverD3: Okay. Let's move them progressively.

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,88 @@
+-- Drop the existing table
+DROP TABLE IF EXISTS OH_SETTINGS;
+
+-- Create settings table structure
+CREATE TABLE OH_SETTINGS (
+  SETT_ID INT(11) NOT NULL AUTO_INCREMENT COMMENT 'Default generated ID',
+  SETT_CODE VARCHAR(50) NOT NULL,
+  SETT_CATEGORY ENUM('general', 'telemetry', 'printer', 'dicom', 'examination', 'report', 'sms') NOT NULL DEFAULT 'general',
```
mwithi: let's use the same grouping already defined in the `settings.properties.dist` in GUI
and then reflect in the `INSERT INTO` below and in the `LoadSettingTable.sql`

sql/step_04_all_following_steps.sql
mwithi, line 1: missing changes on `load_demo_data.sql`

SilverD3: `create_all_demo.sql` include `step_04_all_following_steps.sql` in which I've included `step_a111_settings.sql`. Do I need to load settings again in `load_demo_data.sql`?

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,88 @@
+-- Drop the existing table
+DROP TABLE IF EXISTS OH_SETTINGS;
+
+-- Create settings table structure
+CREATE TABLE OH_SETTINGS (
+  SETT_ID INT(11) NOT NULL AUTO_INCREMENT COMMENT 'Default generated ID',
+  SETT_CODE VARCHAR(50) NOT NULL,
+  SETT_CATEGORY ENUM('general', 'telemetry', 'security', 'dicom', 'accounting', 'reports', 'pharmacy', 'gui', 'experimental', 'application') NOT NULL DEFAULT 'general',
```
mwithi: Sorry, the order matters in ENUM fields.
Changed 'dicom' -> 'imaging' (more broad concept)
Removed 'experimental' -> APISERVER could be left outside 
```suggestion
  SETT_CATEGORY ENUM('general', 'application', 'gui', 'accounting', 'pharmacy', 'imaging', 'reports', 'security', 'telemetry') NOT NULL DEFAULT 'general',
```

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,88 @@
...
+(NULL, 'PATIENTVACCINEEXTENDED', 'bool', NULL, 'FALSE', 'FALSE', 'Use extended patient vaccine', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'MERGEFUNCTION', 'bool', NULL, 'TRUE', 'TRUE', 'Enable patient merge feature', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'SMSENABLED', 'bool', NULL, 'FALSE', 'FALSE', 'Enable SMS sending', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'VIDEOMODULEENABLED', 'bool', NULL, 'FALSE', 'FALSE', 'Whether to enable Video module or not', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'ENHANCEDSEARCH', 'bool', NULL, 'FALSE', 'FALSE', 'Enable enhanced search', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'LABMULTIPLEINSERT', 'bool', NULL, 'FALSE', 'FALSE', 'Use multiple insert for lab exams', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'MATERNITYRESTARTINJUNE', 'bool', NULL, 'FALSE', 'FALSE', 'Restart maternity in June', 'admin', NULL, NOW(), NULL, 1, 1, 'application'),
+(NULL, 'APISERVER', 'bool', NULL, 'FALSE', 'FALSE', 'Whether API server is enabled or not', 'admin', NULL, NOW(), NULL, 1, 1, 'experimental'),
```
mwithi:
```suggestion
```

sql/step_a111_settings.sql
```diff
@@ -0,0 +1,88 @@
...
+(NULL, 'AUTOMATICLOTWARD_TOWARD', 'bool', NULL, 'TRUE', 'TRUE', 'Automatically create and assign lot to ward to ward stock mouvements', 'admin', NULL, NOW(), NULL, 1, 1, 'pharmacy'),
+(NULL, 'INTERNALPHARMACIES', 'bool', NULL, 'FALSE', 'FALSE', 'Allow internal pharmacies', 'admin', NULL, NOW(), NULL, 1, 1, 'pharmacy'),
+(NULL, 'LOTWITHCOST', 'bool', NULL, 'TRUE', 'TRUE', 'Use lot with cost', 'admin', NULL, NOW(), NULL, 1, 1, 'pharmacy'),
+(NULL, 'RECEIPTPRINTER', 'bool', NULL, 'FALSE', 'FALSE', 'Enable bill receipt printing', 'admin', NULL, NOW(), NULL, 1, 1, 'accounting'),
+(NULL, 'ALLOWPRINTOPENEDBILL', 'bool', NULL, 'FALSE', 'FALSE', 'Allow open bills printing', 'admin', NULL, NOW(), NULL, 1, 1, 'accounting'),
+(NULL, 'ALLOWMULTIPLEOPENEDBILL', 'bool', NULL, 'FALSE', 'TRUE', 'Allow multiple open bills for a patient', 'admin', NULL, NOW(), NULL, 1, 1, 'accounting'),
+(NULL, 'DICOMMODULEENABLED', 'bool', NULL, 'FALSE', 'FALSE', 'Whether to enable DICOM module or not', 'admin', NULL, NOW(), NULL, 1, 1, 'dicom'),
+(NULL, 'DICOMTHUMBNAILS', 'bool', NULL, 'TRUE', 'TRUE', 'Enable DICOM thumbnails (previews)', 'admin', NULL, NOW(), NULL, 1, 1, 'dicom'),
```
mwithi, line 71:
```suggestion
(NULL, 'DICOMTHUMBNAILS', 'bool', NULL, 'TRUE', 'TRUE', 'Enable DICOM thumbnails (previews)', 'admin', NULL, NOW(), NULL, 1, 1, 'imaging'),
```

mwithi:
```suggestion
(NULL, 'DICOMMODULEENABLED', 'bool', NULL, 'FALSE', 'FALSE', 'Whether to enable DICOM module or not', 'admin', NULL, NOW(), NULL, 1, 1, 'imaging'),
(NULL, 'DICOMTHUMBNAILS', 'bool', NULL, 'TRUE', 'TRUE', 'Enable DICOM thumbnails (previews)', 'admin', NULL, NOW(), NULL, 1, 1, 'imaging'),
```

sql/step_04_all_following_steps.sql
mwithi: yes, `load_demo_data.sql` is executed after `source delete_all_data.sql`.
we should also update `delete_all_data.sql` with the TRUNCATE of the new table too... missed in my previous comment

src/main/java/org/isf/settings/manager/SettingManager.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	public Setting update(Setting setting) throws OHServiceException {
+		Setting existingSetting = operations.getByCode(setting.getCode());
+		existingSetting.setValue(setting.getValue());
+
+		return operations.save(setting);
+	}
+
+	public Setting getByCode(String code) {
```
dbmalkovsky: Why doesn't this method have `throws OHServiceException` like in other packages?

src/main/java/org/isf/settings/manager/SettingManager.java
```diff
@@ -0,0 +1,73 @@
...
+		return operations.getByCode(code);
+	}
+
+	public Setting getById(int id) {
```
dbmalkovsky: Why doesn't this method have `throws OHServiceException` like in other packages?

src/main/java/org/isf/settings/service/SettingIoOperations.java
```diff
@@ -0,0 +1,67 @@
...
+
+	public List<Setting> findAll() throws OHServiceException {
+		return repository.findAll();
+	}
+
+	public Setting save(Setting setting) throws OHServiceException {
+		return repository.save(setting);
+	}
+
+	public Setting getByCode(String code) {
```
dbmalkovsky: Why doesn't this method have `throws OHServiceException` like in other packages?

src/main/java/org/isf/settings/service/SettingIoOperations.java
```diff
@@ -0,0 +1,67 @@
...
+		return repository.findFirstByCode(code);
+	}
+
+	public Setting getById(int id) {
```
dbmalkovsky: Why doesn't this method have `throws OHServiceException` like in other packages?

src/main/java/org/isf/settings/manager/SettingManager.java
SilverD3: Because it doesn't raise any of the exception caught by `org/isf/utils/db/OHServiceExceptionTranslator.java`

src/main/java/org/isf/settings/service/SettingIoOperations.java
SilverD3: Because it doesn't raise any of the exception caught by `org/isf/utils/db/OHServiceExceptionTranslator.java`

dbmalkovsky: (1) How can you be so sure that it doesn't raise one of the many `RunTime`  or `Throwable` exceptions caught by that method?  One basic one would be "OutOfMemoryError'.

(2) If you are so sure, then why haven't you opened issues for removing the `throws` clause for other instances of similar methods?

mwithi: despite the `JpaRepository` interface doesn't throws direclty any exception, in the project we use to specify the `throws OHServiceException` on each method using it so they can be spotted at compilation time and the dev is forced to catch them

mwithi: so please add them, like in the rest of the code.

mwithi:
This needs to be tested widely, also taking into account GUI and the application starting scripts, I will do it in the next weeks, but probably will finish in November

SilverD3:
> This needs to be tested widely, also taking into account GUI and the application starting scripts, I will do it in the next weeks, but probably will finish in November

Too long . we need this feature to start the integration of ORTHANC

SilverD3:
It will not create a breaking change with the CORE. The CORE rely on GeneralData and ConfigurationProperties to read and write configs. The API will be relying on the database

mwithi:
please resolve conflicts

SilverD3:
> please resolve conflicts

Done!

mwithi:
Superseded by #1527

PR #1434, opened 2024-10-14
OP-1340 Add method for multiple lots update

mwithi:
See OP-1340.

### DIFF
--- a/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
+++ b/src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
@@ -24,12 +24,10 @@ package org.isf.medicalstock.manager;
 import java.math.BigDecimal;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 
 import org.isf.generaldata.GeneralData;
 import org.isf.generaldata.MessageBundle;
-import org.isf.medicalinventory.model.MedicalInventoryRow;
 import org.isf.medicals.model.Medical;
 import org.isf.medicals.service.MedicalsIoOperations;
 import org.isf.medicalstock.model.Lot;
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -352,16 +350,33 @@ protected Movement prepareChargingMovement(Movement movement, boolean checkRefer
...
+	 * @param lots - the list of lots.
+	 * @return the list of {@link Lot}s updated.
+	 * @throws OHServiceException if an error occurs during the check.
+	 */
+	@Transactional
+	public List<Lot> updateLot(List<Lot> lots) throws OHServiceException {
+		for (Lot lot : lots) {
+			lot = ioOperations.updateLot(lot);
+		}
+		return lots;
```
dbmalkovsky: This doesn't look like it is returning a list of the updated `lots` but rather the original list of `lots`.   Most likely they are the same but shouldn't the returned `lot` from `ioOperations.updateLot(lot);` be added to a new `List` of `Lot` objects and that be returned?    As it is not the return value from`ioOperations.updateLot(lot)` is not used and thrown away.

mwithi: Nice catch! I was testing without a new list and it worked (the returned lot is indeed keeping the same reference in memory, although it could not be always true), then I left it like that. Fixed, thanks!

dbmalkovsky (dismissed):
LGTM but it would be better if there was a unit test for the new method

mwithi:
> LGTM but it would be better if there was a unit test for the new method

Done, thanks!

src/test/java/org/isf/medicalstock/Tests.java
```diff
@@ -782,6 +783,26 @@ void testMgrGetLotsByMedicalNull(boolean in, boolean out, boolean toward) throws
 		assertThat(movStockInsertingManager.getLotByMedical(null, true)).isEmpty();
 	}
 
+	@ParameterizedTest(name = "Test with AUTOMATICLOT_IN={0}, AUTOMATICLOT_OUT={1}, AUTOMATICLOTWARD_TOWARD={2}")
+	@MethodSource("automaticlot")
+	public void testMgrUpdateLots() throws Exception {
```
dbmalkovsky:
```suggestion
	void testMgrUpdateLots() throws Exception {
```
With Junit 5 the standard is not to make test methods public; see all the other methods in this file

mwithi: Right! Thanks!

dbmalkovsky (dismissed):
LGTM

PR #1439, opened 2024-10-25
OP-1043 | Implement soft delete on user and usergroup

SteveGT96:
See [OP-1043](https://openhospital.atlassian.net/browse/OP-1043)

[OP-1043]: https://openhospital.atlassian.net/browse/OP-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/isf/menu/service/UserGroupIoOperationRepository.java
```diff
@@ -1,43 +1,50 @@
...
-
-	@Modifying
-	@Query(value = "update UserGroup ug set ug.desc=:description where ug.code=:id")
-	int updateDescription(@Param("description") String description, @Param("id") String id);
-
-	List<UserGroup> findByCodeIn(List<String> userGroupIds);
-}
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
```
mwithi:
```suggestion
 * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)

```

src/main/java/org/isf/menu/model/User.java
```diff
@@ -45,33 +45,36 @@
...
-	@Column(name="US_LOCK_TIME")
+	@Column(name = "US_LOCK_TIME")
 	private LocalDateTime lockedTime;
 
-	@Column(name="US_LAST_LOGIN")
+	@Column(name = "US_LAST_LOGIN")
 	private LocalDateTime lastLogin;
 
+	@Column(name = "US_DELETED", columnDefinition = "BOOLEAN DEFAULT false")
+	private boolean deleted = false;
```
dbmalkovsky: setting the field to `false` is redundant; `false` is the default value.

src/main/java/org/isf/menu/model/UserGroup.java
```diff
@@ -40,12 +40,15 @@
...
 	private String code;
-	
-	@Column(name="UG_DESC")
+
+	@Column(name = "UG_DESC")
 	private String desc;
-	
+
+	@Column(name = "UG_DELETED", columnDefinition = "BOOLEAN DEFAULT false")
+	private boolean deleted = false;
```
dbmalkovsky: setting the field to `false` is redundant; `false` is the default value.

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -175,18 +166,17 @@ public boolean isGroupNamePresent(String groupName) throws OHServiceException {
...
 	 * @throws OHServiceException When failed to create user
 	 */
 	public User newUser(User user) throws OHServiceException {
+		user.setDeleted(false);
 		return repository.save(user);
 	}
 
 	/**
 	 * Updates an existing {@link User} in the DB
-	 *
```
dbmalkovsky, line 189: Should this method check to see if the user is `deleted` before updating it?

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -197,7 +187,6 @@ public boolean updateUser(User user) throws OHServiceException {
 
 	/**
 	 * Updates the password of an existing {@link User} in the DB
-	 *
```
dbmalkovsky, line 200: Should this method check to see if the user is `deleted` before updating it?

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -208,12 +197,12 @@ public boolean updatePassword(User user) throws OHServiceException {
 
 	/**
 	 * Deletes an existing {@link User}
-	 * 
 	 * @param user - the {@link User} to delete
 	 * @throws OHServiceException When failed to delete user
 	 */
 	public void deleteUser(User user) throws OHServiceException {
-		repository.delete(user);
+		user.setDeleted(true);
```
dbmalkovsky: Should an error be thrown if the user is already `deleted`?

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -313,19 +299,16 @@ private GroupMenu insertGroupMenu(UserGroup aGroup, UserMenuItem item) throws OH
...
 	 * Deletes a {@link UserGroup}
-	 * 
 	 * @param aGroup - the {@link UserGroup} to delete
 	 * @throws OHServiceException When failed to delete group
 	 */
 	public void deleteGroup(UserGroup aGroup) throws OHServiceException {
-		groupMenuRepository.deleteWhereUserGroup(aGroup.getCode());
-		groupPermissionIoOperationRepository.deleteAllByUserGroup_Code(aGroup.getCode());
-		groupRepository.delete(aGroup);
+		aGroup.setDeleted(true);
```
dbmalkovsky: I am going to stop asking about whether checks for deletion should be made in update/delete methods.   But a consistent answer is needed.

dbmalkovsky (changes requested):
Left specific and general comments.

mwithi (dismissed):
LGTM. Please provide other PRs on API and GUI. Don't forget about DOC in a later step.

SteveGT96:
> LGTM. Please provide other PRs on API and GUI. Don't forget about DOC in a later step.

Other PRs on GUI and API ? GUI and API are compatible with this PRs. If you try to delete use from GUI or API, it'll be soft deleted and you won't get it any more in future requests. For the doc, which doc are you talking about ? The doc module  ? (I didn't work on it yet), or the API doc ?

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -151,26 +144,23 @@ public void deleteUser(User user) throws OHServiceException {
...
 		ioOperations.updateFailedAttempts(user.getUserName(), newFailAttempts);
 	}
 
 	/**
 	 * Reset the number of failed login attempts to zero for {@link User}.
-	 *
 	 * @param user the {@link User}
 	 */
-	public void resetFailedAttempts(User user) {
+	public void resetFailedAttempts(User user) throws OHServiceException {
```
mwithi: This is impacting GUI

mwithi:
> > LGTM. Please provide other PRs on API and GUI. Don't forget about DOC in a later step.
> 
> Other PRs on GUI and API ? GUI and API are compatible with this PRs. If you try to delete use from GUI or API, it'll be soft deleted and you won't get it any more in future requests. For the doc, which doc are you talking about ? The doc module ? (I didn't work on it yet), or the API doc ?

Left comments under the issue for discussion

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
SteveGT96: I'm going to revert this one ! We don't need it !

SteveGT96:
> > > LGTM. Please provide other PRs on API and GUI. Don't forget about DOC in a later step.
> > 
> > 
> > Other PRs on GUI and API ? GUI and API are compatible with this PRs. If you try to delete use from GUI or API, it'll be soft deleted and you won't get it any more in future requests. For the doc, which doc are you talking about ? The doc module ? (I didn't work on it yet), or the API doc ?
> 
> Left comments under the issue for discussion

OK, I'm gonna check them ASAP.

src/main/java/org/isf/menu/model/User.java
```diff
@@ -45,33 +45,36 @@
...
-	@Column(name="US_LOCK_TIME")
+	@Column(name = "US_LOCK_TIME")
 	private LocalDateTime lockedTime;
 
-	@Column(name="US_LAST_LOGIN")
+	@Column(name = "US_LAST_LOGIN")
 	private LocalDateTime lastLogin;
 
+	@Column(name = "US_DELETED")
+	private boolean deleted = false;
```
dbmalkovsky:
```suggestion
	private boolean deleted;
```
assigning the value `false` is redundant.

src/main/java/org/isf/menu/model/UserGroup.java
```diff
@@ -40,12 +40,15 @@
...
 	private String code;
-	
-	@Column(name="UG_DESC")
+
+	@Column(name = "UG_DESC")
 	private String desc;
-	
+
+	@Column(name = "UG_DELETED")
+	private boolean deleted = false;
```
dbmalkovsky:
```suggestion
	private boolean deleted;
```
assigning the value `false` is redundant.

dbmalkovsky (dismissed):
The conversation was only partially resolved.

mwithi (changes requested):
Left some requests on API and GUI.
Please add to demo data the `oh_groupmenu` permissions `btnlaboratorynew` and `btnlaboratoryedit`  for the laboratorist group

src/main/java/org/isf/menu/model/User.java
```diff
@@ -150,7 +153,12 @@ public LocalDateTime getLastLogin() {
 	public void setLastLogin(LocalDateTime lastLogin) {
 		this.lastLogin = lastLogin;
 	}
-
+	public boolean isDeleted() {
+		return deleted;
+	}
+	public void setDeleted(boolean deleted) {
+		this.deleted = deleted;
+	}
```
mwithi:
```suggestion
	public boolean isDeleted() {
		return deleted;
	}
	
	public void setDeleted(boolean deleted) {
		this.deleted = deleted;
	}
```

src/main/java/org/isf/menu/model/UserGroup.java
```diff
@@ -65,7 +68,12 @@ public String getCode() {
 	public void setCode(String code) {
 		this.code = code;
 	}
-
+	public boolean isDeleted() {
+		return deleted;
+	}
+	public void setDeleted(boolean deleted) {
+		this.deleted = deleted;
+	}
```
mwithi:
```suggestion
	public boolean isDeleted() {
		return deleted;
	}
	
	public void setDeleted(boolean deleted) {
		this.deleted = deleted;
	}
```

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -313,30 +305,32 @@ private GroupMenu insertGroupMenu(UserGroup aGroup, UserMenuItem item) throws OH
...
-		groupMenuRepository.deleteWhereUserGroup(aGroup.getCode());
-		groupPermissionIoOperationRepository.deleteAllByUserGroup_Code(aGroup.getCode());
-		groupRepository.delete(aGroup);
+		ensureUserGroupNotDeleted(aGroup.getCode());
+		if (getUser(aGroup.getCode()).isEmpty()) {
+			groupRepository.delete(aGroup);
+		} else {
+			aGroup.setDeleted(true);
+			groupRepository.save(aGroup);
+		}
```
mwithi: In this way a group will be set to deleted only when there are no users in it, but in this case why to keep it in the DB?

mwithi: sorry, misunderstood because I'm having an error when deleting not empty groups, so maybe is for another reason

mwithi: the check about existing users is already done in the manager... maybe we have not to change the logic of the `deleteGroup` (or `deleteUser`) methods in IoOperations as you are doing, but just use the update methods to set the `deleted=true`

src/main/java/org/isf/menu/service/UserIoOperationRepository.java
```diff
@@ -62,10 +71,16 @@ public interface UserIoOperationRepository extends JpaRepository<User, String> {
 	@Query(value = "update User set lastLogin=:lastLoggedIn where userName=:id")
 	void setLastLogin(@Param("lastLoggedIn") LocalDateTime lockTime, @Param("id") String id);
 
-	@Query("select count(u) from User u where active=1")
+	@Query("select count(u) from User u where active=1 and deleted!=true")
```
dbmalkovsky:
```suggestion
	@Query("select count(u) from User u where active=1 and deleted=false")
```
Why use `!=true`???

src/main/java/org/isf/menu/service/UserIoOperationRepository.java
```diff
@@ -62,10 +71,16 @@ public interface UserIoOperationRepository extends JpaRepository<User, String> {
...
 	long countAllActiveUsers();
 
-	@Query("select count(g) from UserGroup g where active=1")
+	@Query("select count(u) from User u where active=1 and deleted=:deleted")
+	long countAllActiveUsersByDeleted(@Param("deleted") boolean deleted);
+
+	@Query("select count(g) from UserGroup g where active=1 and deleted!=true")
```
dbmalkovsky:
```suggestion
	@Query("select count(g) from UserGroup g where active=1 and deleted=false")
```
Why use `!=true`?   What am I missing?   Especially since if `false` is passed to `countAllActiveGroupsByDeleted()` (see line 83) the condition is `and deleted=false`

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -175,45 +166,49 @@ public boolean isGroupNamePresent(String groupName) throws OHServiceException {
...
 	/**
 	 * Updates an existing {@link User} in the DB
-	 * 
 	 * @param user - the {@link User} to update
 	 * @return new {@link User}
 	 * @throws OHServiceException When failed to update user
 	 */
 	public boolean updateUser(User user) throws OHServiceException {
+		ensureUserNotDeleted(user.getUserName());
 		return repository.updateUser(user.getDesc(), user.getUserGroupName(), user.getUserName()) > 0;
```
dbmalkovsky: Please see https://openhospital.atlassian.net/browse/OP-1130

`new` operations should return the object (which these files seem to do)
`update` operations should return the object (not a boolean as found in these files)
`delete` operations should return null (which these files seem to do)

It appears that I missed updating all the JPA operations in this package when I did all the others.  Sorry

dbmalkovsky (commented):
I left a few comments and suggestions.

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -175,45 +166,49 @@ public boolean isGroupNamePresent(String groupName) throws OHServiceException {
...
 	 * @param user - the {@link User} to delete
 	 * @throws OHServiceException When failed to delete user
 	 */
 	public void deleteUser(User user) throws OHServiceException {
-		repository.delete(user);
+		ensureUserNotDeleted(user.getUserName());
+		if (user.isDeleted()) {
+			repository.save(user);
+		} else {
+			repository.delete(user);
```
dbmalkovsky: Isn't this a hard delete?   Once again I'm confused.
Should it be a call to `user.setDeleted(true)`, and followed by `repository.save(user)`?

src/main/java/org/isf/menu/service/MenuIoOperations.java
mwithi: in other words, hard-deletion should use "delete" methods (and return void) soft-deletion is done with a "update" methods

mwithi (dismissed):

sql/step_04_all_following_steps.sql
```diff
@@ -105,3 +105,4 @@ source step_a108_audit_dicom_type_and_dicom_data.sql;
 source step_a109_update_user_settings_table_constraints_and_add_usergroups_permissions.sql;
 source step_a110_update_operations_table_change_ope_for_to_enum.sql;
 source step_a111_add_missing_lock_columns.sql;
+source step_a112_update_users_and_user_groups.sql;
```
mwithi:
```suggestion
source step_a112_users_and_groups_soft_deletion.sql;
```

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -334,37 +313,42 @@ public UserGroup newUserGroup(UserGroup userGroup, List<Permission> permissions)
...
 	 */
-	public boolean updateUserGroup(UserGroup userGroup) throws OHServiceException {
+	public UserGroup updateUserGroup(UserGroup userGroup) throws OHServiceException {
+		if (userGroup.isDeleted()) {
+			List<User> users = getUser(userGroup.getCode());
+			if (users != null && users.stream().anyMatch(User::isDeleted)) {
+				throw new OHDataValidationException(
+					new OHExceptionMessage(MessageBundle.getMessage("angal.groupsbrowser.thisgrouphasusersandcannotbedeleted.msg")));
+			}
+		}
```
mwithi: this check is already done in `deleteGroup(UserGroup aGroup)`, we don't need it here

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -398,6 +386,20 @@ public boolean updateUserGroup(UserGroup userGroup, List<Permission> permissions
 			groupPermissionIoOperationRepository.saveAll(groupPermissions);
 		}
 
-		return updated;
+		return findByCode(userGroup.getCode());
+	}
+
+	public void ensureUserNotDeleted(String username) throws OHServiceException {
+		User entity = repository.findByUserName(username);
+		if (entity == null) {
```
dbmalkovsky: The method name says `ensure User Not Deleted`; but the implementation is only checking for a hard delete by checking for `null` (i.e., no longer in the database).    I would have thought that given the method name it would be checking the deleted flag.

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -398,6 +386,20 @@ public boolean updateUserGroup(UserGroup userGroup, List<Permission> permissions
...
+			throw new OHServiceException(new OHExceptionMessage("This operation is not allowed"));
+		}
+	}
+
+	public void ensureUserGroupNotDeleted(String code) throws OHServiceException {
+		UserGroup entity = findByCode(code);
+		if (entity == null) {
```
dbmalkovsky: See previous comment as it applies here too.

dbmalkovsky (commented):
Added new comments.

src/main/java/org/isf/menu/service/MenuIoOperations.java
SteveGT96: You're right !

src/test/java/org/isf/menu/Tests.java
```diff
@@ -321,12 +323,52 @@ void testIoUpdateUserGroup() throws Exception {
...
+	@Test
+	void testMgrSoftDeleteUser() throws Exception {
+		String userName = setupTestUser(false);
+		User user = userBrowsingManager.getUserByName(userName);
+		assertThat(user).isNotNull();
+		assertThat(user.isDeleted()).isFalse();
+
+		user.setDeleted(true);
+		User foundUser = userIoOperationRepository.findByUserName(userName);
+		assertThat(foundUser).isNotNull();
```
dbmalkovsky, line 341: how does this work????   you are setting the `isDeleted` field on the local object `user`.   It is never persisted to the database.   So when you `findByUserName()` shouldn't the `foundUser` be equal to the original `user` (that is, not deleted)?

And why in line 335 do you use `getUserByName()` and in line 340 you are using `findByUserName()`?  What about `getUserByNameAndIsDeleted()`?

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -101,40 +97,43 @@ public User newUser(User user) throws OHServiceException {
...
 	 * @param user - the {@link User} to update
-	 * @return {@code true} if the user has been updated, {@code false} otherwise.
+	 * @return the {@link User} that has been updated
 	 * @throws OHServiceException When failed to update user
 	 */
-	public boolean updateUser(User user) throws OHServiceException {
+	public User updateUser(User user) throws OHServiceException {
+		User oldUser = ioOperations.getUserByNameAndIsDeleted(user.getUserName());
+
+		if (oldUser != null && oldUser.isDeleted() && user.isDeleted()) {
```
mwithi: Isn't `getUserByNameAndIsDeleted` already returning a isDeleted user (or null)?
If yes, the condition `... && oldUser.isDeleted() ...` is for sure `true`?

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -334,37 +319,49 @@ public UserGroup newUserGroup(UserGroup userGroup, List<Permission> permissions)
...
-	 * @return {@code true} if the group has been updated, {@code false} otherwise.
+	 * @return the {@link UserGroup} that has been updated
+	 * @throws OHServiceException If failed to update group
 	 */
-	public boolean updateUserGroup(UserGroup userGroup) throws OHServiceException {
+	public UserGroup updateUserGroup(UserGroup userGroup) throws OHServiceException {
+
+		UserGroup group = ioOperations.findByCodeAndIsDeleted(userGroup.getCode());
+
+		if (group != null && group.isDeleted() && userGroup.isDeleted()) {
```
mwithi: Isn't `findByCodeAndIsDeleted` already returning a isDeleted group (or `null`)?
If yes, the condition `... && group.isDeleted() ...` is for sure `true`?

src/main/java/org/isf/menu/manager/UserGroupManager.java
```diff
@@ -39,4 +39,8 @@ public UserGroupManager(UserGroupIoOperationRepository userGroupIoOperationRepos
 	public List<UserGroup> findByIdIn(List<String> userGroupIds) {
 		return this.userGroupIoOperationRepository.findByCodeIn(userGroupIds);
 	}
+
+	public List<UserGroup> findByIdIn(List<String> userGroupIds, boolean deleted) {
+		return this.userGroupIoOperationRepository.findByDeletedAndCodeIn(deleted, userGroupIds);
```
mwithi:
```suggestion
		return this.userGroupIoOperationRepository.findByIdInAndDeleted(userGroupIds, deleted);
```

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -175,44 +185,48 @@ public boolean isGroupNamePresent(String groupName) throws OHServiceException {
...
-	 * 
 	 * @param user - the {@link User} to delete
 	 * @throws OHServiceException When failed to delete user
 	 */
 	public void deleteUser(User user) throws OHServiceException {
+		if (!user.isDeleted()) {
+			ensureUserNotDeleted(user.getUserName());
+		}
 		repository.delete(user);
 	}
```
mwithi: why the if condition here? should not be enough `ensureUserNotDeleted(user.getUserName());` like in the other methods?

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -313,30 +324,27 @@ private GroupMenu insertGroupMenu(UserGroup aGroup, UserMenuItem item) throws OH
...
 	 * Insert a new {@link UserGroup} with a minimum set of rights
-	 * 
 	 * @param aGroup - the {@link UserGroup} to insert
 	 * @return the new {@link UserGroup}
 	 * @throws OHServiceException When failed to create group
 	 */
 	public UserGroup newUserGroup(UserGroup aGroup) throws OHServiceException {
+		aGroup.setDeleted(false);
 		return groupRepository.save(aGroup);
 	}
```
mwithi: You are suggesting that every time we have an entity with a `_DELETED` field we have to be sure that `newEntity...` methods set that field to `false`?

src/main/java/org/isf/menu/service/UserIoOperationRepository.java
```diff
@@ -37,15 +37,29 @@ public interface UserIoOperationRepository extends JpaRepository<User, String> {
...
+	List<User> findAllByDeleted(boolean deleted);
+
+	User findByUserNameAndDeleted(String userName, boolean deleted);
+
 	@Query(value = "select user from User user where user.userGroupName.code=:groupId order by user.userName")
 	List<User> findAllWhereUserGroupNameByOrderUserNameAsc(@Param("groupId") String groupId);
 
+	@Query(value = "select user from User user where user.userGroupName.code=:groupId and deleted=:deleted order by user.userName")
+	List<User> findByDeletedWhereUserGroupNameByOrderUserNameAsc(@Param("groupId") String groupId, @Param("deleted") boolean deleted);
+
```
mwithi: These methods seem wrongly named:
- `findAll` --> when returning a list
- `By` --> standard is to use By (in place of Where)
- `FieldsNames` like parameters using the same order

```suggestion
	@Query(value = "select user from User user where user.userGroupName.code=:groupId order by user.userName")
	List<User> findAllByUserGroupIdOrderByUserNameAsc(@Param("groupId") String groupId);

	@Query(value = "select user from User user where user.userGroupName.code=:groupId and deleted=:deleted order by user.userName")
	List<User> findAllByUserGroupIdAndDeletedOrderByUserNameAsc(@Param("groupId") String groupId, @Param("deleted") boolean deleted);

```

src/test/java/org/isf/menu/Tests.java
SteveGT96: I've passed several time in debuging that. You can see that tests are passing. When I update  user `user.setDeleted(true)`, the change applies automatically in the db. Maybe it's related with hibernate configuration.
getUserByName will retrieve non soft deleted user. repository.findByUsername will retrieve user whether soft deleted or no. getUserByNameAndIsDelete with retrieve only soft deleted user.

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
SteveGT96: Nice catch !

src/main/java/org/isf/menu/service/MenuIoOperations.java
SteveGT96: Your've just said in your previous comments:

> I don't think we should disable this button, because one may need to finally hard-delete the group

> I don't think we should disable buttonEdit and buttonDelete buttons, because one may need to restore the group or hard-delete it (once empty), hence we need to update the GroupEdit window with the new checkbox

src/main/java/org/isf/menu/service/MenuIoOperations.java
SteveGT96: Yes !

src/main/java/org/isf/menu/service/UserIoOperationRepository.java
SteveGT96: Nice catch !

src/main/java/org/isf/menu/service/MenuIoOperations.java
mwithi: I don't think it worth it... when you create new entities, booleans are false by default, and your suggestion is just adding a bug that will be also difficult to spot when an admin wants to create an object directly in deleted status (e.g. for tests)

SteveGT96: Let me remove it then !

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -101,40 +128,43 @@ public User newUser(User user) throws OHServiceException {
...
-	 * @return {@code true} if the user has been updated, {@code false} otherwise.
+	 * @return the {@link User} that has been updated
 	 * @throws OHServiceException When failed to update user
 	 */
-	public boolean updateUser(User user) throws OHServiceException {
+	public User updateUser(User user) throws OHServiceException {
+		User oldUser = ioOperations.getUserByNameAndIsDeleted(user.getUserName());
+
+		if (oldUser != null && user.isDeleted()) {
+			throw new OHServiceException(new OHExceptionMessage("angal.common.denied.msg"));
```
mwithi:
```suggestion
			throw new OHServiceException(new OHExceptionMessage(MessageBundle.getMessage("angal.common.denied.msg")));
```

mwithi: Anyway I don't think this kind of message will be clear for the user since it is not even explained why the operation is denied...  if this new key "`angal.common.denied.msg`" is used only in such case when the object is already soft-deleted, I think we could be more explicit, like "`angal.userbrowser.alreadysoftdeleted.msg = The user is already set as deleted.`"

And, for GUI, we could prompt this message even without calling the manager as we already know if the user is already soft-deleted or not.

image

image

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -301,70 +329,80 @@ public void deleteGroup(UserGroup aGroup) throws OHServiceException {
...
+	 * @return the {@link UserGroup} that has been updated
+	 * @throws OHServiceException If failed to update group
 	 */
-	public boolean updateUserGroup(UserGroup userGroup) throws OHServiceException {
+	public UserGroup updateUserGroup(UserGroup userGroup) throws OHServiceException {
+
+		UserGroup group = ioOperations.findByCodeAndIsDeleted(userGroup.getCode());
+
+		if (group != null && userGroup.isDeleted()) {
+			throw new OHServiceException(new OHExceptionMessage("angal.common.denied.msg"));
```
mwithi:
```suggestion
			throw new OHServiceException(new OHExceptionMessage(MessageBundle.getMessage("angal.common.denied.msg")));
```

mwithi: and same proposal like for `updateUser` but of course with different message "`angal.usergroupbrowser.alreadysoftdeleted.msg = The group is already set as deleted.`"

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -362,25 +387,28 @@ public UserGroup newUserGroup(UserGroup userGroup, List<Permission> permissions)
...
 	 * @param userGroup - the {@link UserGroup} to update
-	 * @return {@code true} if the group has been updated, {@code false} otherwise.
+	 * @return the {@link UserGroup} that has been updated
 	 * @throws OHServiceException When failed to update user group
 	 */
-	public boolean updateUserGroup(UserGroup userGroup, List<Permission> permissions) throws OHServiceException {
+	public UserGroup updateUserGroup(UserGroup userGroup, List<Permission> permissions) throws OHServiceException {
+		UserGroup group = findByCode(userGroup.getCode(), true);
+		if (group.isDeleted() && userGroup.isDeleted()) {
+			throw new OHServiceException(new OHExceptionMessage("angal.common.denied.msg"));
```
mwithi: if the above proposal is accepted, this could be again ""angal.userbrowser.alreadysoftdeleted.msg = The user is already set as deleted."

```suggestion
			throw new OHServiceException(new OHExceptionMessage(MessageBundle.getMessage("angal.userbrowser.alreadysoftdeleted.msg")));
```

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -398,6 +426,20 @@ public boolean updateUserGroup(UserGroup userGroup, List<Permission> permissions
...
+		if (entity != null) {
+			throw new OHServiceException(new OHExceptionMessage("angal.common.denied.msg"));
+		}
+	}
+
+	public void ensureUserGroupNotDeleted(String code) throws OHServiceException {
+		UserGroup entity = groupRepository.findByCodeAndDeleted(code, true);
+		if (entity != null) {
+			throw new OHServiceException(new OHExceptionMessage("angal.common.denied.msg"));
+		}
```
mwithi: Same `MessageBundle.getMessage()` missing here and keys to be changed in case my proposal is accepted.

src/test/java/org/isf/menu/Tests.java
mwithi: Found that JPA [default for flush mode is AUTO](https://www.baeldung.com/spring-jpa-flush#:~:text=The%20two%20main%20flush%20modes,up%2Dto%2Ddate%20data.) means that JPA flushes the persistence context automatically.
If we set `entityManager.setFlushMode(FlushModeType.COMMIT);` in `OHCoreTestCase`, the test fails, along with 118 similar tests...
Doing so (setting COMMIT for all tests) we could enforce the codebase, although I'm not sure would be the best practise... wdyt?

SteveGT96: I'm not proud to have fixed tests that way too. We could address that issue in a refactoring task. Wdyt ?

mwithi: If we decide it worth, for sure

mwithi: Let's move the discussion under OP-1357

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
```diff
@@ -90,9 +87,39 @@ public User getUserByName(String userName) throws OHServiceException {
...
 	}
 
+	/**
+	 * Returns {@link User} from its username
+	 * @param username - the {@link User}'s username
+	 * @param withTrashed - Included soft deleted if set to true
+	 * @return {@link User}
+	 * @throws OHServiceException When error occurs
+	 */
+	public User getUserByName(String username, boolean withTrashed) throws OHServiceException {
```
dbmalkovsky: I would vote to change `withTrashed` to something like `withSoftDeletion`.    `trashed` doesn't really lend itself to both soft and hard deletes IMHO.

src/main/java/org/isf/menu/service/MenuIoOperations.java
```diff
@@ -79,53 +79,69 @@ public List<User> getUser() throws OHServiceException {
...
+
+	/**
+	 * Returns {@link User} from its username
+	 * @param username - the {@link User}'s username
+	 * @param withTrashed - Included soft deleted if set to true
+	 * @return {@link User}
+	 * @throws OHServiceException When error occurs
+	 */
+	public User getUserByName(String username, boolean withTrashed) throws OHServiceException {
+		return withTrashed ? repository.findByUserName(username) : getUserByName(username);
```
dbmalkovsky: same comment about `trashed` as above

dbmalkovsky (commented):
Left some comments

src/main/java/org/isf/menu/manager/UserBrowsingManager.java
mwithi: voted!

mwithi (approved):
LGTM. Thanks a lot!

PR #1444, opened 2024-11-08
OP-1157 use pattern matching for instanceof (JDK 16)

dbmalkovsky:
See OP-1157

Replace local variable declarations with pattern variables in `instanceof`; new with JDK 16

More compact code


mwithi (approved):
LGTM. Thanks a lot!

PR #1445, opened 2024-11-08
OP-1157 use record classes (JDK 16)

dbmalkovsky:
See OP-1157

Record classes focus on modeling immutable data rather than extensible behavior. Automatic implicit implementation of data-driven methods, such as equals() and accessors, helps to reduce boilerplate code.

Part of JDK-16


mwithi (approved):
LGTM. Thanks a lot!

PR #932, opened 2023-04-04
Refactoring

mohit10696:
Refactored Some of the classes and method for better code quality

PR #1451, opened 2024-11-19
Common Code Quality Improvement

mohit10696:
Hi @dbmalkovsky and @mwithi
This PR is a replica of https://github.com/informatici/openhospital-core/pull/932 . which got closed due to inactive from my side.

I am setting up my local for this project. You will most likely see more PR from my side in future.


src/main/java/org/isf/agetype/manager/AgeTypeBrowserManager.java
```diff
@@ -113,10 +113,12 @@ public AgeType getTypeByCode(String code) throws OHServiceException {
 	protected void validateAgeTypes(List<AgeType> ageTypes) throws OHDataValidationException {
 		List<OHExceptionMessage> errors = new ArrayList<>();
 		for (int i = 1; i < ageTypes.size(); i++) {
-			if (ageTypes.get(i).getFrom() <= ageTypes.get(i - 1).getTo()) {
+			boolean isOverlapping = ageTypes.get(i).getFrom() <= ageTypes.get(i - 1).getTo();
+			boolean isGapLargerThanOne = ageTypes.get(i).getFrom() - ageTypes.get(i - 1).getTo() > 1;
+			if (isOverlapping) {
```
dbmalkovsky, line 118: One of the first things I was taught during my many decades of writing commercial software is this "rule":

1. Write something once and leave it as is
2. Write something twice and if it is simple leave it as is; if it is complex then consider exacting it into a method and be willing to incur the overhead
3. Write something three times, then exact it into a method as the other head and garbage collection is worth the cost so there is consistency

Thus I find most/all the occurrences of extracting into local variables and methods unnecessary. 

I don't approve of the changes.

src/main/java/org/isf/dicom/manager/SourceFiles.java
```diff
@@ -134,12 +134,17 @@ private void loadDicomDir(FileDicom fileDicom, File sourceFile, int patient) thr
...
 				dicomLoader.setLoaded(filesLoaded);
-			} else if (!".".equals(value.getName()) && !"..".equals(value.getName())) {
+			} else if (isValidDirectoryName(value.getName())) {
 				loadDicomDir(fileDicom, value, patient);
 			}
 		}
 	}
 
+	private static boolean isValidDirectoryName(String directoryName) {
+		return ".".equals(directoryName) && "..".equals(directoryName);
```
dbmalkovsky, line 144: This change also introduced a new bug; the code should be 
```suggestion
		return !".".equals(directoryName) && !"..".equals(directoryName);
```

src/main/java/org/isf/exa/service/CommonExamIOOperations.java
```diff
@@ -0,0 +1,75 @@
+/*
```
dbmalkovsky, line 1: I don't see the point of this super class given that the code references two different repositories.   I don't agree with this change either.

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperations.java
```diff
@@ -231,6 +221,28 @@ protected void updateStockWardQuantity(MovementWard movement) throws OHServiceEx
 		}
 	}
 
+	/**
+	 * Creates a new {@link MedicalWard} from the specified {@link MovementWard}.
+	 * @param movement the movement ward.
+	 * @param qty the quantity.
+	 * @return the created medical ward.
+	 */
+	private MedicalWard createMedicalWard(MovementWard movement, float qty) {
```
dbmalkovsky, line 230: Yes there is a lot of common code but it is very straight forward and clear as to what is being set when.   I find I have to really study this new method to understand what is happening and when.   I don't think this is useful IMO.

src/main/java/org/isf/serviceprinting/manager/PrintManager.java
```diff
@@ -116,4 +116,4 @@ public void print(String filename, List<?> toPrint, int action) throws OHService
 			LOGGER.error(jrException.getMessage(), jrException);
 		}
 	}
-}
+}
```
dbmalkovsky, line 119: You will find that about half the code ends in a new line and the other half doesn't.   Making this kind of change is self-defeating as the next person will probably change it back.

dbmalkovsky (commented):
See my comments as to why I don't think these changes are needed.

mohit10696:
@dbmalkovsky, I am closing the request. I am trying to pick up something from JIRA, but I am not able to log in. Could you help me with that?

dbmalkovsky:
What is the URL you are using?  Is it [https://openhospital.atlassian.net/](https://openhospital.atlassian.net/)?

@mwithi is there something special that needs to be done?

mohit10696:
image

I am clicking on login here, and it landing me on the dashboard repeatedly.

mwithi:
@mohit10696 it says you are logged out. Please log in or create an Atlassian account first.
If not yet done, please have a look at the [Contribution Guidelines](https://openhospital.atlassian.net/wiki/spaces/OH/pages/6946820/Contribution+Guidelines) and start from [TODO (good-first-issues)](https://openhospital.atlassian.net/issues/?filter=10206).
Join also the [Slack](https://github.com/informatici/openhospital-core?tab=readme-ov-file#community) channel, so you can be supported by the whole community!

mohit10696:
I did look at the Contribution guidelines. But the problem is with JIRA or something. I cannot go into login page. 
Anyway, I am from Canada, and that might be something causing the issue.
See attached video for more ref.
[recording.webm](https://github.com/user-attachments/assets/0f2e59c6-af7f-4688-a048-1a1f912cbbd6)

mwithi:
It might be cache or cookies problem. Try with "InPrivate" mode (SHIFT+CTRL+N) and if it works, then it's a cache problem.

mohit10696:
Worked !!! Thank you @mwithi

PR #1452, opened 2024-11-19
OP-1361 | Fix getMedicalsWard method

ArnaudFofou:
See OP-1361.


mwithi (changes requested):
Please add a script to adjust MINVT_WRD_ID_A varchar(1) NULL and test that module too.
Thanks!

sql/step_a113_alter_table_medicalinventory.sql
```diff
@@ -0,0 +1,12 @@
...
+ALTER TABLE OH_MEDICALDSRINVENTORY 
+DROP CONSTRAINT oh_medicaldsrinventory_ibfk_2;
+
+-- Step 2: Change the data type of the foreign key column
+ALTER TABLE OH_MEDICALDSRINVENTORY 
+MODIFY MINVT_WRD_ID_A CHAR(3);
+
+-- Step 3: Recreate the foreign key constraint
+ALTER TABLE OH_MEDICALDSRINVENTORY 
+ADD CONSTRAINT oh_medicaldsrinventory_ibfk_2
```
mwithi: Let's take the opportunity for renaming the FK with `FK_<THISTABLE>_<REFTABLE>`

And the same we can do for other `oh_medicaldsrinventory_ibfk_*` existing foreign keys

```suggestion
ADD CONSTRAINT FK_MEDICALDSRINVENTORY_WARD
```

mwithi: To avoid duplications we can add `FK_<THISTABLE>_<REFTABLE>_<OTHERMNEMONIC>`

sql/step_a113_alter_table_medicalinventory.sql
```diff
@@ -0,0 +1,12 @@
+-- Step 1: Drop the foreign key constraint
+ALTER TABLE OH_MEDICALDSRINVENTORY 
+DROP CONSTRAINT oh_medicaldsrinventory_ibfk_2;
+
+-- Step 2: Change the data type of the foreign key column
+ALTER TABLE OH_MEDICALDSRINVENTORY 
+MODIFY MINVT_WRD_ID_A CHAR(3);
```
mwithi: Although optional I would specify the object for better readability

```suggestion
ALTER TABLE OH_MEDICALDSRINVENTORY 
MODIFY COLUMN MINVT_WRD_ID_A CHAR(3);
```

sql/step_04_all_following_steps.sql
mwithi, line 1: Could you please update demo data too?

src/main/java/org/isf/medicalstockward/manager/MovWardBrowserManager.java
```diff
@@ -85,7 +85,7 @@ protected void validateMovementWard(MovementWard mov) throws OHDataValidationExc
 	 * @return the retrieved medicals.
 	 * @throws OHServiceException
 	 */
-	public List<MedicalWard> getMedicalsWard(char wardId, boolean stripeEmpty) throws OHServiceException {
+	public List<MedicalWard> getMedicalsWard(String wardId, boolean stripeEmpty) throws OHServiceException {
```
mwithi, line 88: Please fix also API by the moment is using this method

sql/step_04_all_following_steps.sql
ArnaudFofou: I will update which information please!!

mwithi: Sorry, forgot we don't have anymore DDL in demo data.

ArnaudFofou: okay

sql/step_a113_alter_table_medicalinventory.sql
mwithi: > And the same we can do for other oh_medicaldsrinventory_ibfk_* existing foreign keys
> To avoid duplications we can add FK_<THISTABLE>_<REFTABLE>_<OTHERMNEMONIC>

My suggestion was beyond `oh_medicaldsrinventory_ibfk_2`

mwithi (approved):

PR #1453, opened 2024-11-19
OP-1362 | update file step_a104_add_inventory_in_pharmacy.sql

ArnaudFofou:


mwithi:
to be done in https://github.com/informatici/openhospital-core/pull/1452

PR #1454, opened 2024-11-19
Chore: redundant operation on java.time object

dbmalkovsky:
If you look at the implementation of `LocalDataTime.from()' you will find:

<pre>
  public static LocalDateTime from(TemporalAccessor temporal) {
        if (temporal instanceof LocalDateTime) {
            return (LocalDateTime) temporal;
        } else if (temporal instanceof ZonedDateTime) {
</pre>

So the existing line:
<pre>
LocalDateTime firstPay = LocalDateTime.from(today);
</pre>

would take the first if-branch and just return the parameter passed in.

So there is no need to call the method as `today` is a `LocalDateTime` object.


mwithi (approved):
LGTM. Thanks a lot!

PR #1465, opened 2024-12-10
OP-1375 Remove invalid demo data

mwithi:
See OP-1375.

Cc: @ArnaudFofou


dbmalkovsky (commented):
LGTM; checked many of them and yes they should be removed.

dbmalkovsky (approved):
LGTM; checked many of them and yes they should be removed.

PR #1467, opened 2024-12-20
OP-1199 Small improvements

mwithi:
See OP-1199.
To support changes in https://github.com/informatici/openhospital-gui/pull/2107


dbmalkovsky (dismissed):
LGTM

PR #1468, opened 2025-01-06
OP-1348 | Fix error on Laboratory report

ArnaudFofou:
See OP-1348.


src/main/java/org/isf/serviceprinting/manager/PrintLabels.java
```diff
@@ -41,27 +42,21 @@ public class PrintLabels {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(PrintLabels.class);
```
mwithi: Seems now LOGGER is no longer needed.

src/main/java/org/isf/serviceprinting/manager/PrintLabels.java
```diff
@@ -41,27 +42,21 @@ public class PrintLabels {
...
```
mwithi:
```suggestion

```

src/main/java/org/isf/serviceprinting/manager/PrintLabels.java
```diff
@@ -27,10 +27,11 @@
 import java.util.Map;
 
 import org.isf.utils.db.DbSingleJpaConn;
-import org.isf.utils.exception.OHServiceException;
+import org.isf.utils.exception.OHException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
mwithi:
```suggestion
import org.isf.utils.exception.OHException;
```

mwithi (approved):
LGTM. Thanks a lot!

PR #1469, opened 2025-01-14
OP-1199 Fix reports (revert openpdf exclusion)

mwithi:
Reverting part of https://github.com/informatici/openhospital-core/pull/1467


dbmalkovsky (approved):
LGTM

PR #1479, opened 2025-02-17
OP-1380: deprecate Lot constructor without Medical

dbmalkovsky:
See OP-1380

There 4 uses of the deprecated method in GUI

- WardPharmacyRectify
- MovStockMultipleDischarging
- MovStockMultipeCharging (twice)


dbmalkovsky:
Instead of just deprecating the method, add a new method that includes `Medical` and make the corresponding changes in GUI.    See https://github.com/informatici/openhospital-gui/pull/2131

mwithi (approved):
LGTM. Thanks a lot!

PR #1480, opened 2025-03-03
OP-1332 OP-1199 OP-1200 Pharmacy Inventory feature

mwithi:
See [OP-1199](https://openhospital.atlassian.net/browse/OP-1199) and [OP-1200](https://openhospital.atlassian.net/browse/OP-1200) (also addresses [OP-1332](https://openhospital.atlassian.net/browse/OP-1332)).

[OP-1199]: https://openhospital.atlassian.net/browse/OP-1199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1200]: https://openhospital.atlassian.net/browse/OP-1200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1332]: https://openhospital.atlassian.net/browse/OP-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -231,22 +239,38 @@ public MedicalInventory getInventoryByReference(String reference) throws OHServi
...
 		}
 		if (reference == null || reference.equals("")) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.inventory.mustenterareference.msg")));
 		}
+		/*
+		 * Despite MINVT_REFERENCE is VARCHAR(50), this field will be copied into MMV_REFNO and concatenate to either "-charge" or "-discharge" suffixes, hence
+		 * the limit to 40. For similarity, we keep the same validation method for "ward" type inventories, despite the field would be copied into MMVN_DESC
+		 * VARCHAR(100) in those cases.
+		 */
+		if (reference.length() > 40) {
```
dbmalkovsky: Line 256 checks `reference` for `null` so at this line it could be `null` thus throwing a NPE.  Shouldn't the be in a nested `else` clause?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -617,24 +703,30 @@ public MedicalInventory actualizeMedicalInventoryRow(MedicalInventory inventory)
...
 		for (Lot lot : uniqueLotList) {
-			String lotCodeOfMovement = lot.getCode();
+			String lotCode = lot.getCode();
 			Medical medical = lot.getMedical();
 			Integer medicalCode = medical.getCode();
 			// Fetch also empty lots because some movements may have discharged them completely
-			Optional<Lot> optLot = movStockInsertingManager.getLotByMedical(medical, false).stream().filter(l -> l.getCode().equals(lotCodeOfMovement))
+			Optional<Lot> optLot = movStockInsertingManager.getLotByMedical(medical, false).stream().filter(l -> l.getCode().equals(lotCode))
 				.findFirst();
 			double mainStoreQty = optLot.get().getMainStoreQuantity();
```
dbmalkovsky: Should there be a `optLot.isPresent()` or `!optLot.isEmpty()` call before accessing `optLot` here?  Line 723 says the value is `Optional`

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -318,11 +347,12 @@ public void validateMedicalInventoryRow(MedicalInventory inventory, List<Medical
 				double theoQty = medicalInventoryRow.getTheoreticQty();
 				if (mainStoreQty != theoQty) {
 					lotUpdated = true;
+					double difference = mainStoreQty - theoQty;
 					medDescriptionForLotUpdated
 						.append("\n")
```
dbmalkovsky, line 353: in this file all the  string versions of `"\n"` should be replaced with the character version `'\n'`

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -542,66 +617,77 @@ public List<Movement> confirmMedicalInventoryRow(MedicalInventory inventory, Lis
...
 		// get general info
 		Ward selectedWard = wardManager.findWard(inventory.getWard());
 		LocalDateTime now = TimeTools.getNow();
-		String reason = MessageBundle.getMessage("angal.inventory.reason.msg").concat(inventory.getInventoryReference());
+		String reason = inventory.getInventoryReference();
+		List<MovementWard> insertedMovements = new ArrayList<>();
 		for (MedicalInventoryRow medicalInventoryRow : inventoryRowSearchList) {
 			double theoQty = medicalInventoryRow.getTheoreticQty();
 			double realQty = medicalInventoryRow.getRealQty();
 			Double movQuantity = theoQty - realQty;
```
dbmalkovsky: why isn't this just a primitive (`double`) instead of a wrapper type?

src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -260,7 +262,8 @@ void testIoGetMedicalInventory() throws Exception {
 	void testMgrNewMedicalInventory() throws Exception {
 		Ward ward = testWard.setup(false);
 		MedicalInventory medicalInventory = testMedicalInventory.setup(ward, false);
-		MedicalInventory newMedicalInventory = medicalInventoryManager.newMedicalInventory(medicalInventory);
+		List<MedicalInventoryRow> medicalInventoryRows = new ArrayList<MedicalInventoryRow>();
```
dbmalkovsky:
```suggestion
		List<MedicalInventoryRow> medicalInventoryRows = new ArrayList<>();
```
Explicit type replaced with `<>`

dbmalkovsky (changes requested):
Left some changes that I believe are needed.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: good one, fixed... thanks!

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: Yes... fixed, thank!

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: Done, thank you!

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: no reason... changed, thanks!

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -231,21 +239,38 @@ public MedicalInventory getInventoryByReference(String reference) throws OHServi
...
+			 * MMVN_DESC VARCHAR(100) in those cases.
+			 */
+			if (reference.length() > 40) {
+				errors.add(new OHExceptionMessage(MessageBundle.formatMessage("angal.inventory.thereferenceistoolongmaxchars.fmt.msg", 40)));
+			}
+			if (checkReferenceExists) {
+				try {
+					checkReference(medicalInventory);
+				} catch (OHServiceException e) {
+					throw new OHDataValidationException(e.getMessages());
```
dbmalkovsky: Instead of adding the exception to `errors` which then would be caught in line 275; if the reference is too long (see line 265) the user will not see that message.   The question is how important is losing a message?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -231,21 +239,38 @@ public MedicalInventory getInventoryByReference(String reference) throws OHServi
...
-		if (medInventory.getInventoryDate() == null) {
+		String reference = medicalInventory.getInventoryReference();
+		if (medicalInventory.getInventoryDate() == null) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.inventory.pleaseinsertavalidinventorydate.msg")));
 		}
-		if (medInventory.getInventoryDate() != null && medInventory.getInventoryDate().isAfter(tomorrow)) {
+		if (medicalInventory.getInventoryDate() != null && medicalInventory.getInventoryDate().isAfter(tomorrow)) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.inventory.notdateinfuture.msg")));
 		}
 		if (reference == null || reference.equals("")) {
```
dbmalkovsky:
```suggestion
		if (reference == null || reference.isEmpty()) {
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,28 +409,54 @@ public void validateMedicalInventoryRow(MedicalInventory inventory, List<Medical
...
-		StringBuilder medDescriptionForLotUpdated = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewLot = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewMedical = new StringBuilder("\n"); // initial new line
+		StringBuilder medDescriptionForLotUpdated = new StringBuilder('\n'); // initial new line
+		StringBuilder medDescriptionForNewLot = new StringBuilder('\n'); // initial new line
+		StringBuilder medDescriptionForNewMedical = new StringBuilder('\n'); // initial new line
 		boolean lotUpdated = false;
 		boolean lotAdded = false;
 		boolean medicalAdded = false;
+		boolean dateUpdated = false;
```
dbmalkovsky:
```suggestion
		boolean dateUpdated;
```
The variable always has a value set so setting to `false` is redundant.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -257,21 +282,21 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
 	 */
-	public void validateMedicalInventoryRow(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList)
+	public void validateMedicalInventoryRow(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList, boolean allMedicals)
 		throws OHServiceException {
 		LocalDateTime movFrom = inventory.getInventoryDate();
 		LocalDateTime movTo = TimeTools.getNow();
-		StringBuilder medDescriptionForLotUpdated = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewLot = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewMedical = new StringBuilder("\n"); // initial new line
+		StringBuilder medDescriptionForLotUpdated = new StringBuilder('\n'); // initial new line
```
dbmalkovsky:
```suggestion
		StringBuilder medDescriptionForLotUpdated = new StringBuilder("\n"); // initial new line
```
I was wrong; `StringBuilder` takes a `string` argument and not a `char`.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -257,21 +282,21 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+		StringBuilder medDescriptionForNewLot = new StringBuilder('\n'); // initial new line
```
dbmalkovsky:
```suggestion
		StringBuilder medDescriptionForNewLot = new StringBuilder("\n"); // initial new line
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -257,21 +282,21 @@ private void validateMedicalInventory(MedicalInventory medInventory) throws OHDa
...
+		StringBuilder medDescriptionForNewMedical = new StringBuilder('\n'); // initial new line
```
dbmalkovsky:
```suggestion
		StringBuilder medDescriptionForNewMedical = new StringBuilder("\n"); // initial new line
```

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
```diff
@@ -372,28 +409,54 @@ public void validateMedicalInventoryRow(MedicalInventory inventory, List<Medical
...
+	public void validateMedicalWardInventoryRow(MedicalInventory inventory, List<MedicalInventoryRow> inventoryRowSearchList, boolean allMedicals)
+		throws OHServiceException {
 		LocalDateTime movFrom = inventory.getInventoryDate();
 		LocalDateTime movTo = TimeTools.getNow();
-		StringBuilder medDescriptionForLotUpdated = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewLot = new StringBuilder("\n"); // initial new line
-		StringBuilder medDescriptionForNewMedical = new StringBuilder("\n"); // initial new line
+		StringBuilder medDescriptionForLotUpdated = new StringBuilder('\n'); // initial new line
+		StringBuilder medDescriptionForNewLot = new StringBuilder('\n'); // initial new line
+		StringBuilder medDescriptionForNewMedical = new StringBuilder('\n'); // initial new line
```
dbmalkovsky: undo the change for lines 419-421; the character version is correct.  My Bad.

dbmalkovsky (dismissed):
A few small changes/suggestions.   This should be my last review so I'm saying it LGTM.

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: yeah, we could use `errors.addAll(e.getMessages());` instead... or just bypass all other messages because this is more important?

src/main/java/org/isf/medicalinventory/manager/MedicalInventoryManager.java
mwithi: ok, thanks!

dbmalkovsky (approved):
LGTM; given all the changes 😄

PR #1482, opened 2025-03-12
OH2-454 | Triage printing feature

ArnaudFofou:
See OH2-454.
Add method to print patient examination and add jasper file


mwithi (commented):
Is this PR duplicating in CORE some reports from GUI?

ArnaudFofou:
> Is this PR duplicating in CORE some reports from GUI?

yes to use existing report on the UI without create new one. I think we can create task to migrate all the report from gui to the core.

mwithi:
> > Is this PR duplicating in CORE some reports from GUI?
> 
> yes to use existing report on the UI without create new one. I think we can create task to migrate all the report from gui to the core.

That task is already there... see OP-1070, that is blocked by OP-1069 and, as you can see, it was part of a wider plan to have OH2-196 too, and devops is also impacted (Transifex -> GitHub repos change).
I don't like hammering things just to make them fit, like duplicating code or binary files, although sometimes could be needed for a transition.
Please verify with your team:
- the effect on GUI of this change (will the GUI work the same way with duplicated reports?) for a transition
- the effort to properly move all the logic from GUI to CORE for all reports and a plan, if this PR is approved, to fast move to that new set up.

In any case, I would not merge this change into develop before the next release (before May) as it deeply change how the things work now.

PR #1484, opened 2025-03-17
Chore: remove unused method param

dbmalkovsky:
The parameter wasn't used in the method.

Rebuilt the jar and then rebuilt GUI and API without error.


mwithi (approved):
LGTM. Thanks a lot!

PR #1486, opened 2025-03-18
Chore: remove unused constructor method

dbmalkovsky:
The constructor was only used in unit tests.

Rebuilt the jar and did `mvn clean test` for GUI and API repos without error.


mwithi (approved):
LGTM. Thanks a lot!

PR #1487, opened 2025-03-18
Chore: use equals() for boxed types

dbmalkovsky:
It is wrong to use reference equality (== or !=) for java.lang.Integer (and other boxed types and strings) because it is not comparing actual values but locations in memory.


mwithi (approved):
LGTM. Thanks a lot!

PR #1504, opened 2025-04-19
Chore: add missing table to delete data script

dbmalkovsky:
Found this when debugging moving the version of SpringBoot


mwithi (approved):
LGTM. Thanks a lot!

PR #1519, opened 2025-07-31
Chore: Bump spring-boot-starter-parent 3.5.4

dbmalkovsky:
Also disabled one test that was producing inconsistent results which appear to be timing based.   I tried to fix it but was  unable so I disabled the Test for the time being.

This replaces https://github.com/informatici/openhospital-core/pull/1513

### DIFF
--- a/src/test/java/org/isf/medicalsinventory/Tests.java
+++ b/src/test/java/org/isf/medicalsinventory/Tests.java
@@ -75,6 +75,7 @@ import org.isf.ward.model.Ward;
 import org.isf.ward.service.WardIoOperationRepository;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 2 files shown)


src/test/java/org/isf/medicalsinventory/Tests.java
```diff
@@ -938,6 +940,17 @@ void testValidateMedicalWardInventoryRow() {
 		});
 		// Test if exception is OHDataValidationException instance
 		assertThat(throwable).isInstanceOf(OHDataValidationException.class);
+		//  Debugging code
```
mwithi, line 943:
```suggestion
		//  FIXME: Debugging code
```

mwithi, line 943: @ArnaudFofou I think we should come back to this when we have time... created OP-1392

mwithi (approved):
LGTM. Thanks a lot!

PR #1528, opened 2025-09-18
Chore: improve log when the language code is not found

mwithi:
If the language code is not found the log was not giving any clue about it


dbmalkovsky (approved):
LGTM

PR #1535, opened 2025-11-15
OP-1394 Fix second payment saving in open bills

mwithi:
See OP-1394.

- fix also deleteWhereId query for billItems


dbmalkovsky (approved):
LGTM

PR #1564, opened 2026-03-17
OP-1401 Add support for direct TXT printing for thermal POS printers

mwithi:
See OP-1401

### DIFF
--- a/src/main/java/org/isf/serviceprinting/manager/PrintReceipt.java
+++ b/src/main/java/org/isf/serviceprinting/manager/PrintReceipt.java
@@ -126,9 +126,10 @@ public class PrintReceipt {
 	private void printFileTXT(String file, boolean showDialog) {
 
 		PrintService printService;
+		DocFlavor flavor = DocFlavor.BYTE_ARRAY.AUTOSENSE;
+
 		if (showDialog) {
 			PrintRequestAttributeSet pras = new HashPrintRequestAttributeSet();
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/serviceprinting/manager/PrintReceipt.java
```diff
@@ -112,7 +108,74 @@ public PrintReceipt(JasperPrint jasperPrint, String fileName) {
...
+			printService = ServiceUI.printDialog(null, 200, 200, printServices, defaultService, flavor, pras);
+		} else {
+			printService = defaultPrintService;
+		}
+		if (printService == null) {
+			return;
+		}
+
+		DocPrintJob job = printService.createPrintJob();
+		DocFlavor flavor = DocFlavor.BYTE_ARRAY.AUTOSENSE;
```
dbmalkovsky: If you are going to assign a constant to a local variable '`flavor` why do it twice?  See lines 131 and 143.  Why not just move this to the top of the method and do it once?

dbmalkovsky (changes requested):

src/main/java/org/isf/serviceprinting/manager/PrintReceipt.java
mwithi: Good catch! Fixed, thank you!

dbmalkovsky (dismissed):
LGTM

PR #1569, opened 2026-03-26
Chore bump versions of openfeign and feign-gson

dbmalkovsky:
This is an intermediate fix for updating openfeign and feign-gson.

This is the same update as https://github.com/informatici/openhospital-core/pull/1568
That is feign-gson moves from 13.6 to 13.11

And updates this https://github.com/informatici/openhospital-core/pull/1551 but not all the way to version 5.
That is spring-cloud-starter-openfeign moves from 4.0.6 to 4.3.1 (the last version of the 4.x.x track) instead the major update to version 5.

Once this is there we can address the class incompatibility in the compile step.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-openfeign</artifactId>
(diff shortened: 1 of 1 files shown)


dbmalkovsky:
closing as I need to re-evaluate these changes.....

PR #1573, opened 2026-03-31
Chore bump versions of openfeign and feign-gson

dbmalkovsky:
This is an intermediate fix for updating openfeign and feign-gson.

This is the same update as https://github.com/informatici/openhospital-core/pull/1568
That is feign-gson moves from 13.6 to 13.11 and added fiegn-core to resolve a compile dependency.

And updates this https://github.com/informatici/openhospital-core/pull/1551 but not all the way to version 5.
That is spring-cloud-starter-openfeign moves from 4.0.6 to 4.3.1 (the last version of the 4.x.x track) instead the major update to version 5.

Once this is there we can address the class incompatibility in the compile step. This is an intermediate fix for updating openfeign and feign-gson.

It made more sense then trying to fix everything in one PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-openfeign</artifactId>
-			<version>4.0.6</version>
+			<version>4.3.1</version>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


mwithi:
Waiting for https://github.com/informatici/openhospital-core/pull/1579.

mwithi (approved):
LGTM. Thanks a lot!

PR #1578, opened 2026-04-14
OP-1405 Fix bill amount and bill balance recomputation upon save

mwithi:
See OP-1405.

Add recalculateTotals() and validateClosedBillBalance() methods:
- with modified and new tests
- with script for previous fixes


dbmalkovsky (approved):
LGTM with light testing.

BTW, I did not test the sql scripts. Sorry.

PR #1579, opened 2026-04-14
OP-1406 Fix TEL_SW data collector fails with PersistenceException on Hibernate 6+

mwithi:
See OP-1406.
Also improved DEBUG mode log.


dbmalkovsky (approved):
LGTM

PR #1582, opened 2026-04-24
OP-1354 Make SMS sender failure test deterministic

mwithi:
See OP-1354.

### DIFF
--- a/src/test/java/org/isf/sms/Tests.java
+++ b/src/test/java/org/isf/sms/Tests.java
@@ -378,9 +378,9 @@ class Tests extends OHCoreTestCase {
 		sms.setSmsNumber("+1320241494");
 		sms.setSmsText("SomeText");
 		SmsSenderOperations failingSmsSenderOperations = new SmsSenderOperations(new ConfiguredSmsGatewayEnvironmentStub(), List.of(new FailingSmsGateway()));
-		assertThat(failingSmsSenderOperations.initialize()).isTrue();
+		assertTrue(failingSmsSenderOperations.initialize());
 		failingSmsSenderOperations.preSMSSending(sms);
-		assertThat(failingSmsSenderOperations.sendSMS(sms)).isFalse();
+		assertFalse(failingSmsSenderOperations.sendSMS(sms));
 	}
 
 	@Test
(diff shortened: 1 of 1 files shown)


src/test/java/org/isf/sms/Tests.java
```diff
@@ -371,16 +372,16 @@ void testSmsSenderOperationsSendSMSNoGateway() throws Exception {
...
-//		smsSenderOperations.preSMSSending(sms);
-//		assertThat(smsSenderOperations.sendSMS(sms)).isFalse();
-//	}
+	@Test
+	void testSmsSenderOperationsSendSMSFailure() throws Exception {
+		Sms sms = testSms.setup(true);
+		sms.setSmsNumber("+1320241494");
+		sms.setSmsText("SomeText");
+		SmsSenderOperations failingSmsSenderOperations = new SmsSenderOperations(new ConfiguredSmsGatewayEnvironmentStub(), List.of(new FailingSmsGateway()));
+		assertThat(failingSmsSenderOperations.initialize()).isTrue();
```
dbmalkovsky:
```suggestion
		assertTrue(failingSmsSenderOperations.initialize());
```

src/test/java/org/isf/sms/Tests.java
```diff
@@ -371,16 +372,16 @@ void testSmsSenderOperationsSendSMSNoGateway() throws Exception {
...
+		failingSmsSenderOperations.preSMSSending(sms);
+		assertThat(failingSmsSenderOperations.sendSMS(sms)).isFalse();
```
dbmalkovsky:
```suggestion
		assertFalse(failingSmsSenderOperations.sendSMS(sms));
```

src/test/java/org/isf/sms/Tests.java
dbmalkovsky: See line 190 also; same suggestion.

src/test/java/org/isf/sms/Tests.java
dbmalkovsky: See line 195 for same suggestions.

dbmalkovsky (dismissed):
Small suggestions to make code read better IMHO.  They can be ignored.
Otherwise LGTM

mwithi:
> Small suggestions to make code read better IMHO. They can be ignored. Otherwise LGTM

Sorry, I've reverted the last changes because I noticed that we mostly use the assertThat(...) style in the other tests. Even though the .initialize() method returns a boolean, I think it's preferable to keep the same pattern for consistency, wdyt?

dbmalkovsky:
I personally like the shorter and more specific form (and so does some of my code review tools) but I can live with almost anything in tests.

dbmalkovsky (approved):
LGTM

PR #1592, opened 2026-05-17
OP-214: Switch report bundle loading from classpath to filesystem

4bh1gn4:
## Summary
Updates how JasperReports loads language-specific .properties files, switching from classpath-based ResourceBundle.getBundle() to direct filesystem loading.

## Changes
- Updated addReportBundleParameter in JasperReportsManager to load .properties files from filesystem
- Added legacy fallback that can be removed once all installations reflect file migration

#### Depends on openhospital-gui PR
Link: https://github.com/informatici/openhospital-gui/pull/2196

PR #1593, opened 2026-05-29
OP-214: Fix report bundle fallback for EN language

mwithi:
See OP-214.

As continuation of #1592 we are going to fix the "en" fallback as Transifex doesn't need `en/` folder.

### DIFF
--- a/src/main/java/org/isf/stat/manager/JasperReportsManager.java
+++ b/src/main/java/org/isf/stat/manager/JasperReportsManager.java
@@ -1030,6 +1030,7 @@ public class JasperReportsManager {
 	}
 
 	private void addReportBundleParameter(String jasperParameter, String jasperFileFolder, String jasperFileName, Map<String, Object> parameters) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/stat/manager/JasperReportsManager.java
```diff
@@ -1034,36 +1030,32 @@ private String getPatientPhotoFile(String patID) {
 	}
 
 	private void addReportBundleParameter(String jasperParameter, String jasperFileFolder, String jasperFileName, Map<String, Object> parameters) {
-		String language = new Locale(GeneralData.LANGUAGE).getLanguage();
+		String language = GeneralData.LANGUAGE;
```
mwithi, line 1034:
```suggestion
		// This is needed because new Locale() returns lowercase codes (eg. "am_et") and this would fail the path case matching
		String language = GeneralData.LANGUAGE;
```

dbmalkovsky (approved):
LGTM

PR #1594, opened 2026-06-16
OP-219 Honor a preselected series when loading a DICOM file

giuseppesorge:
## What
Supports loading a DICOM image into an **already existing** series instead of always creating a new one.

Part of [OP-219](https://openhospital.atlassian.net/browse/OP-219). The user-facing GUI change is in informatici/openhospital-gui (companion PR, same branch name).

## Why
In `SourceFiles`, the DICOM branch read `seriesNumber` and `seriesInstanceUID` straight from the file and overwrote any value already set on the `FileDicom`. So when the GUI preselected an existing series, the choice was silently ignored for actual `.dcm` files (it worked only for plain images, whose branch already honored a preset series number).

## How
Derive the series number / instance UID from the DICOM file **only when no series was preselected** — detected by an empty `seriesInstanceUID` (never set by `preLoadDicom`, so it is a clean override signal). Default behaviour (single file and multi-series folders) is unchanged.

## Testing
- Existing core DICOM test suites pass (`TestSqlDicomManager`, `TestFileSystemDicomManager`, 38 tests).
- Verified end-to-end with the GUI and two real `.dcm` files: choosing the existing series loads both images into a single series (one `DM_FILE_SER_INST_UID`, 2 rows) instead of two.

[OP-219]: https://openhospital.atlassian.net/browse/OP-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/dicom/manager/SourceFiles.java
+++ b/src/main/java/org/isf/dicom/manager/SourceFiles.java
@@ -389,9 +389,13 @@ public class SourceFiles extends Thread {
 				studyUID = attributes.getString(Tag.StudyInstanceUID) == null ? studyUID : attributes.getString(Tag.StudyInstanceUID);
 				accessionNumber = attributes.getString(Tag.AccessionNumber) == null ? accessionNumber : attributes.getString(Tag.AccessionNumber);
 				studyDescription = attributes.getString(Tag.StudyDescription) == null ? studyDescription : attributes.getString(Tag.StudyDescription);
-				seriesUID = attributes.getString(Tag.SeriesInstanceUID) == null ? seriesUID : attributes.getString(Tag.SeriesInstanceUID);
-				seriesInstanceUID = attributes.getString(Tag.SeriesInstanceUID) == null ? seriesInstanceUID : attributes.getString(Tag.SeriesInstanceUID);
-				seriesNumber = attributes.getString(Tag.SeriesNumber) == null ? generateSeriesNumber(patient) : attributes.getString(Tag.SeriesNumber);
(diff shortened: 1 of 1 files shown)


giuseppesorge:
The build (17) failure is unrelated to this change. The only failing test is TestJsonFileConfigProvider.testJsonFileConfigProvider, which fetches a remote JSON config over HTTP (GeneralData.PARAMSURL) and asserts it contains oh_telemetry_url; that only holds on an HTTP 200. In this run the fetch returned 403, so the map was empty and the assertion failed — an environment/network flake.
  
I reproduced the same failure on a clean develop checkout (without this change), so it's not introduced here. This PR only touches SourceFiles (DICOM); all DICOM suites pass, including TestSourceFiles (12 tests), and the run is 2431 tests, 1 failure — that single network test. Recent develop runs are green.

  Could a maintainer re-run the failed job? Happy to adjust if anything else is needed.

giuseppesorge:
Hi @mwithi, gentle ping for a review — OP-219 (load a DICOM image into an existing series), which you reported. Verified end-to-end with real DICOM files. The failing check is an unrelated flaky network test (`TestJsonFileConfigProvider`, HTTP 403 on the remote config fetch), reproducible on a clean `develop` — could you re-run the failed job? Companions: gui informatici/openhospital-gui#2209, doc informatici/openhospital-doc#276. Thanks!

giuseppesorge:

mwithi (approved):
LGTM. Thanks a lot!

PR #1595, opened 2026-06-17
OP-1388 Log deleted pharmaceutical stock movements

giuseppesorge:
## OP-1388 — Log deleted pharmaceutical stock movements

Adds an audit trail for deleted pharmaceutical stock movements.

### Changes
- New log tables `OH_MEDICALDSRSTOCKMOV_LOG` and `OH_MEDICALDSRSTOCKMOVWARD_LOG` (SQL `step_a118`, registered in `step_04`), keeping a snapshot of the removed movement plus an optional `reason`.
- New entities `MovementLog` / `MovementWardLog` and their JPA repositories.
- `MovBrowserManager.deleteLastMovement(Movement, String reason)` and `MovWardBrowserManager.deleteLastMovementWard(MovementWard, String reason)` now save the audit record within the existing transaction, before deleting the movement (so it rolls back if the delete fails).
- Tests updated for the new signatures, with assertions that the deletion is logged together with the reason.

Paired PRs (same branch name, so the GUI CI cross-builds against this branch): informatici/openhospital-gui#2210, informatici/openhospital-doc#278

JIRA: https://openhospital.atlassian.net/browse/OP-1388


giuseppesorge:
The CI is red only because of `org.isf.generaldata.configProvider.TestJsonFileConfigProvider.testJsonFileConfigProvider`, which performs a live HTTP GET to the remote `PARAMSURL` and failed on an `HTTP 403`. This is a pre-existing, network-dependent flake unrelated to this PR (also seen on other PRs / a clean `develop`).

All changes in this PR are green, e.g. `org.isf.medicalstock.Tests` 633/633, `org.isf.medicalstockward.Tests` 55/55, `org.isf.operation.Tests` 69/69. The paired GUI PR informatici/openhospital-gui#2210 (cross-building this branch) is fully green too.

Could a maintainer kindly re-run the job? Thanks!

giuseppesorge:

src/main/java/org/isf/medicalstock/model/MovementLog.java
```diff
@@ -0,0 +1,179 @@
...
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
+
+import jakarta.persistence.AttributeOverride;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EntityListeners;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
```
dbmalkovsky, line 33: Please use the project import order; see `.ide-settings/eclipse/OpenHopsital-importorder` which lists the following (note it is in reverse order):
<pre>
6=
5=com
4=org
3=jakarta
2=javax
1=java
0=\#
</pre> 

Thus the `jakarta` imports should be before `com` and `org`.

I think there is at least on other file with this same issue.

giuseppesorge:
Fixed — moved the jakarta imports before the org imports in both MovementLog and MovementWardLog (the two files in this PR with jakarta imports), following the project import order. Thanks for the pointer.

mwithi (approved):
LGTM. Thanks a lot!

PR #1597, opened 2026-06-18
OP-1216 Convert FLOAT/DOUBLE money and quantity columns to DECIMAL/INT

giuseppesorge:
## OP-1216 — Convert FLOAT and DOUBLE to DECIMAL for exact precision

Converts the monetary and exact-quantity fields from binary floating point
to `java.math.BigDecimal` (DECIMAL(19,2)), and the pure unit-count fields to
`int`, to guarantee exact precision for financial and stock values.

### Changes
- **Models → BigDecimal**: `Price.price`, `Bill.amount/balance`, `BillItems.itemAmount`,
  `BillPayments.amount`, `MovementWard.quantity`, `MedicalWard.out_quantity`,
  `Lot.cost` (now annotated `precision=19, scale=2`), and
  `MedicalInventoryRow.theoreticQty/realQty` (fields initialized to `ZERO`,
  matching the `NOT NULL DEFAULT 0` columns).
- **Models → int**: `Medical` in/out/min quantity, `MedicalWard.in_quantity`.
- **Managers/services**: arithmetic reworked on `BigDecimal` — zero checks use
  `compareTo`/`signum`, monetary results keep `setScale(2, HALF_UP)`, and the price-list
  step rounding is reproduced exactly. The ward-stock repository out-quantity
  queries now take `BigDecimal` (no `double` round-trip on the DECIMAL column);
  the in-quantity queries take `int`. `MedicalInventoryManager` compares and
  adjusts inventory quantities in `BigDecimal` end-to-end, with no
  `doubleValue()`/`BigDecimal.valueOf(double)` round-trips left.
- **Validation**: a fractional ward-to-ward transfer is rejected on the incoming
  (integer) side, and `confirmMedicalInventoryRow` validates each theoretic/real quantity with
  `intValueExact` (main-store stock is integer-only), rejecting fractional or
  out-of-int-range values instead of silently truncating them — including
  fractional quantities whose difference happens to be whole.
- **SQL**: `sql/step_a121_decimal_precision.sql` (numbered after the merged
  OP-1099 step, per the agreed merge-order numbering), registered in
  `step_04_all_following_steps.sql`.
- Tests updated for the new types (`BigDecimal` fixtures, `isEqualByComparingTo`).

### Notes / decisions
- `MDSRWRD_IN_QTI` stays **INT** and `MDSRWRD_OUT_QTI`/`MMVN_MDSR_QTY` become
  **DECIMAL**, following the ticket's field list; fractional ward-to-ward
  transfers are rejected by validation instead of being truncated.
- `MDSR_INI_STOCK_QTI` is **not** converted: it is not in the ticket's field list
  and its field (`Medical.initialqty`) is `@Deprecated`; converting only the column
  would have left a double-over-int mismatch.
- The main stock-movement quantity (`MMV_QTY`, `OH_MEDICALDSRSTOCKMOV`) becomes
  `INT` as listed.

### Verification
- Full core test suite green (2458 tests).
- Applied the migration to a demo-data DB (old FLOAT/DOUBLE schema, ~1.5k prices,
  bills, stock movements) on MariaDB 10.6 with production flags: all columns
  converted to the expected DECIMAL/INT types with **no data degradation**
  (values numerically identical pre/post; no fractional values in any to-INT column).
- GUI end-to-end on the migrated demo DB: a ward inventory row saved with a
  fractional real quantity persists as an exact `1.50`; a ward rectification of
  `2.5` persists as an exact `-2.50` movement.

Companion PRs (same branch name for CI cross-build) are linked in a comment below.

JIRA: https://openhospital.atlassian.net/browse/OP-1216


giuseppesorge:
OP-1216 spans three repositories (same branch name `OP-1216-decimal-precision` so the gui/api CI cross-builds the core fork branch):
- core: https://github.com/informatici/openhospital-core/pull/1597
- api: https://github.com/informatici/openhospital-api/pull/586
- gui: https://github.com/informatici/openhospital-gui/pull/2212

giuseppesorge:
Note for reviewers: the red CI here is **not** caused by this change. The only failing test is `TestJsonFileConfigProvider:40` — the non-hermetic config-provider test that hits the live (now 403) remote endpoint, which fails on all open core PRs. It is fixed by #1598 (makes the test hermetic). This PR's own tests pass; CI should go green once #1598 is merged and this branch is rebased.

sql/step_a121_decimal_precision.sql
mwithi, line 1: @giuseppesorge did you test it on a demo dataset? are all quantities and prices automatically converted or do we need a migration script?

giuseppesorge: Yes — tested on a copy of the demo dataset (MariaDB 10.6). The `ALTER TABLE ... MODIFY COLUMN` statements in `step_a120` convert every existing value in place, so no separate data-migration script is needed. Results: all 13 columns became `DECIMAL(19,2)`/`INT`; every value was numerically identical pre/post (prices, bill amounts/balances, stock and ward movement quantities); and there were 0 fractional values in any of the to-INT columns (`MMV_QTY`, `MDSR_*_QTI`, `MDSRWRD_IN_QTI`), so nothing was rounded.

giuseppesorge: Yes — verified on a copy of a real demo database (MariaDB 10.6). All existing quantities and prices are converted automatically in place by the `ALTER TABLE ... MODIFY COLUMN` statements; no separate data-migration script is needed:

- 0 fractional values in every column that becomes `INT` (`MDSRWRD_IN_QTI`, `MMV_QTY`, `MDSR_IN_QTI`, `MDSR_OUT_QTI`, `MDSR_MIN_STOCK_QTI`), so the FLOAT→INT conversion loses no data.
- Values preserved (e.g. `SUM(OH_PRICES.PRC_PRICE)` identical before and after the migration).
- Resulting types as intended: money and OUT quantities → `DECIMAL(19,2)`, stock quantities → `INT`.

src/test/java/org/isf/priceslist/Tests.java
```diff
@@ -228,7 +229,7 @@ void testIoCopyListSteps() throws Exception {
 		// then:
 		Price copyPrice = priceIoOperationRepository.findAll().get(1);
 		assertThat(copyPrice.getId()).isEqualTo(id + 1);
-		assertThat(copyPrice.getPrice()).isCloseTo(Math.round(2.0 * price.getPrice() / 3.0) * 3L, within(0.10d));
+		assertThat(copyPrice.getPrice().doubleValue()).isCloseTo(Math.round(2.0 * price.getPrice().doubleValue() / 3.0) * 3L, within(0.10d));
```
mwithi: I think also here we should use `isEqualByComparingTo` and forget `doubleValue`

giuseppesorge:
Thanks @mwithi. Addressed in 57162b13: the `copyList` assertions (`testIoCopyList`, `testIoCopyListSteps`, `testMgrCopyList`, `testMgrCopyListSteps3`) and the `MedicalWard` quantity assertions now use `isEqualByComparingTo`, with the expected values computed with the same `BigDecimal` arithmetic as the production code (`multiply`/`divide` with `RoundingMode.HALF_UP`), and I removed the now-unused `within`/`offset` imports.

I'm now applying the corresponding adaptations in `gui` (the callers of the getters/setters that changed from `Double`/`double`/`float` to `BigDecimal`/`int`) and will follow up with the matching gui changes.

src/test/java/org/isf/priceslist/Tests.java
mwithi: @giuseppesorge what about this part?

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
mwithi: and these methods?

src/test/java/org/isf/priceslist/Tests.java
giuseppesorge: Already addressed in 57162b13 — `testIoCopyList`, `testIoCopyListSteps`, `testMgrCopyList` and `testMgrCopyListSteps3` now use `isEqualByComparingTo`; there is no `doubleValue()`/`isCloseTo` left anywhere in `priceslist/Tests.java`.

src/main/java/org/isf/medicalstockward/service/MedicalStockWardIoOperationRepository.java
giuseppesorge: Done in bc73d513 — `findQuantityInWardWhereMedical` and `findQuantityInWardWhereMedicalAndWard` now return `BigDecimal`, and I adapted the callers in `MedicalStockWardIoOperations` (this also removed a lossy `BigDecimal.valueOf(double)` when populating `MedicalWard.qty`).

giuseppesorge:
The inventory quantity path is now `BigDecimal` at the root: `MedicalInventoryRow.theoreticQty/realQty` converted (entity, manager arithmetic, `DECIMAL(19,2)` in `step_a121`) — commit 882cdc42, details in the updated description. Companion GUI adaptation in informatici/openhospital-gui#2212.

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (approved):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

mwithi (dismissed):
LGTM. Thanks a lot!

giuseppesorge:
Superseded by #1643 — identical tree (verified diff-empty), resubmitted on a clean linear branch because the approvals here kept being auto-dismissed by a review glitch ("the merge base changed after approval" with both refs static). Keeping this PR open only as review-history reference until #1643 is merged, then it can be closed.

mwithi:
Replaced by #1643

PR #1598, opened 2026-06-18
OP-1412 Send a custom User-Agent on the remote configuration fetch (fixes HTTP 403)

giuseppesorge:
## What & why

After the recent migration of the Open Hospital website, fetching the remote configuration file from `https://conf.open-hospital.org/oh-conf.json` (property `PARAMSURL`) returns **HTTP 403**. The migrated server rejects requests sent with the default Java `User-Agent` (or with no `User-Agent` at all), which is exactly what the config-provider code did. Same root cause, two symptoms:

- **App startup** logs `ERROR - Failed to fetch configuration URL. HTTP response code: 403` and the remote configuration is never loaded (silent fallback to an empty map).
- **CI** — `TestJsonFileConfigProvider` performs a live HTTP GET against `PARAMSURL`, so it now fails with 403 on every core PR.

Verified with curl against the live URL: the default `Java/17` User-Agent (and no User-Agent) → **403**, while any custom User-Agent (e.g. `OpenHospital/1.15.0`) → **200**.

## Changes

- **`JsonFileConfigProvider`** (the production path — `ConfigProviderFactory` selects it when `PARAMSURL` ends in `.json`): set `User-Agent: OpenHospital/<version>` on the `HttpURLConnection` before reading the response.
- **`ApiConfigProvider`** (the Feign path): add a request interceptor that sets the same `User-Agent` header.
- **`TestJsonFileConfigProvider`**: refactored to be hermetic with MockServer (no live network) — asserts the config is parsed, that the request carries the custom `User-Agent`, and that a non-200 response yields an empty configuration. Added a matching `User-Agent` assertion to **`TestApiConfigProvider`** so the Feign interceptor is covered too.

## Testing

- `configProvider` tests green (7/7), `mvn` build OK.
- Confirmed the User-Agent assertion **fails without the fix** (the request otherwise carries `Java/17...`, which the server rejects).
- Real end-to-end check against the live URL: the configuration is now fetched successfully (no 403).

Jira: OP-1412

### DIFF
--- a/src/test/java/org/isf/generaldata/configProvider/TestJsonFileConfigProvider.java
+++ b/src/test/java/org/isf/generaldata/configProvider/TestJsonFileConfigProvider.java
@@ -22,23 +22,78 @@
 package org.isf.generaldata.configProvider;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockserver.integration.ClientAndServer.startClientAndServer;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
 
 import java.util.Map;
 
 import org.isf.generaldata.GeneralData;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.model.MediaType;
 
 public class TestJsonFileConfigProvider {
 
+	private static final String CONFIG_JSON = """
+		{
+		  "default": {
+		    "oh_telemetry_url": "https://default-url.com"
+		  }
+		}""";
+
+	private ClientAndServer mockServer;
+
+	@BeforeEach
+	public void startServer() {
+		mockServer = startClientAndServer();
+	}
+
+	@AfterEach
+	public void stopServer() {
+		mockServer.stop();
+	}
+
 	@Test
 	void testJsonFileConfigProvider() throws Exception {
+		mockServer.when(request().withMethod("GET").withPath("/oh-conf.json"))
+			.respond(response().withStatusCode(200)
+				.withContentType(MediaType.APPLICATION_JSON)
+				.withBody(CONFIG_JSON));
+
 		GeneralData.initialize();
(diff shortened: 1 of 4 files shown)


giuseppesorge:
@mwithi tagging you for review — this is the OP-1412 config-provider 403 fix we discussed (custom User-Agent in both providers + hermetic `TestJsonFileConfigProvider`). I couldn't add you via the Reviewers field (fork contributor permissions), so flagging you here. Thanks!

src/test/java/org/isf/generaldata/configProvider/TestJsonFileConfigProvider.java
```diff
@@ -22,23 +22,78 @@
...
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.model.MediaType;
 
 public class TestJsonFileConfigProvider {
 
+	private static final String CONFIG_JSON = """
+		{
+		  "default": {
```
dbmalkovsky, line 42:
```suggestion
		  "anyCurrentVersion": {
```
By using something other than a "real" version number no one will be tempted to change it as the version numbers increase over time.

This will require changes to line 119:
`when(mockVersion.toString()).thenReturn("anyCurrentVersion");` 
and line 135:
`mockServer.verify(request().withMethod("GET").withPath("/test").withHeader("User-Agent", "OpenHospital/anyCurrentVersion"));`

dbmalkovsky (changes requested):
I would strongly suggest making the change outline so as not to cause confusion in the future.

giuseppesorge:
Thanks @dbmalkovsky — done. Replaced `1.15.0` with `anyCurrentVersion` in the config block, the mocked version, and the User-Agent assertion. Good call: it makes clear it's just a fixture and won't tempt anyone to bump it. Tests green.

dbmalkovsky (approved):
LGTM

giuseppesorge:
Worth prioritizing this one: in its current (non-hermetic) form `TestJsonFileConfigProvider` reaches the live remote config endpoint, which now returns **403** after the website migration — so it fails in CI on **every** open core PR, not just here. This PR makes the test hermetic (MockServer), so merging it should turn the core CI green across the board. I'll rebase the other open PRs once it lands.

PR #1599, opened 2026-06-18
OP-1353 Fix and re-enable testMedicalStockGets and testMedicalStockSets

giuseppesorge:
## What & why

Two unit tests in `org.isf.medicalstock.Tests` — `testMedicalStockGets` and `testMedicalStockSets` — were disabled (no `@Test`, marked `// TODO: fix setupTestMedicalStock`) and therefore never ran.

The root cause was the test fixture, not the production code:
- `setupTestMedicalStock(boolean usingSet)` ignored its `usingSet` flag (it always called `testMedicalStock.setup(medical, false)`) and, more importantly, **never persisted** the entities — it returned `medicalStock.getCode()` on an unsaved entity (so `0`), which made `checkMedicalStockIntoDb` fail its `findById(...)` not-null assertion.
- `TestMedicalStock.setup(Medical, usingSet)` had an empty `else` branch: the `usingSet == false` path created a `new MedicalStock()` with no fields set, despite the comment "Create medicalStock with all parameters".

## Changes (test-only)

- `TestMedicalStock`: the `usingSet == false` branch now builds the entity via its all-args constructor — `new MedicalStock(medical, balanceDate, balance, nextMovDate, days)` — matching the constructor-vs-setters convention used by `TestMovement` / `TestLot` / `TestMedical`.
- `Tests.setupTestMedicalStock`: now threads `usingSet` through and persists `medicalType → medical → medicalStock` (FK order, mirroring `setupTestMovement`) before returning the generated code.
- Both tests are re-enabled with `@Test` (plain `@Test` is correct — they don't depend on the `AUTOMATICLOT*` flags that drive the parameterized movement tests).

No production code changed.

## Testing

- `org.isf.medicalstock.Tests` is green: **635/635** (was 633 with the two tests disabled).
- Verified the re-enabled tests have teeth: removing the persistence calls makes both fail with `Expecting actual not to be null`, confirming they exercise the real save → `findById` round-trip.

Jira: OP-1353

### DIFF
--- a/src/test/java/org/isf/medicalstock/Tests.java
+++ b/src/test/java/org/isf/medicalstock/Tests.java
@@ -203,13 +203,13 @@ class Tests extends OHCoreTestCase {
 		checkMovementIntoDb(code);
 	}
 
-	// TODO: fix setupTestMedicalStock
+	@Test
 	void testMedicalStockGets() throws Exception {
 		int code = setupTestMedicalStock(false);
 		checkMedicalStockIntoDb(code);
 	}
 
-	// TODO: fix setupTestMedicalStock
+	@Test
 	void testMedicalStockSets() throws Exception {
 		int code = setupTestMedicalStock(true);
 		checkMedicalStockIntoDb(code);
@@ -1593,7 +1593,10 @@ class Tests extends OHCoreTestCase {
 	private int setupTestMedicalStock(boolean usingSet) throws OHException {
 		MedicalType medicalType = testMedicalType.setup(false);
 		Medical medical = testMedical.setup(medicalType, false);
-		MedicalStock medicalStock = testMedicalStock.setup(medical, false);
+		MedicalStock medicalStock = testMedicalStock.setup(medical, usingSet);
+		medicalTypeIoOperationRepository.saveAndFlush(medicalType);
+		medicalsIoOperationRepository.saveAndFlush(medical);
+		medicalStockIoOperationRepository.saveAndFlush(medicalStock);
 		return medicalStock.getCode();
 	}
 
--- a/src/test/java/org/isf/medicalstock/Tests.java
+++ b/src/test/java/org/isf/medicalstock/Tests.java
@@ -203,14 +203,18 @@ class Tests extends OHCoreTestCase {
 		checkMovementIntoDb(code);
 	}
 
-	@Test
-	void testMedicalStockGets() throws Exception {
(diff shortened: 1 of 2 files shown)


giuseppesorge:
@mwithi tagging you for review — OP-1353: re-enables the two disabled medical-stock fixture tests (test-only change, 635/635 green, teeth-verified). Couldn't set the Reviewers field as a fork contributor, so flagging you here. Thanks!

src/test/java/org/isf/medicalstock/Tests.java
```diff
@@ -203,13 +203,13 @@ void testMovementSets(boolean in, boolean out, boolean toward) throws Exception
 		checkMovementIntoDb(code);
 	}
 
-	// TODO: fix setupTestMedicalStock
+	@Test
```
dbmalkovsky:
```suggestion
		@ParameterizedTest(name = "Test with AUTOMATICLOT_IN={0}, AUTOMATICLOT_OUT={1}, AUTOMATICLOTWARD_TOWARD={2}")
	@MethodSource("automaticlot")
```
These tests are all parameterized and just using `@Test` doesn't try all the combinations.

src/test/java/org/isf/medicalstock/Tests.java
```diff
@@ -203,13 +203,13 @@ void testMovementSets(boolean in, boolean out, boolean toward) throws Exception
...
 	void testMedicalStockGets() throws Exception {
 		int code = setupTestMedicalStock(false);
 		checkMedicalStockIntoDb(code);
 	}
 
-	// TODO: fix setupTestMedicalStock
+	@Test
```
dbmalkovsky:
```suggestion
		@ParameterizedTest(name = "Test with AUTOMATICLOT_IN={0}, AUTOMATICLOT_OUT={1}, AUTOMATICLOTWARD_TOWARD={2}")
	@MethodSource("automaticlot")
```
Same  comment as for line 206

dbmalkovsky (changes requested):
Needs update to really test all the combinations.

giuseppesorge:
Thanks @dbmalkovsky — done. `testMedicalStockGets`/`testMedicalStockSets` are now `@ParameterizedTest` with `@MethodSource("automaticlot")` and call `setGeneralData(in, out, toward)` like the sibling tests, so they run across all the AUTOMATICLOT combinations (and reset the global flags for isolation). medicalstock suite green (649).

dbmalkovsky (approved):
LGTM

PR #1600, opened 2026-06-18
OP-1404 Add input validation to Operation and OperationRow managers

giuseppesorge:
## What & why

The two managers in `org.isf.operation.manager` had no `validate()` method, so invalid input was only caught (if at all) by DB constraints. This adds manager-level validation on save/update, following the established `org.isf.opetype.manager.OperationTypeBrowserManager.validateOperationType` pattern (`List<OHExceptionMessage>` + `OHDataValidationException`).

## Changes

- **`OperationBrowserManager.validateOperation(Operation, boolean insert)`** — called by `newOperation` (insert) and `updateOperation`. Checks: code empty / longer than 10 (`OPE_ID_A varchar(10)`) / already in use (on insert); description empty. Reuses the existing `angal.common.*` messages.
- **`OperationRowBrowserManager.validateOperationRow(OperationRow)`** — called by `newOperationRow`/`updateOperationRow`. `operation`, `prescriber` and `opDate` are mandatory; `opResult` and `remarks` stay **optional** and are only length-checked (max 250), because the GUI legitimately submits them empty (`OperationRowAdm`/`OperationRowOpd` set an empty result when none is selected). New `angal.operationrow.*` messages are added in the GUI bundle (companion PR).
- 3 new tests in `org.isf.operation.Tests`.

## Testing

- `org.isf.operation.Tests` green: **72/72** (was 69). The 3 new tests are regression-verified (they fail if the `validate*` calls are removed).
- Real-app check: launched the GUI, Settings → Operation → New with an empty code → the validation fires and the dialog shows the resolved message ("Please insert a code.").

## Companion PR

GUI i18n messages: informatici/openhospital-gui (same branch `OP-1404-operation-validation`).

Jira: OP-1404


giuseppesorge:
Companion GUI PR (i18n messages): https://github.com/informatici/openhospital-gui/pull/2213 — same branch, build together.

giuseppesorge:

dbmalkovsky:
@giuseppesorge If you update the branch so it is not out of date then the CI will re-run and things should be green again.

giuseppesorge:
Thanks @dbmalkovsky — and thanks for merging #1598 and #1599! You're right: I've updated this branch (and my other open core PRs) onto the latest `develop`, which now includes the hermetic `TestJsonFileConfigProvider`, so the CI should re-run green from here. 🙏

giuseppesorge:
Updated the branch with the latest develop, so CI should re-run and go green. Thanks.

src/main/java/org/isf/operation/manager/OperationBrowserManager.java
```diff
@@ -108,10 +111,38 @@ public Operation newOperation(Operation operation) throws OHServiceException {
 	 * @throws OHServiceException
 	 */
 	public Operation updateOperation(Operation operation) throws OHServiceException {
+		validateOperation(operation, false);
 		// the user has confirmed he wants to overwrite the record
 		return ioOperations.updateOperation(operation);
 	}
 
+	/**
+	 * Verify if the {@link Operation} is valid for CRUD and return a list of errors, if any.
```
dbmalkovsky:
```suggestion
	 * Verify if the {@link Operation} is valid for CRUD and throw an exception with the list of errors, if any.
```
The method is `void` so it doesn't directly return a list of errors.  Nitpick, I know.

giuseppesorge: Good catch -- fixed. The method is void, so I updated the Javadoc to "Verify if the Operation is valid for CRUD and throw an exception with the list of errors, if any." (commit 2c006083). Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #1596, opened 2026-06-17
OP-896 Add password lease (force password change) tracking to core

giuseppesorge:
## OP-896 — Password lease (core)

Adds the foundation for the "password lease" security feature (STRIDE S03).

### Changes
- `OH_USER`: new columns `US_PASSWD_LAST_CHANGED` and `US_PASSWD_MUST_CHANGE` (SQL `step_a119`, registered in `step_04`; existing users are initialized to the upgrade date so nobody is forced to change immediately).
- `User` entity: matching `passwdLastChanged` / `passwdMustChange` fields.
- `GeneralData.PASSWORDLEASE` config parameter (days; `0` = disabled).
- `updatePassword()` now resets the last-changed timestamp and persists the caller-provided must-change flag (an administrator reset forces a change at next login, a self change clears it).
- `UserBrowsingManager.isPasswordExpired()` reports when a password has outlived its lease.
- Unit tests for the new behaviour.

Part of the OP-896 multi-repo change (core + api + gui + ui + doc), all on branch `OP-896-password-lease`.

JIRA: https://openhospital.atlassian.net/browse/OP-896

PR #1601, opened 2026-06-18
OP-1099 Add JPA/DB uniqueness constraints to user-related entities

giuseppesorge:
## What & why

OP-1099: enforce data integrity on user-related entities at the JPA/DB level.

## Changes

- **Entities** (annotations matched to the *real* schema):
  - `User`: `userName` `@Column(US_ID_A, nullable=false, length=50)` (it is `@Id`); `userGroupName` `@JoinColumn(US_UG_ID_A, nullable=false)`; `passwd` `@Column(US_PASSWD, nullable=false, length=60)`.
  - `UserGroup`: `code` `@Column(UG_ID_A, nullable=false, length=50)`; `desc` `length=128`.
  - `Permission`: named unique on `P_NAME` via `@Table(uniqueConstraints=@UniqueConstraint(UX_PERMISSIONS_NAME, "P_NAME"))`, `name` `nullable=false, length=50`.
  - `GroupPermission`: composite `@UniqueConstraint(UX_GP_GROUP_PERM, {GP_UG_ID_A, GP_P_ID_A})`.
- **Bug fix** exposed by the new composite unique: `MenuIoOperations.updateUserGroup(...)` deleted then re-inserted group permissions, but Hibernate ordered the inserts before the deferred (derived) deletes, transiently violating the unique. Added a `flush()` after the delete.
- **Migration** `sql/step_a118_user_related_unique_constraints.sql` (registered): adds `UNIQUE(P_NAME)` first (fails fast, no destructive change, if duplicate permission names exist — they are NOT auto-removed since a permission may be referenced), then de-dups `(user group, permission)` rows and adds the composite unique.
- **Tests**: two new tests asserting `DataIntegrityViolationException` on a duplicate permission name and on a duplicate (group, permission) pair.

## Note on the ticket spec

The lengths/column names in the ticket description were inaccurate; the real schema was used instead: `UG_ID_A`=50 (not 10), `UG_DESC`=128 (not 255), `P_NAME`=50 (not 100), `US_PASSWD`=varchar(60); the permission column is `P_NAME` (not `PRM_NAME`) and the group-permission FK is `GP_P_ID_A` (not `GP_PRM_ID`). The migration follows the `sql/step_aNNN` convention rather than a `sql/upgrade/` script.

## Testing

- `menu`, `permissions`, `security` packages green: **92/92** (incl. the 2 new tests, regression-checked — they fail if the constraints are removed). Quality reviewed.

## Companion PR

Doc note: informatici/openhospital-doc (same branch `OP-1099-user-jpa-constraints`).

Jira: OP-1099


giuseppesorge:
Companion doc PR: https://github.com/informatici/openhospital-doc/pull/281 — same branch.

sql/step_a120_user_related_unique_constraints.sql
mwithi, line 1: Please remove the `oh` database selection from all queries and we use UPPERCASE tables names in scripts.

mwithi (dismissed):

sql/step_a120_user_related_unique_constraints.sql
giuseppesorge: Done in 0342b0ba — rewrote `step_a121` with unqualified UPPERCASE table names (dropped the `oh.` selector and the identifier backticks), matching the house style in `step_a115`. Re-verified on MariaDB 10.6: the two constraints are created and duplicate inserts are rejected (ERROR 1062).

dbmalkovsky (commented):
@giuseppesorge  

This looks good with a few comments:
(1) The column, "US_PASSWD" has length 60 and given the GUI checks for length 72 at first I was slightly confused.  Then I remembered that `BCrypt.hashpw()` takes a maximum of 72 characters and returns 60. Not sure if this is worthy of a note or not.
(2) Running the SQL in a Intellij console reports that the `DELETE gp1 FROM OH_GROUPPERMISSION gp1` is risky as it appears on the surface to delete records without a `WHERE` clause.  Is there some way to add an option around this query so as not to cause user confusion or require interaction?
(3) The usual numbering issue with the SQL statement is present ....Before the commit the step number in `source step_a121_user_related_unique_constraints.sql;` needs to be modified.

Other than that LGTM

giuseppesorge:
Thanks for the review — all three points addressed in b49a5858:

1. **US_PASSWD 60 vs 72** — added a note on the column: 60 is the BCrypt hash output length, while the GUI caps the raw password at 72 chars, the maximum BCrypt processes. Exactly the reconstruction you made.
2. **"Risky" DELETE** — moved the `GP_ID` condition from the join to an explicit `WHERE` clause (same semantics for an inner join), so consoles/linters no longer flag it as an unfiltered multi-table delete; also expanded the comment to state that already-unique rows are never touched. Re-verified on a demo-DB copy with a planted duplicate: the duplicate with the higher GP_ID is removed, both constraints apply, and re-inserting the pair fails with ERROR 1062.
3. **Step numbering** — renumbered to `step_a119` (the next free number on develop) since this PR merges first; OP-896 (#1596), which previously held a119, moved to a121 in the same pass so the pending block stays contiguous with no duplicates.

dbmalkovsky (dismissed):
LGTM

@mwithi guess this is now for you to re-review.

giuseppesorge:
Thanks for the review!

**1. US_PASSWD length (60 vs 72):** you're right to double-check. BCrypt truncates its input at 72 bytes and always returns a 60-character hash, so the column is correctly sized at 60; the GUI's 72 limit just guards the raw input before hashing. This PR doesn't touch that column, so I'd keep it out of scope here — happy to add a short schema comment in a follow-up if you think it's worth documenting.

**2. `DELETE gp1 FROM OH_GROUPPERMISSION gp1`:** good catch that it reads like an unconditional delete — it isn't. The `INNER JOIN` together with `WHERE gp1.GP_ID > gp2.GP_ID` bound it: a row is removed only when another row with the same (user group, permission) and a lower `GP_ID` exists, so on a clean database the statement deletes nothing. I've pushed a commit (`fdd089c0`) that spells this out in a comment right above the statement.

**3. SQL step numbering:** the step is now `step_a119`, which matches the merge order mwithi settled (OP-1099 → a119). Following his note I'll use the `step_aXXX` placeholder on future PRs and leave the final numbering to merge time.

Thanks!

giuseppesorge:
@dbmalkovsky @mwithi the last commit (the clarifying comment on the duplicate-cleanup DELETE) dismissed the prior approval — the PR is unchanged in behavior and ready for a re-review whenever you have a moment. Thanks!

giuseppesorge:
Thanks for the review!

- (3) Fixed: the migration step is now the `aXXX` placeholder (`step_aXXX_user_related_unique_constraints.sql` + its `source` line), so the final number is assigned at merge time — it had collided with the `a119` that came in when OP-1388 was merged into develop.
- (2) The `DELETE gp1 FROM OH_GROUPPERMISSION` is bounded: it self-joins on the same `(GP_UG_ID_A, GP_P_ID_A)` pair and removes a row only when another row with a lower `GP_ID` exists (`WHERE gp1.GP_ID > gp2.GP_ID`), so on a clean database it deletes nothing; the inline NOTE documents this. The "delete without WHERE" appearance is MySQL/MariaDB safe-update mode in an interactive console — the migration is applied via `source` in the updater, so there is no prompt or interaction.
- (1) `US_PASSWD` is 60 by design: BCrypt stores a 60-char hash while accepting up to 72 chars of input. Happy to add a one-line comment on the column if you think it is worth it.

dbmalkovsky:
@giuseppesorge if you change the file name to `a120` and update `step_04_all_following_steps` I will merge the request.... thanks

giuseppesorge:
Done in 52d249fe — renamed the migration to `step_a120_user_related_unique_constraints.sql` and updated the `source` line in `step_04_all_following_steps.sql`. Thanks @dbmalkovsky!

dbmalkovsky (dismissed):
LGTM

dbmalkovsky (approved):
LGTM

mwithi (approved):
Reviewed and approved by @dbmalkovsky

PR #607, opened 2022-02-18
Chore: combine all recent dependabot updates into one PR to test all of them together

dbmalkovsky:
Will rebase the other PRs when this is committed.

PR #1602, opened 2026-06-18
OP-892 Snapshot patient sex and age on each admission for anonymous statistics

giuseppesorge:
## What & why

OP-892 (Information Disclosure 3/3 — Separation of personal data, STRIDE I08), **core scope**. Replicate the patient's sex and age onto each admission so aggregate statistics can be produced without joining (and exposing) the patient — as already done for `Laboratory` and OPD.

## Changes

- `Admission`: new `ADM_SEX` (String, length 1) and `ADM_AGE` (Integer) columns + `applyPatientData()` that snapshots the linked patient's sex and **age at the admission date** (computed from birth date vs `ADM_DATE_ADM`).
- `AdmissionIoOperations.newAdmission`: calls `applyPatientData()` before saving (covers `newAdmissionReturnKey` and the manager, which funnel through `newAdmission`).
- The snapshot is **immutable**: it is intentionally not refreshed on patient merge (`AdmissionPatientMergedEventListener` documents this) — the patient's sex/age at the historical admission do not change. This is the deliberate difference from `Laboratory`, which stores current age on merge.
- Migration `sql/step_a119_admission_sex_age.sql` (registered): adds the two nullable columns and back-fills existing admissions (age computed at the admission date; `NULL` when the patient has no birth date).
- Test `testIoNewAdmissionStoresPatientSexAndAge` (parameterized like the sibling tests), asserting the stored age = age-at-admission and that it is persisted.

## Scope note

This PR covers the core schema/logic + migration + doc. The `.jrxml` statistical-report updates (removing the `OH_PATIENT` joins / a new aggregate report) are a larger, separate piece and are intentionally left as a **follow-up**.

## Note on the ticket spec

Used the real schema/pattern: Java fields follow `Laboratory` (`String sex` / `Integer age`); the migration uses `ADM_DATE_ADM` / `ADM_PAT_ID` (the ticket's `ADM_IN_DATE` does not exist) and the `sql/step_aNNN` convention.

## Testing

- `org.isf.admission.Tests` green (164), incl. the new test (teeth-verified: fails without the snapshot). Runtime age now matches the migration backfill (age-at-admission), so app-created and back-filled rows agree.

## Companion PR

Doc note: informatici/openhospital-doc (same branch).

Jira: OP-892


giuseppesorge:
Companion doc PR: https://github.com/informatici/openhospital-doc/pull/283 — same branch.

giuseppesorge:
Note for reviewers: the red CI here is **not** caused by this change. The only failing test is `TestJsonFileConfigProvider:40` — the non-hermetic config-provider test that hits the live (now 403) remote endpoint, which fails on all open core PRs. It is fixed by #1598 (makes the test hermetic). This PR's own tests pass; CI should go green once #1598 is merged and this branch is rebased.

sql/step_a122_admission_sex_age.sql
```diff
@@ -0,0 +1,15 @@
+-- OP-892: store a snapshot of the patient's sex and age on each admission (STRIDE I08), so that aggregate
+-- statistics can be produced without joining (and exposing) the patient's personal data.
+
+ALTER TABLE `oh`.`oh_admission`
```
dbmalkovsky:
```suggestion
ALTER TABLE  OH_ADMISSION
```
Better match the other alter scripts

sql/step_a122_admission_sex_age.sql
```diff
@@ -0,0 +1,15 @@
...
+	ADD COLUMN ADM_SEX CHAR(1) NULL,
+	ADD COLUMN ADM_AGE INT NULL;
+
+-- Backfill existing admissions with the patient's sex and the patient's age at the admission date.
+UPDATE `oh`.`oh_admission` a
```
dbmalkovsky:
```suggestion
UPDATE OH_ADMISSION a
```
Better match the other alter scripts

sql/step_a122_admission_sex_age.sql
```diff
@@ -0,0 +1,15 @@
...
+	INNER JOIN `oh`.`oh_patient` p ON a.ADM_PAT_ID = p.PAT_ID
```
dbmalkovsky:
```suggestion
	INNER JOIN OH_PATIENT p ON a.ADM_PAT_ID = p.PAT_ID
```

sql/step_04_all_following_steps.sql
```diff
@@ -116,3 +116,6 @@ source step_a119_log_deleted_pharmaceutical_movements.sql;
 source step_a120_user_related_unique_constraints.sql;
 source step_a121_decimal_precision.sql;
 source step_a122_password_lease.sql;
+source step_aXXX_admission_sex_age.sql;
+source step_aXXX_examination_sex_age.sql;
```
mwithi: The two can be combined into the same step I guess

giuseppesorge: Done in ee8de60f — the admission and examination columns + back-fills are now a single migration step (step_aXXX_sex_age_snapshot.sql).

sql/step_aXXX_sex_age_snapshot.sql
```diff
@@ -0,0 +1,29 @@
+-- OP-892: store a snapshot of the patient's sex and age on each admission and each patient examination
+-- (STRIDE I08), so that aggregate statistics can be produced without joining (and exposing) the patient's
+-- personal data.
+
+ALTER TABLE OH_ADMISSION
+	ADD COLUMN ADM_SEX CHAR(1) NULL,
+	ADD COLUMN ADM_AGE INT NULL;
```
mwithi: I would put them in different order if possible:

```suggestion
ALTER TABLE OH_ADMISSION
	ADD COLUMN ADM_SEX CHAR(1) NULL AFTER ADM_PAT_ID,
	ADD COLUMN ADM_AGE INT NULL AFTER ADM_SEX;
```

sql/step_aXXX_sex_age_snapshot.sql
```diff
@@ -0,0 +1,29 @@
...
+	INNER JOIN OH_PATIENT p ON a.ADM_PAT_ID = p.PAT_ID
+	SET a.ADM_SEX = p.PAT_SEX,
+		a.ADM_AGE = CASE
+			WHEN p.PAT_BDATE IS NOT NULL THEN TIMESTAMPDIFF(YEAR, p.PAT_BDATE, a.ADM_DATE_ADM)
+			ELSE NULL
+		END;
+
+ALTER TABLE OH_PATIENTEXAMINATION
+	ADD COLUMN PEX_SEX CHAR(1) NULL,
+	ADD COLUMN PEX_AGE INT NULL;
```
mwithi: Same here:

```suggestion
ALTER TABLE OH_PATIENTEXAMINATION
	ADD COLUMN PEX_SEX CHAR(1) NULL AFTER PEX_PAT_ID,
	ADD COLUMN PEX_AGE INT NULL AFTER PEX_SEX;
```

sql/step_aXXX_sex_age_snapshot.sql
giuseppesorge: Applied in `98df3861` — the admission columns are now added right after `ADM_PAT_ID` with the `AFTER` clauses. Thanks!

sql/step_aXXX_sex_age_snapshot.sql
giuseppesorge: Applied in `98df3861` — same for the examination columns, added right after `PEX_PAT_ID`. Thanks!

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`17ba0add`): added ADM_SEX/ADM_AGE and PEX_SEX/PEX_AGE to oh_admission and oh_patientexamination, back-filled from the patient sex and the age at admission/examination. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`17ba0add`): added ADM_SEX/ADM_AGE and PEX_SEX/PEX_AGE to oh_admission and oh_patientexamination, back-filled from the patient sex and the age at admission/examination. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

mwithi:
> @dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`17ba0add`): added ADM_SEX/ADM_AGE and PEX_SEX/PEX_AGE to oh_admission and oh_patientexamination, back-filled from the patient sex and the age at admission/examination. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

The problem here is that each PR should include in demo data not only its own changes, but also the ones introduced by already merged PRs... I guess this will emerge as "conflict" after other merges, so we'll see... Please keep monitoring this, thanks!

giuseppesorge:
Merged `develop` (955717f9). Besides `step_04`, `sql/load_demo_data.sql` conflicted as well: the dump is generated, and both sides had touched the same rows. Here it was the `oh_admission` rows: `ADM_SEX` and `ADM_AGE` land in 7th and 8th position, develop's two `ADM_DELETED_*` columns at the end, 38 values per row.

dbmalkovsky (commented):
This looks good to me with the following changes:

1. Patient Examination files have an updated copyright year but the Admission related files do not.   There is a maven task that will update all changed files to the most recent change year which is usually run at release time.   So either change all or change none.
2. Rename `step_aXXX_  ...` to `step_a124_ ....` so the PR can be merged.

mwithi:
> This looks good to me with the following changes:
> 
> 1. Patient Examination files have an updated copyright year but the Admission related files do not.   There is a maven task that will update all changed files to the most recent change year which is usually run at release time.   So either change all or change none.
> 2. Rename `step_aXXX_  ...` to `step_a124_ ....` so the PR can be merged.

I agree with suggestions if also other PR linked to this issue are ready to be merged, is it so @dbmalkovsky ?
@giuseppesorge please be ready

giuseppesorge:
Thanks @dbmalkovsky, @mwithi — both points are done.

1. **Migration name.** `sql/step_aXXX_sex_age_snapshot.sql` → `sql/step_a124_sex_age_snapshot.sql` (`a123` is the last one on `develop`), with the `source` line in `step_04_all_following_steps.sql` updated to match. (`ad564aa2`)
2. **Copyright year.** The three Admission classes and `admission/Tests.java` were the ones left behind; all the files changed by this PR now read `2006-2026`. (`6b5650ff`)

Verified:

- Built a demo database from `create_all_demo.sql` on MariaDB 10.6: the renamed step is picked up through the `step_04` chain and creates `ADM_SEX`/`ADM_AGE` and `PEX_SEX`/`PEX_AGE`.
- The backfill matches the age computed from `PAT_BDATE` at the admission date on **460 of 460** demo admissions, and the sex snapshot never diverges from the patient's.
- `mvn test` on `org.isf.admission` + `org.isf.examination`: 195 tests, all green.

@mwithi — GUI #2237 is now rebased on `develop` (it was failing to compile against a stale base, unrelated to this change) and API #609 carries the same copyright change, so the three should be ready to go together.

giuseppesorge:
@dbmalkovsky @mwithi — a follow-up on the copyright point from this thread, since it applies well beyond this PR.

I checked all my open PRs against the "either change all or change none" rule and **26 of them were internally inconsistent** — a mix of 2023/2024/2025/2026 headers across the files each one touches, which would have earned the same comment one PR at a time. I've aligned those to `2006-2026`, touching only files the PR already changes and nothing else:

- **core** — #1604, #1617, #1620, #1621, #1622, #1630, #1632, #1633, #1644, #1645, #1646, #1647, #1648
- **api** — #589, #590, #592, #593, #594, #595, #601, #603, #604, #607
- **gui** — #2218, #2220, #2236

Each is a single commit, *"Align the copyright year across all the changed files"*.

**What I deliberately left alone, and the question:** a further group of PRs is already *uniform* on an older year (2023, 2024 or 2025 throughout). Those satisfy "change none" as it stands, so I didn't touch them — but they also don't carry the current year on files we modified, which is what @mwithi asked for a while back. Do you want those bumped to 2026 as well, or is it better to leave them to the Maven task at release time?

Happy to revert any of the 26 if this reads as overreach on PRs that were already reviewed — just say which and I'll drop the commit.

giuseppesorge:
One more thing I found while re-testing this on a real database, pushed as `582eb3dd`.

The backfill and the runtime disagreed for patients recorded without a birth date. `Admission.applyPatientData()` falls back to the recorded `PAT_AGE` in that case, but `step_a124` wrote `NULL`:

```sql
-- before
ELSE NULL
-- after
ELSE p.PAT_AGE
```

It matters because the aggregate reports filter on `ADM_AGE < 5` and `ADM_AGE >= 5`, and in MariaDB `NULL` satisfies neither — so the admission was not landing in the wrong band, it was dropping out of the statistics altogether. The same patient would therefore be counted when admitted after the upgrade and invisible when backfilled.

Verified on a demo database built from `create_all_demo.sql`, with a patient carrying `PAT_BDATE = NULL` and `PAT_AGE = 7`:

| backfill | `ADM_AGE` | `< 5` | `>= 5` |
|---|---|---|---|
| before | `NULL` | `NULL` | `NULL` |
| after | `7` | false | **true** |

The 460 demo admissions are unaffected — they all have a birth date, so `sql/load_demo_data.sql` does not change and the existing numbers are identical.

I also added `testIoNewAdmissionUsesRecordedAgeWhenBirthDateIsMissing` to `admission/Tests.java`: that branch of `applyPatientData()` had no coverage. To be clear about what it does and does not prove — it pins the runtime behaviour the migration now mirrors; the migration itself is not exercised by the Java suite, so the table above is the evidence for that half. `org.isf.admission.Tests` is green at 168.

This is arguably beyond what you asked for on this PR, so if you would rather keep the change minimal I am happy to drop the commit and open a separate ticket. My reason for proposing it now is that `step_a124` has not been merged yet, so correcting it costs two lines — afterwards it would need a whole new migration step to repair the rows already backfilled to `NULL`.

mwithi:
> This is arguably beyond what you asked for on this PR, so if you would rather keep the change minimal I am happy to drop the commit and open a separate ticket. My reason for proposing it now is that `step_a124` has not been merged yet, so correcting it costs two lines — afterwards it would need a whole new migration step to repair the rows already backfilled to `NULL`.

Every patient should not have bdate `null` because whenever an age or a agetype is specified, it is automatically converted into bdate before the persistence. The only patient with bdate `null` is patient 0 that is there for backward compatibility with OH simplified usage (without patient registration process unless for admissions)

mwithi:
> > This is arguably beyond what you asked for on this PR, so if you would rather keep the change minimal I am happy to drop the commit and open a separate ticket. My reason for proposing it now is that `step_a124` has not been merged yet, so correcting it costs two lines — afterwards it would need a whole new migration step to repair the rows already backfilled to `NULL`.
> 
> Every patient should not have bdate `null` because whenever an age or a agetype is specified, it is automatically converted into bdate before the persistence. The only patient with bdate `null` is patient 0 that is there for backward compatibility with OH simplified usage (without patient registration process unless for admissions)

anyway, I guess the guard is a good thing in any case.

mwithi (approved):
LGTM. Thanks a lot!

PR #1603, opened 2026-06-18
OP-1005 Add DELETED_BY and DELETED_DATE audit fields for soft deletions

giuseppesorge:
## What & why

OP-1005 — audit **soft deletions** the same way creations and modifications are audited. Adds `DELETED_BY` / `DELETED_DATE` so we record *who* soft-deleted a row and *when*.

## Changes

- New `org.isf.utils.db.SoftDeletableAuditable<U>` (`@MappedSuperclass` extending `Auditable`) carrying `DELETED_BY` / `DELETED_DATE` + a `recordDeletion(deletedBy)` helper that stamps both (timestamp = now in one place). These are **not** Spring-managed (there is no `@DeletedBy`), so they are set explicitly on the soft-delete path.
- The 5 entities that actually soft-delete now extend it and add the two `@AttributeOverride`s for their `<PREFIX>_DELETED_BY/DATE`: **Patient, Admission, Supplier, User, UserGroup**.
- Populate on every soft-delete path, from the current auditor (`AuditorAwareInterface` — the same source Spring uses for `createdBy`/`lastModifiedBy`) + timestamp:
  - entity-save paths call `recordDeletion(...)` before `save` (`AdmissionIoOperations.setDeleted`, `SupplierOperations.delete`);
  - `@Modifying` JPQL paths add the two columns to the UPDATE and the caller passes them (`Patient` via `updateDeleted`; `User`/`UserGroup` via `MenuIoOperations`, conditional on the `deleted` flag so a regular update / un-delete clears the deletion audit).
- Migration `sql/step_a123_add_deleted_by_and_date.sql`: adds the columns (mirroring the existing `*_LAST_MODIFIED_BY/DATE` types) **with a foreign key to `OH_USER`**, like every other `*_BY` audit column (see `step_a115`). Verified on a real MariaDB (columns, FK creation and FK enforcement).

## Scope note

**Medical** and **MedicalType** carry a `*_DELETED` flag but their delete is currently **physical** (`repository.delete`), so they are intentionally **left on `Auditable`** — adding deletion-audit columns to entities that never soft-delete would be dead schema. They can be migrated when a real soft delete is implemented for them.

## Testing

- Tests for all 5 instrumented entities: `DELETED_DATE` is stamped on a soft delete and stays null on a regular update; parameterized where the sibling tests are. Teeth-verified (removing the populate fails the test).
- Full `org.isf` suite green (2454 tests, 0 failures). Migration applied and FK-checked on MariaDB 10.6.

Jira: OP-1005


src/main/java/org/isf/patient/model/Patient.java
mwithi, line 146: should we at the same time get rid of previous fields with the same goal?
Also, in terms of migrating a previous version, should we reflect already soft-deleted data into the new fileds?

giuseppesorge: On dropping the previous fields — I'd keep `DELETED_BY`/`DELETED_DATE` separate. `LAST_MODIFIED_BY`/`LAST_MODIFIED_DATE` do capture the actor/time of the soft-delete UPDATE, but they get overwritten by any later edit or un-delete, so they aren't a stable record of the deletion event; the dedicated columns give a durable audit of the deletion itself.

On back-filling already-deleted rows — my recommendation is to leave them NULL and only populate the new columns for deletions performed from now on. `DELETED_BY` is an audit fact, and back-filling it from `LAST_MODIFIED_BY` would assert that a specific user did the deletion when we can't actually be sure of that (LAST_MODIFIED could reflect a later change). NULL cleanly means "deleted before auditing existed / unknown".

That said, since in practice a soft-deleted row isn't edited afterwards, a best-effort backfill (`DELETED_DATE = LAST_MODIFIED_DATE`, `DELETED_BY = LAST_MODIFIED_BY` where the deleted flag is set) would be accurate for the vast majority of rows — happy to add that to `step_a123` if you'd rather have the continuity. Let me know which you prefer.

mwithi: @giuseppesorge I was referring to the pre-exising PAT_DELETED field for soft-deletion, not the other audit fields

giuseppesorge: Good catch — you're right that my previous reply was about the `LAST_MODIFIED_*` columns. On `PAT_DELETED` itself: I'd keep it.

`PAT_DELETED` is the *state* flag — it's what every soft-delete query filters on (`deleted = 'N' or is null`), what the soft-delete sets (`... set deleted = 'Y'`), and what the active-patient count uses. The two new columns are, by design, who/when only: `SoftDeletableAuditable` deliberately carries no `deleted` flag, so they augment `PAT_DELETED` rather than replace it — the same split `Auditable` already makes between a record and its `*_BY`/`*_DATE` audit columns. The soft-delete path now sets all three atomically (`deleted = 'Y'`, `deletedBy`, `deletedDate`), so the flag and the audit stay consistent.

We *could* drop the flag and treat `PAT_DELETED_DATE IS NOT NULL` as the deleted marker, but that means rewriting the filter predicates (and losing the cheap indexed boolean) on all five soft-deletable entities, each with its own `XXX_DELETED` column — plus a breaking change to a long-standing column that reports/other modules read. I'd keep it out of this PR; happy to open a follow-up ticket if you'd prefer to move to a single `DELETED_DATE`-based indicator project-wide.

On migrating existing data: `step_a123` only adds the columns (NULL default), so rows already soft-deleted keep `DELETED_BY`/`DELETED_DATE` NULL — i.e. "deleted before auditing existed". We can't reconstruct who/when reliably (best-effort from `LAST_MODIFIED_DATE` would be a guess), so I left them NULL rather than inventing values.

mwithi: You're right the new auditing fields may be seen as augmentation of the previous _DELETED state flag. So let's keep them, thanks. I will test it and check other changes to see the final merge order.

mwithi: Update step_a123 to a118 (going to merge first!)
Thanks!

dbmalkovsky (approved):
LGTM

PR #1604, opened 2026-06-19
OP-1069 Move report launcher discovery logic to CORE

giuseppesorge:
## OP-1069 — Move report launcher logic from GUI to CORE

Part of [OP-1069](https://openhospital.atlassian.net/browse/OP-1069). This is the **CORE** piece (companion GUI + API PRs linked below).

### What
Adds `JasperReportsManager.getReportsList()`, which scans the `rpt_stat` and `rpt_extra` folders for compiled `.jasper` reports and, for each, resolves:
- the localized title (`<folder>/<lang>/<name>.properties` → `<folder>/<name>.properties` fallback, same convention as `addBundleParameter`),
- the parameters the report prompts the user for (e.g. `month`/`year` or `fromdate`/`todate`).

Returns a new `ReportLauncherDto` list sorted by title. This is the logic that previously lived in the GUI `ReportLauncher`; moving it to CORE lets it be reused (the new REST endpoint, see API PR).

Scanning is **best-effort**: a missing/unreadable folder is logged and skipped rather than failing the whole listing (preserves the launcher's old non-fatal behavior).

### Notes
- The `rpt_*` folders are resolved relative to the process working directory, matching the existing `rpt_base` convention in the same class. Per the ticket the folders stay in the GUI for now (relocation is a follow-up task).

### Tests
- New unit test in `org.isf.stat.Tests` covering folder scan, localized + default title resolution, prompt-parameter extraction (system params excluded), sorting, and skipping reports without a title.
- `org.isf.stat.Tests` 5/5 green.

### Verification
Booted against the real `rpt_stat`/`rpt_extra` shipped with the GUI → 32 reports discovered with correct folder/title/file name/parameters.

Companion PRs (same branch name so GUI/API CI cross-build this core branch):
- GUI: _linked in a comment_
- API: _linked in a comment_

[OP-1069]: https://openhospital.atlassian.net/browse/OP-1069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Companion PRs for OP-1069:
- GUI: informatici/openhospital-gui#2215
- API: informatici/openhospital-api#588

giuseppesorge:
Merged `develop` and pushed 0f549bf2 on top of it.

While resolving the conflict on the GUI side (informatici/openhospital-gui#2215, whose only conflict was against OP-214 #2203) it turned out that the discovery logic moved here had lost part of that fix. `getReportsList()` resolved the language through `new Locale(GeneralData.LANGUAGE).getLanguage()`, which lowercases the languages that carry a country: `am_ET` becomes `am_et`, so on a case-sensitive filesystem the localized folder is never found and every report silently falls back to the English title. The release ships `am_ET`, `sv_SE` and `zh_CN` under `rpt_stat` and `rpt_base`. OP-214 had already fixed this in the GUI by using the configured value as it is, and moving the logic must not lose it.

The commit uses `GeneralData.LANGUAGE` unchanged, restores the `Files.exists` guard from #2203 — so a missing localized bundle, which is the normal case for English, no longer logs an error per report — and adds a test for a country-specific language. Worth saying explicitly: that test cannot fail on a case-insensitive filesystem such as macOS, it is CI on Linux that exercises the difference.

PR #1612, opened 2026-06-19
OP-1065 Consolidate and deprecate getLaboratoryForPrint (GUI-only)

giuseppesorge:
## OP-1065 — Optimize getLaboratoryForPrint methods (only for GUI)

CORE part (companion GUI PR linked below). Sub-task of OP-1062.

These methods back only the GUI report print, and there were several overloads doing the same thing. This reduces them to one method per layer and deprecates them (they are GUI-only).

### Changes
- **`LabIoOperations`**: merged the `()`, `(exam,from,to)` and `(exam,from,to,patient)` overloads into a single `(exam,from,to,patient)`. **Behaviour preserved**: with a patient → exact exam-description match + patient data in the result; without a patient → partial (`Containing`) match + order by exam type (the existing production GUI behaviour). The two query semantics were intentionally kept (conservative consolidation).
- **`LabManager`**: merged the two overloads into one `(exam,from,to,patient)` that **always** applies `setLabMultipleResults` — the previous patient overload skipped it (an inconsistency, now fixed).
- Both consolidated methods are `@Deprecated` (meant for GUI use only).

### Tests
- `org.isf.lab.Tests` updated to the consolidated methods + a new case covering the patient path with multiple results. Lab suite **196** green; full core suite **2448** green.

### Verification
Real MariaDB run of the exact GUI call (`getLaboratoryForPrint(type, from, to, null)`) returns the expected rows (partial match), and a real Swing run of the Laboratory Browser print opens the report viewer correctly.

Companion PR (same branch name → GUI CI cross-builds this core branch):
- GUI: _linked in a comment_


giuseppesorge:
Companion GUI PR for OP-1065: informatici/openhospital-gui#2216

mwithi (approved):
LGTM. Thanks a lot!

PR #1613, opened 2026-06-19
OP-1413 Pin commons-collections4 to 4.5.0

giuseppesorge:
## OP-1413 — Bean-collection report printing broken (dependency convergence)

CORE part (companion GUI PR linked below).

### Problem
Printing any report built from a `JRBeanCollectionDataSource` (e.g. Laboratory → Print table) fails at runtime:
```
java.lang.NoClassDefFoundError: org/apache/commons/collections4/map/ConcurrentReferenceHashMap
  at org.apache.commons.beanutils2.BeanUtils.createCache(...)
  at net.sf.jasperreports.engine.data.JRBeanCollectionDataSource.getFieldValue(...)
  at net.sf.jasperreports.engine.fill.JasperFillManager.fillReport(...)
  at org.isf.serviceprinting.manager.PrintManager.print(...)
```

### Root cause (`mvn dependency:tree -Dverbose`)
- `org.apache.poi:poi-ooxml:5.3.0 → poi:5.3.0 → commons-collections4:4.4`
- `net.sf.jasperreports:7.0.6` (and its `commons-beanutils2:2.0.0-M2`) require `commons-collections4:4.5.0`
- Maven mediation keeps **4.4** (`4.5.0 omitted for conflict with 4.4`)
- `commons-beanutils2 2.0.0-M2` uses `ConcurrentReferenceHashMap`, which only exists from collections4 **4.5.0** → missing at runtime.

### Fix
Manage `commons-collections4` to `4.5.0` in `dependencyManagement` (4.5.0 is backward compatible with POI 5.3.0).

### Verification
- `mvn dependency:tree` now resolves collections4 **4.5.0**; full core suite **2448** green.
- Real Swing run of the Laboratory Browser **Print table** now opens the report viewer correctly (previously threw the NoClassDefFoundError).

Companion PR (the GUI module needs the same pin since dependencyManagement is not inherited from the core artifact):
- GUI: _linked in a comment_

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
 		<jasperreports.version>7.0.6</jasperreports.version>
 		<license.maven.plugin.version>5.0.0</license.maven.plugin.version>
 		<spring-cloud.version>2025.0.1</spring-cloud.version>
+		<commons-collections4.version>4.5.0</commons-collections4.version>
 	</properties>
 	
 	<dependencyManagement>
@@ -38,6 +39,15 @@
 	            <version>${spring-cloud.version}</version>
 	            <type>pom</type>
 	            <scope>import</scope>
+	        </dependency>
+	        <!--
+	            Force commons-collections4 4.5.0: jasperreports 7.0.6 (via commons-beanutils2) needs 4.5.0, but poi-ooxml pulls 4.4 and wins
+	            version mediation. With 4.4 any JRBeanCollectionDataSource print fails at runtime (missing ConcurrentReferenceHashMap). See OP-1413.
+	        -->
+	        <dependency>
+	            <groupId>org.apache.commons</groupId>
+	            <artifactId>commons-collections4</artifactId>
+	            <version>${commons-collections4.version}</version>
 	        </dependency>
 	    </dependencies>
 	</dependencyManagement>


giuseppesorge:
Companion GUI PR for OP-1413: informatici/openhospital-gui#2217

dbmalkovsky:
Nice investigation. Yes your method will work but there is IMHO an easier solution. Move the `org.apache.poi` dependency (lines 109-113) in the `pom.xml` file after `jasperreports` and before the `org.apche.commons` dependency That is, before line the old line 178 and the after the new line 172. This changes the `commons-collections4` dependency to be resolved from `jasperreports` and not `org.apache.poi`. The nice side effect is that there is no need to change the GUI pom.xml file. Does this make sense?

dbmalkovsky:
Another `solution` is to update `pop-ooxml` to a more current version; see [https://mvnrepository.com/artifact/org.apache.poi/poi-ooxml/5.5.1](https://mvnrepository.com/artifact/org.apache.poi/poi-ooxml/5.5.1).   The nice side effect is that it fixes some direct and indirect vulnerabilities.

I just noticed there is a PR to do exactly that, move to 5.5.1, see https://github.com/informatici/openhospital-core/pull/1542

giuseppesorge:
Closing in favour of #1542, which upgrades poi-ooxml to 5.5.1 — that resolves the commons-collections4 version conflict at its source (and also clears some POI vulnerabilities), as you suggested in review. Thanks for the pointer.

PR #1615, opened 2026-06-19
OP-1358 Add virtual flag to Ward and Supplier (Phase 1)

giuseppesorge:
## OP-1358 — Define virtual supplier and virtual wards (Phase 1) — CORE

Introduces a `virtual` boolean on Ward and Supplier so virtual wards (LOST, BROKEN, EXPIRED, INVENTORY) and virtual suppliers (INVENTORY, RECTIFICATION) can be used as targets/origins of Main Store discharge/charge movements.

### Changes
- `Ward.isVirtual` (`WRD_IS_VIRTUAL`) and `Supplier.isVirtual` (`SUP_IS_VIRTUAL`), mirroring the existing `WRD_IS_PHARMACY` flag (getter/setter only; no constructor change, to avoid test ripple).
- Migration `step_a121_add_virtual_flag_to_ward_and_supplier.sql` (`TINYINT(1) NOT NULL DEFAULT 0`), registered in `step_04_all_following_steps.sql`.
- Test fixtures round-trip the new flag.

### Scope / decisions
- **Phase 1 only.** The broader unification of wards/suppliers into a single `location` model (multi-level warehouse) described in the ticket is explicitly out of scope.
- No demo rows seeded (virtual wards/suppliers are user-created); decoration not filtering (combos show all + "(virtual)"); `equals`/`hashCode` unchanged (display flag).
- **Migration numbered `a121`** to avoid collision with other open migration PRs (a118–a120); may need renumbering at merge.

### Verified
core ward+supplier tests 66 (round-trip); real MariaDB migration adds both columns.

Companion PRs (same branch → gui/api CI cross-build this core branch): gui/api/doc linked in a comment.


giuseppesorge:
Companion PRs for OP-1358:
- gui: informatici/openhospital-gui#2218
- api: informatici/openhospital-api#590
- doc: informatici/openhospital-doc#285

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`d9a423dc`): added WRD_IS_VIRTUAL and SUP_IS_VIRTUAL (default 0) to oh_ward and oh_supplier. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`d9a423dc`): added WRD_IS_VIRTUAL and SUP_IS_VIRTUAL (default 0) to oh_ward and oh_supplier. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

giuseppesorge:
Merged `develop` (9d1271ec). Besides `step_04`, `sql/load_demo_data.sql` conflicted as well: the dump is generated, and both sides had touched the same rows. Here it was the four `oh_supplier` rows: this branch inserts `SUP_IS_VIRTUAL` right after `SUP_DELETED`, while develop appends `SUP_DELETED_BY` and `SUP_DELETED_DATE` at the end. The merged rows carry all of them, 18 values matching the 18 columns of the merged `CREATE TABLE`.

One thing for whoever merges first: this branch claims `step_a125`, and so does #1634 (OP-1021) — whichever lands second will have to be renumbered.

giuseppesorge:
Merged `develop` in to clear the conflict the merge back left (4f93fae7).

`sql/load_demo_data.sql` merged on its own here, which is the case worth double checking rather than trusting, so I rebuilt the expected result independently - develop's regenerated dump with the two virtual-flag columns added - and it came out identical to what the merge produced. None of the demo data the release brought in was lost.

The only conflict was the step list: develop's wins and `step_a125_add_virtual_flag_to_ward_and_supplier.sql` goes back at the end, after a124. Note that #1634 also claims `a125`, so one of the two will need renumbering at merge time.

PR #1616, opened 2026-06-19
OP-1242 Make patient profession an editable lookup table

giuseppesorge:
See OP-1242.

## What changed

Patient professions are no longer hard-coded in `PatientBrowserManager`: they are now a managed lookup table (`OH_PROFESSION`), mirroring the existing DiseaseType pattern.

- New `Profession` entity/repository/IO-operation and `ProfessionBrowserManager` (CRUD + validation).
- Refactor `PatientBrowserManager.getProfessionList`/`getProfessionTranslated`/`getProfessionKey` to read from the table, with their signatures unchanged so GUI callers are unaffected. Remove the hard-coded map.
- Migration `step_a123_create_profession_table.sql`: create `OH_PROFESSION`, seed the 12 legacy professions (the code keeps the old key, so existing `PAT_PROFESSION` values keep resolving) and register the Profession menu item (`INSERT IGNORE` for fresh-install idempotency).
- Register the menu row in the fresh-install dumps (`step_02`, `step_05`, `load_demo_data`).
- Add Profession unit tests and update patient tests for the DB-backed manager.

## Tests

`profession.Tests` 19/19, `patient.Tests` 71 (0 failures).

## Notes

- **i18n trade-off**: the 12 built-in professions become DB free-text English descriptions, losing per-language translation, consistent with all other OH lookup tables.
- SQL step number `a123` was chosen to avoid colliding with a118 to a122 (reserved by other open PRs); it may need renumbering at merge (a comment in the file notes this).
- Related PRs in openhospital-api, openhospital-gui and openhospital-doc depend on this one being merged first.


src/main/java/org/isf/profession/manager/ProfessionBrowserManager.java
```diff
@@ -0,0 +1,141 @@
...
+	 *
+	 * @param profession the profession to remove.
+	 * @throws OHServiceException
+	 */
+	public void deleteProfession(Profession profession) throws OHServiceException {
+		ioOperations.deleteProfession(profession);
+	}
+
+	/**
+	 * Verify if the object is valid for CRUD and return a list of errors, if any
```
dbmalkovsky:
```suggestion
	 * Verify if the object is valid for CRUD and throw an exception with the list of errors, if any.
```
Again, a `void` method so this is more precise.

giuseppesorge: Done in d9a9ee93 -- updated the Javadoc to your suggested wording. Thanks.

giuseppesorge:
Merged `develop` (c40276b6). Besides `step_04`, `sql/load_demo_data.sql` conflicted as well, on a single `oh_groupmenu` row: id 349 for `profession` on this branch, id 350 for `lotmanagement` on develop. Both rows are kept and the ids do not clash. Checked afterwards that no id is duplicated in the file, that every `INSERT` carries as many values as its `CREATE TABLE` declares, and that this branch's delta against `develop` is the same as before the merge (the menu row and the `oh_menuitem` row).

For the record, since it came up while reviewing the merge: the new `oh_profession` table is deliberately absent from the dump. `step_a124` creates it and seeds the twelve professions, `delete_all_data.sql` does not touch it and `load_demo_data.sql` does not drop it, so a demo database gets the seeded table. Only the two menu rows need to be in the dump, because `load_demo_data.sql` does recreate `oh_menuitem` and `oh_groupmenu`.

One thing for whoever merges first: this branch claims `step_a124`, and so does #1620 (OP-1037) — whichever lands second will have to be renumbered.

sql/step_05_update_menu_i18n.sql
```diff
@@ -18,6 +18,7 @@ UPDATE MENUITEM SET MNI_BTN_LABEL='angal.menu.btn.disease', MNI_LABEL='angal.men
 UPDATE MENUITEM SET MNI_BTN_LABEL='angal.menu.btn.exams', MNI_LABEL='angal.menu.exams' WHERE MNI_ID_A='exams';
 UPDATE MENUITEM SET MNI_BTN_LABEL='angal.menu.btn.operation', MNI_LABEL='angal.menu.operation' WHERE MNI_ID_A='operation';
 UPDATE MENUITEM SET MNI_BTN_LABEL='angal.menu.btn.vaccine', MNI_LABEL='angal.menu.vaccine' WHERE MNI_ID_A='vaccine';
+UPDATE MENUITEM SET MNI_BTN_LABEL='angal.menu.btn.profession', MNI_LABEL='angal.menu.profession' WHERE MNI_ID_A='profession';
```
mwithi: This line is not needed because it is added in the new step

```suggestion
```

sql/step_aXXX_create_profession_table.sql
```diff
@@ -0,0 +1,36 @@
...
+  ('farming', 'Farming', 'admin', NOW(), 'admin', NOW()),
+  ('construction', 'Construction', 'admin', NOW(), 'admin', NOW()),
+  ('medicine', 'Medicine', 'admin', NOW(), 'admin', NOW()),
+  ('foodhospitality', 'Food/Hospitality', 'admin', NOW(), 'admin', NOW()),
+  ('homemaker', 'Homemaker', 'admin', NOW(), 'admin', NOW()),
+  ('mechanic', 'Mechanic', 'admin', NOW(), 'admin', NOW()),
+  ('business', 'Business', 'admin', NOW(), 'admin', NOW()),
+  ('janitorial', 'Janitorial Services', 'admin', NOW(), 'admin', NOW()),
+  ('mining', 'Mining', 'admin', NOW(), 'admin', NOW()),
+  ('engineering', 'Engineering', 'admin', NOW(), 'admin', NOW());
```
mwithi, line 30: first insert is ok for English, but what about other languages? we need to follow CSV schema I guess

mwithi (changes requested):
Please review my comments

sql/step_05_update_menu_i18n.sql
giuseppesorge: Done in 24689629 — and you were right that the row belongs to the new step only: I also removed the menu additions from `step_02_dump_menu.sql`, because with the step_05 line gone a fresh installation would have kept the literal `Profession` labels from step_02 (the step's `INSERT IGNORE` found the row already there and did nothing). Now the migration step is the single source of the menu item for fresh and migrated databases alike, with the i18n keys directly.

sql/step_aXXX_create_profession_table.sql
giuseppesorge: Good point, done in 24689629 following the vaccine/dicomtype pattern: each `create_all_<lang>.sql` now sources a `step_03_dump_profession_data_<lang>.sql` after step_04, which deletes the English defaults seeded by the migration step and loads `data_<lang>/profession.csv`. The descriptions are taken verbatim from the existing GUI bundles (`angal.patient.profession.*.txt`), so each language shows exactly what it showed before this change. The Portuguese bundle has no profession entries, so its CSV keeps the English texts the application falls back to today.

Migrated databases keep the English defaults from the step — the migration cannot know the installation language — and they can be edited in the new Profession browser.

Verified on MariaDB 10.6: fresh installs from `create_all_it/fr/ar` get the localized rows (accents and Arabic intact), running only the migration step on a database without the table gets the English seeds plus menu item and grant, and `create_all_demo` still builds with the profession table populated.

giuseppesorge:
While testing the localized seeds in the running GUI (fresh Italian database, patched core in the client package) I noticed a behavioural regression this PR had introduced, now fixed in b3cfa4ed: the old hard-coded list started with `unknown`, so the profession combo box preselected it for new patients; reading the table in description order silently made the alphabetically first profession the default — a new patient saved without touching the combo would have ended up as *Business*/*Affari*. `getProfessionList()` now returns `unknown` first and the rest sorted by localized description, and `testMgrGetProfessionList` asserts the first element so it cannot regress again.

Verified live: the combo preselects *Sconosciuto* on the Italian database and lists *Affari, Agricoltura, Altro, …* alphabetically below it.

PR #1618, opened 2026-06-23
OP-1414 Fix StaleObjectStateException when saving a new main-store inventory

giuseppesorge:
## OP-1414 — Fix StaleObjectStateException when saving a new main-store inventory

JIRA: https://openhospital.atlassian.net/browse/OP-1414

### Problem
Creating a new main-store inventory (Pharmacy → Inventory → New → add medicals → Save) fails with `ObjectOptimisticLockingFailureException` / `StaleObjectStateException: ...[MedicalInventoryRow#0]` ("unsaved-value mapping was incorrect"), and the inventory is not saved.

### Root cause
`MedicalInventoryRow` declares `@Id @GeneratedValue(IDENTITY) Integer id`, but the GUI builds every new row with `new MedicalInventoryRow(0, ...)` (id = 0). Because the id is a wrapper `Integer`, Spring Data's new-entity detection treats a non-null `0` as an *existing* id and issues a `merge`/update instead of an `insert`; with the `@Version` field present, Hibernate then throws because there is no row with id 0.

### Fix
In `MedicalInventoryRowIoOperation.newMedicalInventoryRow`, null a new row's id (`id == 0`) before `save`, so Spring Data treats it as transient and inserts it.

Alternative/root fix worth considering (left to the maintainer): change `MedicalInventoryRow.id` from `Integer` to a primitive `int`, consistent with the other OH entities where `0` already means "new" — this avoids the wrapper-zero ambiguity at the source. Kept this PR to the minimal, low-ripple change.

### Verification
Verified end-to-end on the real Swing GUI against a demo database: before the fix, saving a new main-store inventory threw the exception; after the fix, the inventory saves (428 rows persisted, no exception) and can be reopened, validated and actualized normally.

Note: this bug was found while testing OP-1350 (the inventory editor highlight). The `org.isf.medicalinventory` module currently has no unit-test harness in core, so this persistence fix is covered by the end-to-end run above; happy to add an integration test if you'd like the module's first one.


giuseppesorge:
While testing OP-1216 end-to-end: this fix also resolves the same `StaleObjectStateException` on the **ward** inventory save (`InventoryWardEdit`) — both inventory screens build new rows with id 0 and go through `newMedicalInventoryRow`. Verified on a MariaDB 10.6 demo DB: with this fix applied both the main-store and the ward inventory save correctly; without it the failure is silent in the GUI (the runtime exception escapes the Swing listener, so no error dialog appears and the window won't close on Save). Merging this first also unblocks end-to-end testing of OP-1216 (#1597 / gui#2212).

mwithi:
Closing this PR as superseded by the fix implemented in GUI https://github.com/informatici/openhospital-gui/commit/85c6f856d6290498c48ca889801c904a1a725f9c

PR #1620, opened 2026-06-23
OP-1037 Record the administrative issues of a patient in a table of their own

giuseppesorge:
## OP-1037 — Administrative issues (core)

Records the administrative issues the hospital raises on a patient — a missing document, an identity still to be verified, an open dispute — in a table of their own, as @mwithi asked on the ticket: an issue is a hospital decision, not the patient's will, so it does not belong to `PATIENT_CONSENSUS`, and a single overwritten column would lose who flagged the patient, when, and when the issue was cleared.

### Changes
- New table `OH_PATIENT_ADMIN_ISSUE` (`sql/step_aXXX_patient_admin_issue.sql`, registered in `step_04_all_following_steps.sql`): `PAI_ID`, `PAI_PAT_ID` (FK to `OH_PATIENT`, `ON DELETE NO ACTION`), `PAI_REASON NOT NULL`, `PAI_FROM NOT NULL`, `PAI_TO NULL` (null = still open) and the standard audit block. A patient is flagged while at least one row has `PAI_TO IS NULL`; several issues can be open at once, and resolving one only sets its `PAI_TO`.
- New package `org.isf.patadminissue`: the `PatientAdminIssue` entity, its repository and io operations, and `PatientAdminIssueBrowserManager` with `getOpenIssues` (one patient, or a batch of patient codes in a single query), `getIssues` (history), `openIssue` / `resolveIssue`, and `saveIssues` that opens and resolves a set of issues in a single transaction, validating everything before writing anything.
- `AdminIssuePatientMergedEventListener`: the issues follow the surviving patient when two patients are merged, like every other patient-scoped record.
- `PatientConsensus` and its manager are back to `develop`: nothing of this feature touches the consensus any more.
- `sql/load_demo_data.sql`: the `oh_patient_consensus` block is identical to `develop` again; the new `oh_patient_admin_issue` block carries one open issue on patient 16 and, on patient 42, one resolved and one open issue, so both the warning and the history are visible on a demo install.
- Tests: open, open twice, resolve keeps the row and leaves the other issues open, resolve twice or an issue never opened, blank / too long / exactly 255 chars reason, missing patient, ordering of open issues and history, batch query across patients and with no patient, all-or-nothing save, patient merge, `toString()`.

### Notes for reviewers
- **Step number**: still the `aXXX` placeholder, to be assigned by merge order.
- **Upgrade path**: as for the other open post-1.15 migrations, the step is in the fresh-install aggregator and will need rolling into the next `update_1_15-1_16.sql`.

### Testing
- Full core suite green on JDK 17.
- MariaDB 10.6: the migration applies on a database loaded from the `develop` demo dump; the regenerated dump loads clean on a fresh database and, against `develop`, differs only by the new table (checked table by table).

Part of OP-1037: API informatici/openhospital-api#592, GUI informatici/openhospital-gui#2222, doc informatici/openhospital-doc#296, web UI informatici/openhospital-ui#817.


giuseppesorge:
Merged `develop` (7816c42c) to clear the conflict in `sql/step_04_all_following_steps.sql`: OP-1231 appended `step_a123_add_lot_management_in_pharmacy.sql` at the end of the file, so this branch's own `source` line now comes after it. Nothing else changed, and the branch's diff against `develop` is the same as before the merge.

One thing for whoever merges first: this branch claims `step_a124`, and so does #1616 (OP-1242) — whichever lands second will have to be renumbered.

giuseppesorge:
Updated after @mwithi's design decisions on [OP-1037](https://openhospital.atlassian.net/browse/OP-1037): the flag now carries an optional free-text reason.

- `PTC_ADMINISTRATIVE_REASON VARCHAR(255)` added to the same migration step as the flag, and to `load_demo_data.sql` (all 525 rows), so a clean demo install and a migrated database end up with the same schema.
- The manager rejects a reason longer than the column and a reason given without the flag, so the two fields cannot drift apart. The length lives in one place, `PatientConsensus.ADMINISTRATIVE_REASON_LENGTH`, and the GUI field is capped with it.
- The 3-argument `PatientConsensus` constructor is left as it was: the new fields are set through their setters, so `TestPatient` is out of the diff again.
- Two demo patients (5 and 16) now carry the flag with a plausible reason, so the feature is visible on a demo install.

Verified on MariaDB 10.6: fresh `create_all_demo` loads clean, and the migration applied to a database that already had 525 consensus rows leaves them at `0`/`NULL`. Full core suite green (2478 tests).

The step is still `step_aXXX`: it needs to become `a125` or later at merge time, since `a124` is taken.

[OP-1037]: https://openhospital.atlassian.net/browse/OP-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

giuseppesorge:
Small follow-up on the demo data: the flag is now on patients 16 and 42 instead of 5.

Patient 5 is the first non-deleted patient, so anything that picks "the first patient" — demo walkthroughs, and our own UI tests — would land on a flagged patient and meet the warning every time. Two flagged patients further down the list still demonstrate the feature without getting in the way.

giuseppesorge:
Merged `develop` in to clear the conflict the merge back left (c3708a42).

The regenerated dump took `oh_patient_consensus` from 525 to 634 rows, so the patch this branch had applied no longer fitted and the conflict left the table with 9 values against 11 columns. Rather than hand-merging, I rebuilt the file: develop's dump with `PTC_ADMINISTRATIVE` and `PTC_ADMINISTRATIVE_REASON` added after `PTC_CONSENSUS` across all 634 rows, and the two demo rows that carry a reason - PTC_ID 16 and 42 - restored to exactly the values this branch had.

Checked that `oh_patient_consensus` is the only table differing from develop, and that every INSERT in the file still matches its CREATE TABLE.

PR #1622, opened 2026-06-24
OP-1063 Remove the deprecated v1 String-based Laboratory insert methods

giuseppesorge:
Completes OP-1062 (the deprecation of the v1 String-based `Laboratory` insert methods, already merged) by removing them and keeping only the `LaboratoryRow`-based v2 variants.

Removed:
- `LabManager.newLaboratory(Laboratory, List<String>)`
- `LabManager.newLaboratory(List<Laboratory>, List<List<String>>)`
- protected `LabManager.newLabSecondProcedure(Laboratory, List<String>)`
- `LabIoOperations.newLabSecondProcedure(Laboratory, List<String>)`

**Behavioural note (please review):** the removed v1 `newLaboratory` had a procedure-2 guard `if (labRow == null || labRow.isEmpty()) throw OHDataValidationException(...)` that `newLaboratory2` did **not** have (v2 would NPE on `null` and silently accept an empty list). To preserve behaviour that guard is ported into `newLaboratory2` (case 2); the batch `newLaboratory2(List, List<List<LaboratoryRow>>)` inherits it via delegation. `LabIoOperations.newLabSecondProcedure2` is intentionally left unguarded, matching the original v1 IO method.

Tests (`org.isf.lab.Tests`): v1 tests that duplicated existing v2 tests were removed; the remaining v1 tests were migrated to the v2 methods via a small `toLaboratoryRows` helper. One pre-existing test (`testMgrNewLaboratory2ProcedureEquals2NullLabRows`) was actually calling the v1 method despite its name and is now wired to v2, so the ported guard is genuinely exercised. Result: 178 run, 0 failures, 2 skipped.

This is a **core + GUI + API** change pushed on the same branch name `OP-1063-remove-v1-lab-methods`, so the GUI/API CI cross-builds this core branch. The GUI (`LabEdit`/`LabEditExtended`) and API (`LaboratoryController`) callers are migrated to `newLaboratory2` in the sibling PRs (linked in a comment).

Note: this removes the methods in the same cycle they were deprecated. Happy to hold/retime if you would rather keep them deprecated for a release first.


giuseppesorge:
Part of OP-1063 (core + GUI + API), pushed on the same branch name `OP-1063-remove-v1-lab-methods` so the GUI/API CI cross-builds the core fork branch:

- core: informatici/openhospital-core#1622
- gui: informatici/openhospital-gui#2224
- api: informatici/openhospital-api#593

giuseppesorge:
Merged `develop` (63a94280). The conflict in `LabIoOperations.java` looked like a real one but it was not: this branch removes `newLabSecondProcedure(Laboratory, List<String>)`, while OP-1065 (#1612, now on develop) removed the two `getLaboratoryForPrint` overloads — different methods that simply sit next to each other in the file. Both removals are kept. `mvn compile` is clean and no reference to the removed method is left anywhere in the repository.

PR #1625, opened 2026-06-29
OP-1419 Fixes an issue where a NPE caused by using the JPMS  with >= JDK 16

dbmalkovsky:
See OP-1419.

Fixes an issue where a NPE caused by using the JPMS  (Java Platform Module System) on Java 16 or higher without granting the necessary deep reflection permissions to dcm4che. 

Added args to configuration to allow the deep reflection.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,6 +251,10 @@
 				<configuration>
 					<redirectTestOutputToFile>true</redirectTestOutputToFile>
 					<reportsDirectory>${project.build.directory}/test-reports</reportsDirectory>
+					<argLine>
+						--add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED
+						--add-opens java.base/java.io=ALL-UNNAMED
+					</argLine>
 				</configuration>
 			</plugin>
 			<plugin>


mwithi:
The point of this issue is that we need to add those options to the JVM arguments list, so we need to update the startup script as well, I guess... what do you propose @mizzioisf?

giuseppesorge:
The same two `--add-opens` are also needed on the application launch, otherwise the packaged GUI still hits this NPE at runtime — the surefire `argLine` here only covers the test JVM.

In `openhospital-gui` the three launchers start `org.isf.Application` directly, so the flags go on each `java` invocation:

```sh
# oh.sh (start_gui) and ohmac.sh (start_gui)
$JAVA_BIN -client --add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED \
  --add-opens java.base/java.io=ALL-UNNAMED \
  -Xms64m -Xmx1024m -Dsun.java2d.dpiaware=false ... org.isf.Application
```
```bat
REM oh.bat
%JAVA_BIN% -client --add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED ^
  --add-opens java.base/java.io=ALL-UNNAMED ^
  -Xms64m -Xmx1024m -Dsun.java2d.dpiaware=false ... org.isf.Application
```

The launcher uses `-classpath` (unnamed module), so `ALL-UNNAMED` is the right target, matching this PR. I can open a small GUI PR (`oh.sh` / `ohmac.sh` / `oh.bat`) with exactly these if that works for you. @mizzioisf @mwithi

mwithi:
> The same two `--add-opens` are also needed on the application launch, otherwise the packaged GUI still hits this NPE at runtime — the surefire `argLine` here only covers the test JVM...

Exactly @giuseppesorge. While we wait advice from @mizzioisf on this, can we assure that the two new options flag will not interfere with other existing behaviours?

mwithi:
Also, better to open and link a Jira issue for this fix, so we can include it in the release note.

giuseppesorge:
Thanks @mwithi.

On the safety question: the two `--add-opens` are non-interfering. `--add-opens <module>/<package>=ALL-UNNAMED` is purely additive — it only grants deep-reflection (`setAccessible`) access to code on the classpath; it doesn't remove or change any existing access, so code that doesn't use reflection is unaffected, and at most it turns a would-be `InaccessibleObjectException` into success. Both launchers start `org.isf.Application` from the classpath (the unnamed module), so `ALL-UNNAMED` is the correct target. It is also a standard option recognised on every JDK 9+ (we ship a JRE 17), so it won't cause an "unrecognized option" failure at startup. The only trade-off is a small relaxation of module encapsulation — a hardening/maintainability note, not a runtime-behaviour change — so it cannot break existing functionality.

On the Jira issue: agreed. As noted above, the same two flags are also needed at runtime (this PR's `argLine` only covers the test JVM), so they should go on the `org.isf.Application` launch in `oh.sh`, `ohmac.sh` and `oh.bat`. Happy to open an OP issue for this and link it here for the release notes, and to add the flags in a small GUI PR — I'll firm both up once @mizzioisf has weighed in on the approach.

giuseppesorge:
Done on the runtime side: opened **OP-1419** and GUI PR informatici/openhospital-gui#2226, adding the same two `--add-opens` to all four launchers (`oh.sh`, `ohmac.sh`, `oh.bat`, `oh.ps1`):

```
--add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED
--add-opens java.base/java.io=ALL-UNNAMED
```

Verified with the WRIX dataset attached to OP-219, exercising the GUI's `getImageFromDicom` decode path: without the flags it fails (`InaccessibleObjectException` on `FileCacheImageInputStream.cache` → null → NPE), with them the image decodes. Pairs with this PR's test-JVM `argLine`.

mwithi (approved):
LGTM. Thanks a lot!

PR #1630, opened 2026-07-01
OP-1300 Add a configurable lot-cost variance check

giuseppesorge:
## OP-1300 — Lot Costs variance check (core)

When `LOTWITHCOST=yes`, the application prompts for a cost when loading a new lot in the main store. A wrong cost gets stored in `MEDICALDSRLOT` and is currently unmodifiable, skewing stock-value estimation from then on. This adds a validation that flags a cost deviating significantly from the average cost of the previous lots of that same medical, so the operator can catch data-entry mistakes before saving.

### What this core PR adds
- **`GeneralData.LOTCOSTVARIANCEPERCENT`** — configurable percentage threshold (default `50`; `0` disables the check).
- **`MovStockInsertingManager.getAverageLotCost(Medical)`** — average of the previous lots' positive costs (`null` when there is no such history).
- **`MovStockInsertingManager.isLotCostWithinVariance(BigDecimal cost, BigDecimal averageCost)`** — pure threshold check; passes when the check is disabled or when there is no history to compare against.
- Unit tests in `medicalstock/Tests.java`: average (ignoring zero/no-cost lots), threshold (incl. the exact-boundary case), and the disabled/no-history paths.

### Defaults chosen (open for discussion in review)
- Threshold expressed as a **percentage** (`LOTCOSTVARIANCEPERCENT`, default `50`, `0` = off).
- Average taken over **all previous lots of the medical that have a positive cost**.
- Behaviour: **warn + ask confirmation**, re-prompting for the cost on decline (in the GUI PR).
- Scope limited to the **main-store new-lot charge** flow, per the ticket wording; the Inventory / Ward-rectify screens are intentionally left out — happy to extend if preferred.

### Companion PRs (same branch `OP-1300-lot-cost-variance`)
- gui: informatici/openhospital-gui#2228
- doc: informatici/openhospital-doc#288

> **Merge order:** this core PR should be merged first — the GUI PR builds against the new manager methods.

Jira: https://openhospital.atlassian.net/browse/OP-1300


src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -285,6 +286,54 @@ public List<Lot> getLotByMedical(Medical medical, boolean removeEmpty) throws OH
...
+			if (cost != null && cost.compareTo(BigDecimal.ZERO) > 0) {
+				sum = sum.add(cost);
+				count++;
+			}
+		}
+		if (count == 0) {
+			return null;
+		}
+		return sum.divide(BigDecimal.valueOf(count), 2, RoundingMode.HALF_UP);
+	}
```
mwithi, line 306: I would suggest calculating the average directly in the database using an aggregate query such as AVG(cost), filtering by medical code and positive costs.
The current implementation loads and materializes every matching Lot, including an unnecessary ordering by due date, only to calculate a single value in Java. Returning the aggregate directly would reduce data transfer, memory usage, and entity-management overhead, while scaling better as the lot history grows. The manager could still handle the null result, rounding policy, and configured threshold.

_As a separate improvement, using the median of the most recent N lots could make the check more representative when prices are affected by significant economic fluctuations, such as seasonal variations, supply-chain disruptions, or economic crises. Since this would require an additional configurable parameter and would change the functional rule, I think it would be better handled in a **separate issue**._

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -285,6 +286,54 @@ public List<Lot> getLotByMedical(Medical medical, boolean removeEmpty) throws OH
...
+		int threshold = GeneralData.LOTCOSTVARIANCEPERCENT;
+		if (threshold <= 0 || cost == null || averageCost == null || averageCost.compareTo(BigDecimal.ZERO) <= 0) {
+			return true;
+		}
+		BigDecimal deviationPercent = cost.subtract(averageCost).abs()
+			.multiply(BigDecimal.valueOf(100))
+			.divide(averageCost, 2, RoundingMode.HALF_UP);
+		return deviationPercent.compareTo(BigDecimal.valueOf(threshold)) <= 0;
+	}
+
```
mwithi, line 328: I would avoid rounding the percentage before comparing it with the threshold. The comparison can be performed without division:
```
BigDecimal difference = cost.subtract(averageCost).abs();

return difference.multiply(BigDecimal.valueOf(100))
    .compareTo(
        averageCost.multiply(BigDecimal.valueOf(threshold))
    ) <= 0;
```
This prevents values close to the boundary from being accepted or rejected because the calculated percentage was rounded first.

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
giuseppesorge: Applied in `41e59c66` — `getAverageLotCost` now delegates to a new repository aggregate `select avg(l.cost) from Lot l where l.medical.code = :medical and l.cost > 0`, so the lots are no longer materialized and the unnecessary due-date ordering is gone; the manager still handles the null result and the rounding/threshold.

Agreed the median-of-recent-N refinement changes the functional rule and is better handled as a separate issue whenever we want to pursue it.

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
giuseppesorge: Applied in `41e59c66` — the variance check now compares `|cost - average| * 100` against `average * threshold` directly, without dividing, so a value right at the boundary is no longer accepted or rejected because the percentage was rounded first.

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -285,6 +286,46 @@ public List<Lot> getLotByMedical(Medical medical, boolean removeEmpty) throws OH
...
+	public boolean isLotCostWithinVariance(BigDecimal cost, BigDecimal averageCost) {
+		int threshold = GeneralData.LOTCOSTVARIANCEPERCENT;
+		if (threshold <= 0 || cost == null || averageCost == null || averageCost.compareTo(BigDecimal.ZERO) <= 0) {
+			return true;
+		}
+		// Compare without dividing (and therefore without rounding the percentage first), so that a cost right at the
+		// variance boundary is not accepted or rejected because of a premature rounding: |cost - average| * 100 <= average * threshold.
+		BigDecimal difference = cost.subtract(averageCost).abs();
+		return difference.multiply(BigDecimal.valueOf(100))
+			.compareTo(averageCost.multiply(BigDecimal.valueOf(threshold))) <= 0;
```
dbmalkovsky, line 326: A nitpick:  `BigDecimal.valueOf()` takes a `long` as its parameter.  In lines 325 and 326 the parameter is an `int` which therefore requires an implicit conversion.  Line 325, could be `BigDecimal.valueOf(100L)` but to "fix" line 326 the data type of `GeneralData.LOTCOSTVARIANCEPERCEN` would need to change.  I'm not sure this really needs to be "fixed".  wdyt?

giuseppesorge: Agreed with your own reading: I would leave both as they are.

`BigDecimal.valueOf(long)` is the only integral overload, so an `int` argument always widens — a compile-time primitive widening conversion, no boxing, no allocation, no loss. Every one of the 8 `BigDecimal.valueOf(...)` calls in the module passes an `int` and none writes an `L` suffix, so `valueOf(100L)` would make this the one line in core that does.

And it would not remove the asymmetry, it would move it: line 325 would carry the suffix and line 326, right underneath, could not. Making that one match means changing `GeneralData.LOTCOSTVARIANCEPERCENT` to `long`, which reads worse than what it fixes — it is a percentage read from a settings file, validated as `< 0`, with an `int` default. A `long` would claim a range the value never has.

So: no change, unless you would rather I extract the `100` into a named constant, which is a different (and arguably better) reason to touch the line.

PR #1632, opened 2026-07-02
OP-1395 Manage Bill items and payments through JPA one-to-many relationships

giuseppesorge:
See [OP-1395](https://openhospital.atlassian.net/browse/OP-1395).

`Bill` now owns its items and payments through `@OneToMany(cascade = ALL, orphanRemoval = true)` collections (getters only, so the DTO mapper can never write them); the two `deleteWhereId` `@Modifying` queries are removed; `newBillItems`/`newBillPayments` are rewritten as a merge-sync (id 0 = insert, matching id = update in place, missing = orphan-removed — no more `setId(0)`); `updateBill` is load-and-copy; `newBill` rejects a nonzero id. Signatures are unchanged, so manager, API and GUI callers compile untouched.

Behavior changes to review (money path):
- Merge semantics: retained rows keep their database ids and audit columns (previously delete-all + reinsert). Tests updated accordingly.
- `updateBill` no longer compares the caller-supplied `@Version` lock for detached callers (last-write-wins; `@Version` still guards concurrent transactions). The GUI always sent lock 0, so the old path could spuriously fail repeated edits.
- `newBill` with a nonzero id now fails fast ("already exists") — it was an accidental scalar upsert, and with cascading collections a detached merge could have orphan-deleted stored rows.

Deliberate deviations from the ticket text:
- Child->bill stays EAGER: `getBills(List<BillPayments>)` returns bills the GUI renders detached; LAZY would throw LazyInitializationException.
- `optional = false` not applied: `BLI_ID_BILL`/`BLP_ID_BILL` are nullable in the production schema, and the implied inner join would hide legacy orphan rows. A NOT NULL migration could be a follow-up.
- `deleteBill` now also cascades at JPA level (row-by-row), correct on H2 as well as MariaDB.

Verified: full core suite green (2455 tests); API and GUI at develop compile unchanged against this core (no callers of the removed methods — grep + compile check); live end-to-end on MariaDB through `PUT /bills/{id}`: retained item kept its id with the updated amount, omitted items were orphan-removed from the table, the new item got a fresh IDENTITY id, totals recalculated correctly.

[OP-1395]: https://openhospital.atlassian.net/browse/OP-1395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/accounting/model/Bill.java
+++ b/src/main/java/org/isf/accounting/model/Bill.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
--- a/src/main/java/org/isf/accounting/model/Bill.java
+++ b/src/main/java/org/isf/accounting/model/Bill.java
@@ -23,8 +23,11 @@ package org.isf.accounting.model;
 
 import java.math.BigDecimal;
 import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
 
 import jakarta.persistence.AttributeOverride;
+import jakarta.persistence.CascadeType;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.EntityListeners;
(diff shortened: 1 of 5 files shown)


giuseppesorge:
Rebased onto current `develop` — the branch was 33 commits behind and its test file no longer compiled against OP-1216, which converted the Bill amounts to `BigDecimal` after this PR was opened.

Two changes on top of the original work:
- `5be2fc30` adapts the four assertions in `accounting/Tests.java` to `BigDecimal`, following the idiom the file already uses on `develop` — `isEqualByComparingTo(new BigDecimal("99.9"))` rather than a `isCloseTo`/`offset` pair, so there is no rounding before the comparison.
- `7942d2cb` is the copyright-year alignment discussed in #1602.

Full core suite green after the rebase: **2477 tests, 0 failures, 4 skipped**.

PR #1633, opened 2026-07-02
OP-887 Aggregate all patient-linked data for GDPR portability export

giuseppesorge:
See [OP-887](https://openhospital.atlassian.net/browse/OP-887).

Adds the aggregation side of the GDPR Art. 20 export: a `PatientExport` plain aggregate and a new `PatientExportManager` collecting, via the existing managers, the patient record plus admissions, OPD, laboratories, therapies, operations, vaccines, examinations and bills (with items and payments). Two small manager additions back it: `PatVacManager.getPatientVaccines(patientCode)` and `BillBrowserManager.getPatientBills(patID)`.

Notes for review:
- The aggregation lives in a dedicated manager rather than `PatientBrowserManager`: `TherapyManager` already constructor-injects `PatientBrowserManager`, so the reverse edge would create a constructor cycle.
- New permission `patient.export`, seeded by `sql/step_a128_add_permission_patient_export.sql` (P_ID_A 172, granted to `admin` as GP_ID 318) and mirrored in `load_demo_data.sql`. The step number and the two ids may need renumbering at merge: a118-a127 and the neighboring ids are reserved by other open PRs.
- Naming: the ticket says `patient.export` (singular); every existing patient permission is plural (`patients.*`). Happy to rename to `patients.export` if preferred — one-word change across the dependent PRs.
- Deliberately not exported (documented in the doc PR): profile photo, ward movements, visits, malnutrition, patient consensus, anamnesis, DICOM files. A follow-up ticket can add any of these if wanted.

Verified: new aggregation tests green (vaccines explicitly asserted, per the ticket note about the clinical sheet); the SQL was validated on a real MariaDB 10.6 for all three paths (fresh-install chain, upgrade-only step on an existing schema, full demo load).

Dependent PRs (same branch name for the fork cross-build): API, GUI, web UI and doc follow.

[OP-887]: https://openhospital.atlassian.net/browse/OP-887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Dependent PRs building on this branch: informatici/openhospital-api#595, informatici/openhospital-gui#2229, informatici/openhospital-ui#778, informatici/openhospital-doc#289. Suggested merge order: this PR first (api/gui/ui CI cross-build against the fork branch until then).

giuseppesorge:
Merged `develop` (31c105e7) to clear the conflict in `sql/step_04_all_following_steps.sql`: OP-1231 appended `step_a123_add_lot_management_in_pharmacy.sql` at the end of the file, so this branch's own `source` line now comes after it. Nothing else changed, and the branch's diff against `develop` is the same as before the merge.

sql/step_04_all_following_steps.sql
```diff
@@ -117,4 +117,4 @@ source step_a120_user_related_unique_constraints.sql;
 source step_a121_decimal_precision.sql;
 source step_a122_password_lease.sql;
 source step_a123_add_lot_management_in_pharmacy.sql;
-
+source step_a128_add_permission_patient_export.sql;
```
mwithi:
```suggestion
source step_aXXX_add_permission_patient_export.sql;
```

giuseppesorge:
Good catch, they are not needed. `P_ID_A` and `GP_ID` are `auto_increment` and `step_86` already inserts without them, so the script now lets the database assign both, and the group permission resolves the permission id with a subselect on `P_NAME` (b0a2489a).

They were also in the way: #1650 (OP-1434) claims the same 172/318, so on a database where both migrations run the second one stopped with `Duplicate entry 172 for key 'PRIMARY'`. Checked on MariaDB 12.3 seeded up to permission 171: with the ids dropped the two scripts apply in either order, 9 permissions and 9 group rows, no dangling reference. #1650 now carries the same change.

`load_demo_data.sql` keeps its explicit ids, being a full dump, so whichever of the two PRs merges second will need those demo rows renumbered - git raises that block as a conflict.

PR #1637, opened 2026-07-06
OP-1421 Fix initial lock value in native insert

mwithi:
See OP-1421.

Remove the columnDefinition from MedicalWard version field to keep the mapping Hibernate-friendly.


dbmalkovsky (approved):
LGTM

PR #1641, opened 2026-07-09
OP-886 Anonymize a patient's personal data (GDPR right to erasure)

giuseppesorge:
Implements the CORE part of **OP-886 – GDPR 1/4 Oblivion (Art. 17, right to erasure)**: irreversibly anonymize a patient's personal identification data while keeping the clinical records for statistical use.

### Changes
- `Patient`: new `PAT_ANONYMIZED` flag (+ getter/setter).
- `PatientIoOperations.anonymizePatient(Patient)`: overwrites the identifying fields with sentinels (first/second/mother/father name → `ANONYMIZED`, address/city/telephone → empty, tax code → `DELETED`), always detaches the profile photo (and deletes the filesystem copy when applicable), sets the flag and saves.
- `PatientBrowserManager.anonymizePatient(int)`: not-found / already-anonymized guards, best-effort removal of every linked DICOM series, then delegates the erasure.
- Migration `sql/step_aXXX_patient_anonymized.sql`: adds `PAT_ANONYMIZED`, registers the `patient.anonymize` permission (REST API) + the `btnpatientanonymize` menu grant (GUI), both granted to `admin`.
- Tests: `testMgrAnonymizePatient`, `testMgrAnonymizePatientAlreadyAnonymizedThrows` (full patient suite green).

### Notes
- Who/when is covered by the existing `Auditable` columns, no extra audit table.
- SQL step name is a `step_aXXX` placeholder — to be renumbered at merge.

### Related PRs (merge order: **core first**)
- api: informatici/openhospital-api#602
- gui: informatici/openhospital-gui#2231
- ui: informatici/openhospital-ui#779
- doc: informatici/openhospital-doc#291

### Verification
Migration applied and the manager exercised on a MariaDB 10.6 demo DB; anonymization overwrites the PII, keeps birth date/sex and all clinical records, and re-anonymizing is rejected.


giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`02184139`): added PAT_ANONYMIZED (default 0) to oh_patient plus the anonymize permission and menu rows. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`02184139`): added PAT_ANONYMIZED (default 0) to oh_patient plus the anonymize permission and menu rows. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

giuseppesorge:
Merged `develop` (26edc308). Besides `step_04`, `sql/load_demo_data.sql` conflicted as well: the dump is generated, and both sides had touched the same rows. Here there were four hunks: the `AUTO_INCREMENT` of `oh_groupmenu` (351 kept, since the merged table holds both new rows), the two `oh_groupmenu` rows themselves (349 and 350, both kept), the new `oh_patient` columns and the patient rows, which end up ordered `PAT_DELETED_BY`, `PAT_DELETED_DATE`, `PAT_ANONYMIZED`.

giuseppesorge:
Merged `develop` in to clear the conflict the merge back left (f3bd68a5).

The regenerated dump took `oh_patient` from 525 to 634 rows. I took develop's rows and added `PAT_ANONYMIZED` to all 634, rather than keeping the older set this branch carried.

The rows this branch adds to `oh_permissions`, `oh_menuitem`, `oh_groupmenu` and `oh_grouppermission` had merged cleanly and are untouched - checked that each still carries its extra row, and that `oh_patient` is the only table whose definition differs from develop.

PR #1642, opened 2026-07-09
OP-1231 Validate lots on update and add the Lot Management menu item

giuseppesorge:
Part of OP-1231 (Lot Management feature): https://openhospital.atlassian.net/browse/OP-1231

- Validate the lot on `MovStockInsertingManager.updateLot`: the due date cannot be before the preparation date and the cost cannot be negative; unit test `testMgrUpdateLotWithNegativeCostFails`.
- Seed the new **Lot Management** menu item under the Pharmacy sub-menu (`sql/step_aXXX_add_lot_management_in_pharmacy.sql` — `step_aXXX` placeholder per the settled step-numbering convention, to be numbered at merge).

The per-lot quantity stays computed — the persisted `LT_QTY` design is tracked in [OP-1428](https://openhospital.atlassian.net/browse/OP-1428), and the internal numeric lot id / editable lot number in [OP-1427](https://openhospital.atlassian.net/browse/OP-1427), as agreed on the issue.

Companion PRs: gui informatici/openhospital-gui#2232 · doc informatici/openhospital-doc#292

[OP-1428]: https://openhospital.atlassian.net/browse/OP-1428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1427]: https://openhospital.atlassian.net/browse/OP-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/test/java/org/isf/medicalstock/Tests.java
```diff
@@ -1615,4 +1616,41 @@ private void checkMedicalStockIntoDb(int code) {
...
+		firstUpdate.setDueDate(postponedDueDate);
+		Lot secondUpdate = movStockInsertingManager.updateLot(firstUpdate);
+		assertThat(secondUpdate.getCost()).isEqualByComparingTo(new BigDecimal("10.50"));
+		assertThat(secondUpdate.getDueDate()).isEqualTo(postponedDueDate);
+
+		// re-submitting the stale first instance must fail on the optimistic lock (LT_LOCK)
+		lot.setCost(new BigDecimal("11.00"));
+		assertThatThrownBy(() -> movStockInsertingManager.updateLot(lot))
+			.isInstanceOf(OHDataLockFailureException.class);
+	}
```
mwithi, line 1637: This has been introduced because a GUI glitch, the CORE was already returning an updated Lot object, so I think there is no need to test subsequent updates. I’d suggest removing `testMgrUpdateLotTwiceInARow`. Although it verifies optimistic-locking behavior, this PR does not change that behavior.

Instead, could we add a test covering the other validation introduced here—rejecting a lot whose due date is earlier than its preparation date? Currently, only the negative-cost validation is explicitly tested.

giuseppesorge: Done in c2e61043: removed `testMgrUpdateLotTwiceInARow` and added `testMgrUpdateLotWithDueDateBeforePreparationDateFails`, covering the due-date-earlier-than-preparation-date validation (only the negative-cost branch was tested before). Agreed the consecutive-update scenario belongs at the GUI layer rather than here.

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
```diff
@@ -406,6 +406,14 @@ public List<Lot> updateLot(List<Lot> lots) throws OHServiceException {
 	 * @throws OHServiceException if an error occurs during the check.
 	 */
 	public Lot updateLot(Lot lot) throws OHServiceException {
+		List<OHExceptionMessage> errors = new ArrayList<>();
+		validateLot(errors, lot, true);
+		if (lot.getCost() != null && lot.getCost().signum() < 0) {
```
dbmalkovsky, line 411: Does this mean a lot can have a cost of zero?  On the surface that seems strange to me but then what do I know .......

giuseppesorge: Yes, a lot can legitimately have a cost of zero — it's the historical default for a lot with no recorded price (donated/free stock, or when `LOTWITHCOST` is off, where lot costs are ignored entirely).

Where a positive cost is actually required, OH already enforces it contextually: at charge time, when `GeneralData.LOTWITHCOST` is enabled, `validateMovement` rejects a zero/absent cost (`angal.medicalstock.multiplecharging.zerocostsarenotallowed.msg`). `updateLot` deliberately only rejects a *negative* cost (a nonsensical value) rather than duplicating that charge-time policy, since editing a lot's dates/cost isn't necessarily a priced charge. It's also consistent with the OP-1300 variance check, which averages only lots with `cost > 0` (treating zero as "no price recorded").

Happy to tighten it to reject zero as well if you'd prefer, but I think keeping zero valid is the right behavior here.

giuseppesorge:
@dbmalkovsky — good call, thanks. Synced `load_demo_data.sql` for this change (`a3436ee3`): added the Lot Management OH_MENUITEM/OH_GROUPMENU rows. I regenerated it via a MariaDB round-trip of the migration (load the current demo dump → run the migration → mysqldump the changed tables) and verified it loads into a clean DB. I'll keep the demo dump in sync going forward.

dbmalkovsky:
I still can't start the app up after loading the demo data. I get 
<pre>[21/Jul/2026 14:10:45] [:] WARN - Error: 1054-42S22: Unknown column 'u1_0.us_deleted_by' in 'field list'
[21/Jul/2026 14:10:45] [:] WARN - SQL Error: 1054, SQLState: 42S22
[21/Jul/2026 14:10:45] [:] ERROR - (conn=70) Unknown column 'u1_0.us_deleted_by' in 'field list'
[21/Jul/2026 14:10:45] [:] ERROR - Exiting.
</pre>

which is caused by the demo data for this PR not including the new columns added by `step_a118_add_deleted_by_and_date.sql`

The demo data needs updates for all recent `step_a1*_*.sql` files for this to work

Note lines 8517-8531 in `load_demo_data.sql`; there is no mention of `US_DELETED_BY` or `US_DELETED_DATE`

giuseppesorge:
Thanks @dbmalkovsky — you're right, and digging in it turned out to be a bit deeper than the columns from this PR. I fixed OP-1231's demo data (`bb0badb6`) and want to flag the root cause so we handle it at the right level.

**What I did here:** the `us_deleted_by` failure comes from the soft-delete columns in `step_a118` (`*_DELETED_BY`/`*_DELETED_DATE` on admission, supplier, patient, user, usergroup) plus the password columns in `step_a122` (`US_PASSWD_LAST_CHANGED`/`US_PASSWD_MUST_CHANGE`). This branch carries both, but `load_demo_data.sql` was still at the pre-a118 schema. I regenerated the affected demo tables via a MariaDB round-trip of those migrations and verified the dump now loads into a clean DB — so startup should get past that error.

**The root cause is broader, though:** `step_a118`/`step_a122` come from OP-1005 (#1603) and OP-896 (#1596), which were *merged into develop* back in July — but they're **not on develop anymore** (the merge commits aren't reachable from the current develop HEAD, so develop looks like it was reset/rewritten). ~14 of my open branches were cut from that post-merge base, so they all carry a118–a122 while develop doesn't, and none of their demo data reflects those columns. So this isn't really per-PR — patching 14 stale branches would be churn that a rebase would undo.

Do you know if that develop reset was intentional (release re-org)? If the branches should be rebased onto the current develop, the cleaner fix is to re-establish a118–a122 (with the demo-data update) on develop once, and let the branches pick it up — rather than me patching each branch. Happy to do whichever you prefer; just didn't want to spray the same demo-data delta across 14 branches if they're about to be rebased.

giuseppesorge:
Verified end-to-end on a clean DB just now: loading `create_all_demo.sql` (which sources `load_demo_data.sql`) into a fresh MariaDB 10.6 completes with no errors, `oh_user` carries both the a118 columns (`US_DELETED_BY`/`US_DELETED_DATE`) and the a122 ones (`US_PASSWD_LAST_CHANGED`/`US_PASSWD_MUST_CHANGE`), and the GUI boots straight into Single User mode — no more `Unknown column 'us_deleted_by'`. So this branch is unblocked.

The develop-reset question above still stands for the other stale branches, though — happy to go whichever way you prefer there.

giuseppesorge:
Correction to my comment above — I got the develop part wrong, sorry. There was **no reset**: informatici/`develop` (HEAD `f4940d04`) does carry `step_a118`–`a122`. I'd been reading my own stale fork's `develop` by mistake.

The real gap is narrower and lives on develop itself: its `load_demo_data.sql` was never updated for those migrations — 0 of the a118/a122 columns (`*_DELETED_BY`/`_DATE`, `US_PASSWD_LAST_CHANGED`/`_MUST_CHANGE`) in any table. So loading develop's demo dump into a clean DB and starting the app fails with the same `Unknown column 'us_deleted_by'`; CI misses it because it builds the schema from the step scripts, not the dump. This branch is already fixed (`bb0badb6`), but the proper fix belongs on develop so every branch inherits it — working out the cleanest way to land that with @dbmalkovsky.

giuseppesorge:
Opened #1654 against `develop` with the fix at the source: it re-syncs `load_demo_data.sql` for the `a118`–`a122` migrations (which is why a fresh demo DB was failing at startup) plus clears the demo admin's `US_LAST_LOGIN` so it doesn't idle-lock on first login. Tracked as OP-1436. There's a scope note on the PR about `a121` making the diff large — happy to narrow it if you prefer.

dbmalkovsky (commented):
LGTM

dbmalkovsky (dismissed):
LGTM

src/main/java/org/isf/medicalstock/manager/MovStockInsertingManager.java
mwithi: Yes, the zero cost is the "legacy" value for "lot without costs". It would have been better to adopt `null` values in such cases, and maybe we can do that at latter time, but as for now I would suggest not changing this behaviour within OP-1231.

mwithi:
> Opened #1654 against `develop` with the fix at the source: it re-syncs `load_demo_data.sql` for the `a118`–`a122` migrations (which is why a fresh demo DB was failing at startup) plus clears the demo admin's `US_LAST_LOGIN` so it doesn't idle-lock on first login. Tracked as [OP-1436](https://openhospital.atlassian.net/browse/OP-1436). There's a scope note on the PR about `a121` making the diff large — happy to narrow it if you prefer.

Yes, in general, when DB is touched, also demo data should be updated along with the DB scripts. This can be tracked with the "demo_data" label on the Jira issue. Thanks!

[OP-1436]: https://openhospital.atlassian.net/browse/OP-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

giuseppesorge:
Filled in the SQL step placeholder as requested in informatici/openhospital-gui#2232: `step_aXXX_add_lot_management_in_pharmacy.sql` → `step_a123_add_lot_management_in_pharmacy.sql`, with the `source` line in `step_04_all_following_steps.sql` updated (`b7a30d96`). `a123` is the first free slot — `develop` stops at `a122` and the other open PRs claim `a124`/`a125`/`a128` — so no renumbering is needed whatever merges next. Content of the script is unchanged, and the `MNI_CLASS` FQCN matches `org.isf.lotmanagement.gui.LotBrowser` in both the step script and `load_demo_data.sql`. Ready to merge from my side.

mwithi (approved):
LGTM. Thanks a lot!

PR #1643, opened 2026-07-10
OP-1216 Convert FLOAT/DOUBLE money and quantity columns to DECIMAL/INT

giuseppesorge:
> Replaces #1597 — identical tree (verified diff-empty), resubmitted on a clean linear branch because approvals on #1597 kept being auto-dismissed by a review glitch. Full review history in #1597.

## OP-1216 — Convert FLOAT and DOUBLE to DECIMAL for exact precision

Converts the monetary and exact-quantity fields from binary floating point
to `java.math.BigDecimal` (DECIMAL(19,2)), and the pure unit-count fields to
`int`, to guarantee exact precision for financial and stock values.

### Changes
- **Models → BigDecimal**: `Price.price`, `Bill.amount/balance`, `BillItems.itemAmount`,
  `BillPayments.amount`, `MovementWard.quantity`, `MedicalWard.out_quantity`,
  `Lot.cost` (now annotated `precision=19, scale=2`), and
  `MedicalInventoryRow.theoreticQty/realQty` (fields initialized to `ZERO`,
  matching the `NOT NULL DEFAULT 0` columns).
- **Models → int**: `Medical` in/out/min quantity, `MedicalWard.in_quantity`.
- **Managers/services**: arithmetic reworked on `BigDecimal` — zero checks use
  `compareTo`/`signum`, monetary results keep `setScale(2, HALF_UP)`, and the price-list
  step rounding is reproduced exactly. The ward-stock repository out-quantity
  queries now take `BigDecimal` (no `double` round-trip on the DECIMAL column);
  the in-quantity queries take `int`. `MedicalInventoryManager` compares and
  adjusts inventory quantities in `BigDecimal` end-to-end, with no
  `doubleValue()`/`BigDecimal.valueOf(double)` round-trips left.
- **Validation**: a fractional ward-to-ward transfer is rejected on the incoming
  (integer) side, and `confirmMedicalInventoryRow` validates each theoretic/real quantity with
  `intValueExact` (main-store stock is integer-only), rejecting fractional or
  out-of-int-range values instead of silently truncating them — including
  fractional quantities whose difference happens to be whole.
- **SQL**: `sql/step_a121_decimal_precision.sql` (numbered after the merged
  OP-1099 step, per the agreed merge-order numbering), registered in
  `step_04_all_following_steps.sql`.
- Tests updated for the new types (`BigDecimal` fixtures, `isEqualByComparingTo`).

### Notes / decisions
- `MDSRWRD_IN_QTI` stays **INT** and `MDSRWRD_OUT_QTI`/`MMVN_MDSR_QTY` become
  **DECIMAL**, following the ticket's field list; fractional ward-to-ward
  transfers are rejected by validation instead of being truncated.
- `MDSR_INI_STOCK_QTI` is **not** converted: it is not in the ticket's field list
  and its field (`Medical.initialqty`) is `@Deprecated`; converting only the column
  would have left a double-over-int mismatch.
- The main stock-movement quantity (`MMV_QTY`, `OH_MEDICALDSRSTOCKMOV`) becomes
  `INT` as listed.

### Verification
- Full core test suite green (2458 tests).
- Applied the migration to a demo-data DB (old FLOAT/DOUBLE schema, ~1.5k prices,
  bills, stock movements) on MariaDB 10.6 with production flags: all columns
  converted to the expected DECIMAL/INT types with **no data degradation**
  (values numerically identical pre/post; no fractional values in any to-INT column).
- GUI end-to-end on the migrated demo DB: a ward inventory row saved with a
  fractional real quantity persists as an exact `1.50`; a ward rectification of
  `2.5` persists as an exact `-2.50` movement.

Companion PRs (same branch name for CI cross-build) are linked in a comment below.

JIRA: https://openhospital.atlassian.net/browse/OP-1216


mwithi (approved):
LGTM. Thanks a lot!

PR #1646, opened 2026-07-16
OP-1427 Introduce an internal numeric lot id (stage 1)

giuseppesorge:
OP-1427 (Introduce an internal numeric lot ID) — **stage 1**: https://openhospital.atlassian.net/browse/OP-1427

Follow-up of OP-1231. Introduces an auto-generated numeric `LT_ID` as the primary key of `OH_MEDICALDSRLOT` and moves the referencing tables (movements, ward stock, inventory) to the numeric foreign key. **Zero behaviour change in this stage**: the printed lot number `LT_ID_A` stays globally UNIQUE and remains the business identifier shown to users; `Lot.equals/hashCode` stay code-based. Making the printed number editable / non-unique is a deliberate stage 2 (separate change), to keep this reviewable.

- `Lot.id` becomes the IDENTITY `@Id`; `LT_ID_A` is demoted to a plain unique attribute. All public manager/service signatures keep taking the lot code and resolve internally (`resolveLotId` resolves by code, so update/store/delete reproduce the pre-swap semantics exactly — verified no gui/api caller signature changes).
- The report-facing varchar columns `MMV_LT_ID_A` (movements) and `MINVTR_LT_ID_A` (inventory rows) are **kept and dual-written** by the entities, because the GUI Jasper reports (ProductLedger, PharmaceuticalStock, AMC, Expiration, StockWard, Inventory…) still join on them; stage 2 will rewrite those reports to the numeric FK. The columns with a real FK and no report usage (`MDSRWRD_LT_ID_A`, `MMVN_LT_ID_A`) are dropped.
- `sql/step_aXXX_internal_lot_id.sql` + `load_demo_data.sql` updated so a migrated DB and a fresh demo install end on the identical schema. Verified live on MariaDB 10.6 (PK swap, all FKs backfilled, `SHOW CREATE TABLE` identical); full core suite green (2537).

**Merge order:** OP-1428 (openhospital-core#1645) must merge **before** this — its LT_QTY backfill joins on `MDSRWRD_LT_ID_A`, which this migration drops. This PR then rebases and its `step_aXXX` is numbered after OP-1428's.


giuseppesorge:
Merged `develop` (85b17063) to clear the conflict in `sql/step_04_all_following_steps.sql`: OP-1231 appended `step_a123_add_lot_management_in_pharmacy.sql` at the end of the file, so this branch's own `source` line now comes after it. Nothing else changed, and the branch's diff against `develop` is the same as before the merge.

giuseppesorge:
Merged `develop` in to clear the conflict the merge back left (a9c1e96c).

The regenerated dump took the lot table from 48 to 93 lots, and the tables referencing it grew with it. The textual merge did not survive that: two tables came out with a value missing against their definition, and the three `oh_medicaldsrinventoryrow` rows were dropped altogether.

So I re-applied the shape this branch already uses rather than hand-merging it. The dump never carries the numeric ids: it inserts by lot code and lets the database resolve them - `INSERT ... SELECT ... WHERE LT_ID_A = '<code>'` for the two tables where the varchar column goes away, and the two trailing `UPDATE ... JOIN` for the ones where it stays. Applied to develop's data that is 93 lots, 356 movements, 147 ward rows, 134 ward movements and 3 inventory rows.

Checked before pushing:

* every lot code referenced by the rewritten rows resolves to a lot in the dump - no orphans, and no row had a NULL code to begin with;
* the column list of each rewritten `INSERT` matches its table definition, in the same order;
* the five table definitions differ from develop only by the lot id changes - develop had not touched them otherwise;
* the remaining 66 tables still have every `INSERT` matching its `CREATE TABLE`.

Still true and worth repeating at merge time: the header of this branch's migration notes that it drops `OH_MEDICALDSRWARD.MDSRWRD_LT_ID_A`, which the OP-1428 backfill in #1645 joins on, so #1645 has to be numbered and merged before this one.

PR #1647, opened 2026-07-16
OP-1430 Expose the password policy from the core manager

giuseppesorge:
Part of OP-1430 (align password validation with a single, backend-owned policy): https://openhospital.atlassian.net/browse/OP-1430

Extracts the strength regex from `UserBrowsingManager.isPasswordStrong` into a reusable `PASSWORD_STRENGTH_REGEX` constant + a pre-compiled `Pattern` (cheap perf win, and it becomes introspectable), and adds three accessors — `isStrongPasswordEnabled()`, `getPasswordMinLength()`, `getPasswordStrengthRegex()` — mirroring the existing `getPasswordLeaseDays()`, so the API layer can serve the effective password policy without reaching into `GeneralData` statics.

No behaviour change: `isPasswordStrong`/`isPasswordValid` are functionally unchanged. Unit tests cover the accessors and assert the exposed regex accepts/rejects the same passwords the strength check enforces.

Companion PRs: api informatici/openhospital-api#607 · ui informatici/openhospital-ui#783
Builds on OP-896 (merged).

### DIFF
--- a/src/main/java/org/isf/menu/manager/UserBrowsingManager.java
+++ b/src/main/java/org/isf/menu/manager/UserBrowsingManager.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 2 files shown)


giuseppesorge:
Merged `develop` in to clear the conflict the merge back left (d2f3933f).

Both sides had added a test method at the same point of `TestUserBrowsingManager`: this branch the two password policy accessor tests, develop the `isSameAsCurrentPassword` one that came in with OP-1431. All three are kept - the conflict markers just happened to cut across the method boundary.

`mvn test-compile` passes locally.

PR #1648, opened 2026-07-16
OP-1431 Reject reusing the current password on a password change

giuseppesorge:
## What

Adds `UserBrowsingManager.isSameAsCurrentPassword(User user, String newPassword)`, which BCrypt-checks a candidate new password against the user's stored hash, so callers can reject reusing the current password on a change.

## Why

Follow-up to the password-policy work (OP-896 / OP-1430): a user forced to renew their password must not be able to set it back to the current one. Centralising the check in the core lets the API and the Swing GUI share a single implementation.

## Notes

- Returns `false` for a null user, a null/empty candidate, or a user without a stored hash, so the check is a safe no-op in those cases.
- Unit test in `TestUserBrowsingManager` covers match, mismatch, empty/null inputs and a user without a password.

Jira: OP-1431

---
Related PRs: openhospital-core #1648 · openhospital-api #608 · openhospital-gui #2236


dbmalkovsky (dismissed):
LGTM; successfully tested against GUI and API

mwithi (approved):
Reviewed and approved by @dbmalkovsky

giuseppesorge:
@dbmalkovsky — apologies: my copyright commit dismissed your approval here and on gui#2236, and you had already tested all three. That was my doing and I should have thought it through before pushing.

Context: after your comment on core#1602 I applied the "all or none" rule to my other open PRs that were internally inconsistent, and this one was among them. api#608 was uniform on 2025 so I had left it alone — which left OP-1431 split across the three repos, so I have now aligned that one too (`4393980`). All three are consistently `2006-2026` on the files they change.

The three commits are copyright headers only, nothing else. When you have a moment, could you re-approve? Happy to drop the commits instead and put the ticket back exactly as you left it — your call, just say the word.

PR #1649, opened 2026-07-17
OP-1403 Add optional startup compilation of stale JRXML reports

giuseppesorge:
## OP-1403

Adds an opt-in mechanism to (re)compile JasperReports templates (`.jrxml`) into their compiled form (`.jasper`), so that a template edited on disk does not require a manual recompilation and does not fail at report time with a stale or missing `.jasper`.

### Changes
- New `org.isf.stat.manager.JasperReportCompiler`: scans the given report folders (recursively) and (re)compiles every `.jrxml` whose `.jasper` is missing or older than the template. A template that fails to compile is logged and skipped, so a single broken report does not stop the others.
- New configuration flag `GeneralData.AUTOCOMPILEJRXML` (default `false`) that guards the feature.
- Unit test `TestJasperReportCompiler`: compiles a valid JasperReports 7 template, skips up-to-date and recompiles stale reports, and verifies a failing template does not stop the others.

### Notes
- No new dependency: the project's JasperReports 7.0.6 already compiles templates at runtime.
- The compiler expects templates in the JasperReports 7 format; legacy JasperReports 6 templates must be migrated first (e.g. with Jaspersoft Studio).
- Security: JRXML templates can contain executable expressions, so the feature is disabled by default and must only be enabled when the report files are trusted.

The GUI startup hook and the setting are in openhospital-gui; the documentation is in openhospital-doc.

### DIFF
--- a/src/main/java/org/isf/generaldata/GeneralData.java
+++ b/src/main/java/org/isf/generaldata/GeneralData.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
@@ -41,6 +41,7 @@ public final class GeneralData extends ConfigurationProperties {
 	public static boolean AUTOMATICLOT_OUT;
 	public static boolean AUTOMATICLOTWARD_TOWARD;
 	public static boolean LOTWITHCOST;
+	public static boolean AUTOCOMPILEJRXML;
 	public static String PATIENTSHEET;
 	public static String VISITSHEET;
 	public static String EXAMINATIONCHART;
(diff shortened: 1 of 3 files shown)


giuseppesorge:
Small addition to this PR: a headless **JasperReports 6 → 7 migration** helper, `JasperReportCompiler.migrateToJasperReports7(...)`.

Context: `compileStaleReports` can only compile templates in the JR7 format — JasperReports 7.0.6 cannot compile the legacy JR6-format `.jrxml` that OH still ships (they need a format migration, which normally means opening each one in Jaspersoft Studio). The new helper does that migration headless: a compiled `.jasper` already holds a JR7 model, so it loads the committed `.jasper`, writes it back with `JRXmlWriter` as a JR7 `.jrxml` (no legacy namespace), and recompiles it with the pom's JR version. After migration, `compileStaleReports` keeps the `.jasper` up to date.

This is also the answer to the earlier "a Maven task to compile the jrxml with the pom's JR version" request: the reusable primitive now lives here and can be driven from a build step.

I used it to migrate OP-892's aggregate reports to JR7 (gui#2237) — text-render output was byte-identical to the previous `.jasper` on every report, so it's a faithful round-trip. Added a unit test (`testMigratesCompiledReportToJasperReports7`); the class' tests are green (6/6).

PR #1650, opened 2026-07-19
OP-1434 Add API authority permissions for billing and stock movements

giuseppesorge:
## What

Adds the authority permissions that the API authorization rules require for the billing and stock movement endpoints (see the companion PR informatici/openhospital-api#610).

`/bills/**` and `/stockmovements/**` previously had no operation-specific authorization rule in the API, and no clean `bills.*` / `stockmovements.*` permissions existed (billing only had GUI menu-item codes, and `/stockmovements` is distinct from the already-secured `/medicalstockmovements`). Without these permissions, no user could be granted the new authorities.

## Changes

- New permissions `bills.create/read/update/delete` and `stockmovements.create/read/update/delete` (`P_ID_A` 172–179), granted to the `admin` group (`GP_ID` 318–325) so existing supported workflows keep working for appropriately configured users.
- Added via a new migration `sql/step_aXXX_add_bills_and_stockmovements_api_permissions.sql` (sourced last in `step_04_all_following_steps.sql`) and to `load_demo_data.sql`. The frozen `step_98_reset_permissions.sql` baseline is intentionally left untouched.

The `step_aXXX` file name and the `P_ID_A` / `GP_ID` values are placeholders and can be renumbered at merge time according to merge order.

## Testing

Migration verified on a real MariaDB 10.6 (`lower_case_table_names=1`): loaded the current demo baseline (171 permissions), applied the migration, confirmed the 8 permissions and 8 admin grants are inserted with valid foreign keys and no collisions.

Related to OP-1434. Merge this before the API PR (the API authorities resolve from these permissions at runtime).


giuseppesorge:
Merged `develop` (c34f2d36) to clear the conflict in `sql/step_04_all_following_steps.sql`: OP-1231 appended `step_a123_add_lot_management_in_pharmacy.sql` at the end of the file, so this branch's own `source` line now comes after it. Nothing else changed, and the branch's diff against `develop` is the same as before the merge.

giuseppesorge:
Following @mwithi's question on #1633, the explicit `P_ID_A`/`GP_ID` are gone here too (d2c8fefb): both columns are `auto_increment` and `step_86` already inserts without them, so the eight group permissions now resolve their permission ids with a subselect on `P_NAME`.

The literals were not harmless: this script and #1633 both claimed 172/318, so whichever migration ran second stopped with `Duplicate entry '172' for key 'PRIMARY'`. Checked on MariaDB 12.3 seeded up to permission 171: both scripts now apply in either order, with no dangling reference.

`load_demo_data.sql` still carries explicit ids, being a full dump, so the PR that merges second will need those demo rows renumbered.

PR #1654, opened 2026-07-22
OP-1436 Sync load_demo_data.sql with the a120 and a121 schema

giuseppesorge:
Fixes [OP-1436](https://openhospital.atlassian.net/browse/OP-1436).

> Reworked after the release, as asked. The regenerated dump absorbed most of what this branch was syncing, so the scope is now a120 and a121 only.

`sql/load_demo_data.sql` is a standalone mysqldump, so a schema migration has to be mirrored into it or a freshly loaded demo database ends up with a different schema than a migrated one.

## What the release already absorbed

Checked one migration at a time against `develop`'s regenerated dump:

| migration | what it adds | in the dump |
| --- | --- | --- |
| a118 | ten `*_DELETED_BY` / `*_DELETED_DATE` columns | already in, all ten |
| a119 | the two movement log tables | already in |
| a122 | the two password lease columns | already in |
| a123 | the lot management menu item and group menu row | already in |
| a124 | the admission and examination sex/age columns | already in |
| **a120** | two unique constraints | **missing** |
| **a121** | fifteen column type changes | **missing** |

So this branch now carries only a120 and a121.

## What is left

Two constraints:

* `UX_PERMISSIONS_NAME` on `oh_permissions (P_NAME)`
* `UX_GP_GROUP_PERM` on `oh_grouppermission (GP_UG_ID_A, GP_P_ID_A)`

and fifteen columns that are still `double` or `float` in the dump where a121 makes them `decimal(19,2)` or `int`: `PRC_PRICE`, `BLL_AMOUNT`, `BLL_BALANCE`, `BLI_ITEM_AMOUNT`, `BLP_AMOUNT`, `LT_COST`, `MMVN_MDSR_QTY`, `MDSRWRD_OUT_QTI`, `MINVTR_THEORETIC_QTY`, `MINVTR_REAL_QTY`, `MDSRWRD_IN_QTI`, `MDSR_MIN_STOCK_QTI`, `MDSR_IN_QTI`, `MDSR_OUT_QTI`, `MMV_QTY`.

The definitions are the ones this branch already carried, which came from a genuinely migrated database - I checked all fifteen match what it had before, character for character.

## Checked

* the demo data already satisfies both unique constraints - no duplicate permission name among 171, no duplicate group/permission pair among 312 - so the de-duplicating `DELETE` that a120 performs has nothing to do here and the constraints can be added as they are;
* none of the five columns turning into `int` holds a non-integer value, so nothing is rounded away;
* every `INSERT` in the file still matches its `CREATE TABLE`, across all 66 tables;
* the whole change is 17 added lines - only the definitions move, the values are left as they are and MySQL coerces them on load;
* the only tables that differ from `develop` are the twelve this change touches.

[OP-1436]: https://openhospital.atlassian.net/browse/OP-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
This one is a decision rather than a conflict to resolve, so I stopped before touching it.

`develop` already carries the a118-a122 sync: comparing the two dumps, the 64 tables they have in common declare identical column lists. The part that actually mattered — the columns, whose absence breaks a clean demo database — is already in.

What is left of this PR is two tables the dump does not contain, `oh_medicaldsrstockmov_log` and `oh_medicaldsrstockmovward_log`, plus a few thousand lines of churn from regenerating the dump (timestamps and row order). Those two tables are created by `step_a119`, which is sourced from `step_04`, and `load_demo_data.sql` never drops them, so a clean demo database ends up with them (empty) either way. It is completeness, not correctness.

So either this PR is closed, or it is reduced to just adding those two tables, which would also make it reviewable. Happy to do whichever you prefer.

mwithi:
We need to rework it after the release because demo data will change a lot with more data

giuseppesorge:
Reworked after the release, as you asked (80771cfe).

The regenerated dump turned out to have absorbed most of what this branch was syncing, so rather than redoing the whole thing I checked one migration at a time against develop's dump: a118's ten deleted_by/date columns, a119's two log tables and a122's two password lease columns are all in already, and so are a123 and a124.

What is still missing is a120 and a121, and that is all the branch carries now - the two unique constraints and the fifteen columns that are still `double` or `float` where the migration makes them `decimal(19,2)` or `int`. Title and description updated to match.

Two things I checked before pushing, since both could have bitten on load: the demo data already satisfies both unique constraints, so the de-duplicating DELETE in a120 has nothing to do here; and none of the columns turning into `int` holds a non-integer value, so no value is rounded away.

One judgement call worth your opinion: 17 of the 2492 added lines are the actual change, the rest is the values of those columns re-rendered with two decimals, which is what mysqldump emits for a `decimal(19,2)` column. I kept it for fidelity with a real dump, but I am happy to drop it and leave the values untouched if you would rather review a small diff - MySQL coerces them on load either way.

giuseppesorge:
Dropped the decimal re-rendering, as suggested - the branch now changes only the definitions and leaves the values untouched, so the diff is 17 lines instead of 2492. MySQL coerces the values on load either way.

Force-pushed rather than stacking a second commit, since the first one was only minutes old and a "did it then undid half of it" history would have been worse to read. Description updated to match.

mwithi (approved):
LGTM. Thanks a lot!

PR #1664, opened 2026-07-31
Reject negative bill item values

diegoforni:
## What

- reject bill items with a null or negative amount
- reject bill items with a negative quantity
- enforce the invariant in `BillBrowserManager` before bill writes
- add update regressions for both manipulated fields

## Why

Bill totals are recalculated from item amount multiplied by quantity. Negative values were accepted during bill updates and could therefore corrupt the bill amount and balance.

## Security impact

Malformed bill items now fail with `OHDataValidationException` before the bill or its items are updated. The check lives in the domain manager so it protects REST and non-REST callers. Zero remains allowed to avoid changing existing business semantics beyond the confirmed negative-value vulnerability.

## Checks

- `./mvnw -q -Dtest=org.isf.accounting.Tests test` (61 tests)
- `git diff --check`

### DIFF
--- a/src/main/java/org/isf/accounting/manager/BillBrowserManager.java
+++ b/src/main/java/org/isf/accounting/manager/BillBrowserManager.java
@@ -85,6 +85,12 @@ public class BillBrowserManager {
 		if (billItems.isEmpty()) {
 			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.newbill.abillmustcontainatleastoneitem.msg")));
 		}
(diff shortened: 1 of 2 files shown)


mwithi:
Closing because negative bill items are specifically allowed in the application see https://github.com/informatici/openhospital-doc/blob/develop/doc_user/UserManual.adoc#6215-insert-a-refund-refund and https://github.com/informatici/openhospital-doc/blob/develop/doc_user/UserManual.adoc#12211-other-prices-other-prices (Discharge)

PR #1677, opened 2026-08-27
OP-1461 Report an unknown log folder instead of throwing a NullPointerException

giuseppesorge:
## [OP-1461](https://openhospital.atlassian.net/browse/OP-1461)

Found while testing the v1.15.1 pre-release, so this targets `release_1_15_1`. Rebuilt on that base rather than cherry-picked.

Opening the log folder from the GUI kills the action with an unhandled exception on the event dispatch thread, and the user is told nothing:

```
Exception in thread "AWT-EventQueue-0" java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.isf.menu.gui.MainMenu.launchApp(MainMenu.java:445)
Caused by: java.lang.NullPointerException
	at java.base/java.io.File.<init>(Unknown Source)
	at org.isf.utils.log.LogUtil.openLogFileLocation(LogUtil.java:70)
	at org.isf.help.LogViewer.<init>(LogViewer.java:46)
```

`openLogFileLocation()` dereferenced two values that can legitimately be null:

* `getLogFileAbsolutePath()` returns null when none of the root logger's appenders writes to a file — `TestLogUtil` already says as much, *"depending on the testing setup the value is either null or the value in the properties file"*;
* `getParent()` returns null for a file name with no folder, so `appender.rolling.fileName=openhospital.log` reaches the same crash by another route.

The resolution moved into `logFileFolder(String)`, which throws `IOException` — the exception the caller already handles — when the path is unknown, and resolves a bare file name against the working directory, where the file is actually written.

## Verified

`TestLogUtil` covers the three cases, 4 tests green.

The GUI side is in openhospital-gui and needs this merged first, since the null is produced here.

[OP-1461]: https://openhospital.atlassian.net/browse/OP-1461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/utils/log/LogUtil.java
+++ b/src/main/java/org/isf/utils/log/LogUtil.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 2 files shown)


mwithi (approved):
LGTM. Thanks a lot!

PR #1683, opened 2026-09-01
Merge back after release 1.15.1

mwithi:


sql/load_demo_data.sql
```diff
@@ -1,9 +1,8 @@
-/*M!999999\- enable the sandbox mode */ 
--- MariaDB dump 10.19  Distrib 10.6.23-MariaDB, for linux-systemd (x86_64)
+-- MariaDB dump 10.19  Distrib 10.6.28-MariaDB, for Win64 (AMD64)
 --
 -- Host: localhost    Database: ohdemo
 -- ------------------------------------------------------
--- Server version	10.6.23-MariaDB
+-- Server version	10.6.13-MariaDB
```
dbmalkovsky: I realize this is just a comment but I was wondering my the down grading of versions from `23` to `13` is about?  In the 1.15.1 track it is still `23`

sql/step_a108_audit_dicom_type_and_dicom_data.sql
```diff
@@ -9,7 +9,7 @@ ADD COLUMN DCMT_ACTIVE TINYINT(1) NOT NULL DEFAULT 1;
 
 -- alter table oh_dicom_data --------------
 
-ALTER TABLE `oh_dicom_data`
+ALTER TABLE OH_DICOM_DATA
```
dbmalkovsky, line 12: why was the same change (upper casing and removing the back quotes) not done in `step_a111_add_missing_lock_columns.sql`

dbmalkovsky (commented):
LGTM given someone looks at my two comments

pom.xml
```diff
@@ -20,7 +20,7 @@
 	<groupId>org.isf</groupId>
 	<artifactId>OH-core</artifactId>
 	<packaging>jar</packaging>
-	<version>1.15.0</version>
+	<version>1.16.0-SNAPSHOT</version>
```
dbmalkovsky, line 23: Why this value? In the pom.xml in the GUI PR it is 1.15.1

sql/load_demo_data.sql
mwithi: yes, good catch! Thanks!

sql/step_a108_audit_dicom_type_and_dicom_data.sql
mwithi: Just oversight... thanks!

pom.xml
mwithi: it will be updated also there, thanks!

dbmalkovsky (approved):
LGTM

PR #1684, opened 2026-09-02
Fix demo bills after the merge back

giuseppesorge:
Follow-up to #1683, on top of `74ccd71f`.

**It targets `release_1_15_1`, so it needs to go in before #1683 is merged**: merging it here updates #1683's head and the fix travels into `develop` with the merge back. If #1683 is merged first, the duplicated rows land on `develop` and this has to be redone against `develop` instead.

### What happened

[Merge branch 'develop' into release_1_15_1](https://github.com/informatici/openhospital-core/pull/1683/commits/fb787c88ede86c9633ba7d1a5bd33d55c96f30aa) brought back the bill items and payments that [Update demo data fixing empty bills and updating pricelists](https://github.com/informatici/openhospital-core/pull/1683/commits/7987a1c7b632e8d4065503996986645c1a69aa93) had deleted when it rewrote them, so `sql/load_demo_data.sql` ended up carrying both the old and the new rows: 297 items and 229 payments instead of 268 and 211.

[Fix bills amount and balance](https://github.com/informatici/openhospital-core/pull/1683/commits/74ccd71f5c46a5b086a7d769ac32ccadc38cbaa2) then recomputed the headers. The script does the right thing on healthy data, but it groups items by price, description and unit amount and *sums the quantities*, so over the duplicated rows it made every header agree with its items at an inflated amount, and left 8 bills with a negative balance.

### What this changes

Between the merge base (`e67f8984`) and `develop`, none of `OH_BILLS`, `OH_BILLITEMS` and `OH_BILLPAYMENTS` was ever touched, so for those three tables the release side is the only authority. This drops the 47 duplicated rows (29 items + 18 payments) and restores the 14 bill headers to their released values. Nothing else in the dump is touched.

|                          | on `74ccd71f` | with this PR | release `9a57ce2f` |
| ------------------------ | ------------: | -----------: | -----------------: |
| items / payments         |     297 / 229 |    268 / 211 |          268 / 211 |
| duplicated rows          |       15 + 18 |            0 |                  0 |
| bills with negative balance |          8 |            0 |                  0 |

### How it was verified

`sql/create_all_demo.sql` from this branch loaded into a clean MariaDB 10.6:

- 176 bills / 268 items / 211 payments, matching the release
- no duplicated items or payments, no bill with a negative balance
- 66 tables, schema column by column identical to the demo database built from `develop`
- comparing the four dumps (merge base, release, `develop`, this branch) row by row on primary keys, across all 66 tables: no row resurrected, none of the release's rows lost, none introduced by the merge

Bill 64 still reports a balance that does not match `amount - payments`. That is the released behaviour, not a defect: the `DISCOUNT -23` item is excluded from `BLL_AMOUNT` but not from the balance. It is left untouched.

Sorry for landing on top of work you had just pushed, and happy to close this if you would rather regenerate the dump yourself, it is your call.


giuseppesorge:
One thing worth flagging separately, since it is not a defect of this PR but it caught us here and may catch us again.

`sql/extra/fix_bill_amount_balance.sql` groups items by bill, price, description and unit amount and sums the quantities (lines 57-60). That is exactly right when the write-off rows are legitimate negative-quantity cancellations, which is what the script was written for. But when a bill genuinely holds two copies of the same item, the group sums them instead of flagging them, and the recomputed header silently doubles rather than the mismatch showing up. That is what happened after the merge back: the totals became internally consistent and the underlying duplication became invisible.

Two options, whichever you prefer:

- a short note in the script header saying it assumes no duplicated item rows, and to check for those first;
- or a preliminary `SELECT` that reports duplicate groups and stops before the `UPDATE` if it finds any.

I did not touch the script in this PR to keep it to the data fix. Happy to add either one here or in a separate PR if you think it is worth it.

mwithi:
> Two options, whichever you prefer:
> 
> * a short note in the script header saying it assumes no duplicated item rows, and to check for those first;
> * or a preliminary `SELECT` that reports duplicate groups and stops before the `UPDATE` if it finds any.
> 
> I did not touch the script in this PR to keep it to the data fix. Happy to add either one here or in a separate PR if you think it is worth it.

Interesting but the duplicated rows are human tampering with the database, should not be something even possible with the application itself, so I don't think we should care about such guardrails, am I wrong?

giuseppesorge:
You are right, and my framing was off, sorry.

I went and checked: the application does produce repeated item lines. `PatientBillEdit.addItem` simply appends to an `ArrayList` (`billItems.add(item)`), there is no merge step and no unique constraint on `OH_BILLITEMS`, so picking the same item twice instead of setting a quantity leaves two rows with the same price code, description and unit amount.

The released demo data already has one: bill 8 carries `OTH1 / Amount per day` twice, 30 x 2 and 30 x 6, and the script groups them and sums the quantities to 240, which is exactly its `BLL_AMOUNT`. So the grouping is doing the right thing, and a "stop if there are duplicate groups" check would have refused to run on a perfectly legal bill.

What went wrong here was only the data. The merge brought back rows that were superseded revisions of items that had since been rewritten, and nothing in the table distinguishes those from a user adding the same item twice, so no check could have told them apart.

Withdrawing the suggestion, the script is fine as it is. The data fix in this PR stands on its own.

